PR #7, opened 2020-09-08
4244 - Add TriggerField component

jmacaluso711:
Adding a `IdsTriggerField` component. The IdsTriggerField component acts as a wrapper for other components we want to use to trigger events. At the moment it contains 3 different properties: 

**Properties**
- `tabbable`: which will turn on/off the tabindex
- `appearance`: potentially adds options for Normal and Compact mode or any other appearances we might need.
- `disableNativeEvents`: not sure about this one, open to feedback. Was thinking this could disable our events if the end user wants to attached their own events and not use ours.

**Events**
- `trigger`: For now this just dispatches an event called `triggerfield`

**Example**
```
<ids-trigger-field>
  <ids-input type="text"></ids-input>
  <ids-trigger-button>
     <ids-icon icon="clock"><ids-icon>
  </ids-trigger-button>
</ids-trigger-field>
```

I've also added a basic input component called `IdsInput` and a trigger button component called `IdsTriggerButton`. 

In addition, I added a file to hold all our re-usable properties, which is in `ids/base/ids-constants.js` and a `stringToBool` function to the utilities.

related to [#4244](https://github.com/infor-design/enterprise/issues/4244)

- pull and run branch
- go to http://localhost:4300/
- check the TriggerField component and make sure the event fires.
- review the code and provide any feedback necessary.


src/ids-base/ids-base.scss
```diff
@@ -3,3 +3,9 @@
 
 /* Other Css Modules */
 @import '../ids-loader/ids-loader.scss';
+
+*,
+*::before,
```
tmcconechy, line 8: Trying not to have anything in here thats global css - also for the standalone css. Can we put this right on the ids-trigger-field component itself?

src/ids-base/ids-constants.js
```diff
@@ -0,0 +1,11 @@
+export const props = {
```
tmcconechy, line 1: Whats the advantage of doing this rather than just having the strings in the element?

src/ids-base/ids-string-utils-mixin.js
```diff
@@ -12,6 +12,15 @@ const IdsStringUtilsMixin = {
       .replace(/([-_ ]){1,}/g, ' ')
       .split(/[-_ ]/)
       .reduce((cur, acc) => cur + acc[0].toUpperCase() + acc.substring(1));
+  },
+
+  /**
+   *
```
tmcconechy, line 18: Add a description?

src/ids-input/ids-input.js
```diff
@@ -0,0 +1,97 @@
...
+    return [props.TYPE, props.PLACEHOLDER, props.TABBABLE];
+  }
+
+  /**
+   * Create the Template for the contents
+   * @returns {string} The template
+   */
+  template() {
+    return `
+      <input class="ids-input-field" tabindex="0" type="${types[this.type]}" ${this.placeholder ? `placeholder="${this.placeholder}"` : ''}/>
```
tmcconechy, line 50: Is tabindex = "0" needed?

src/ids-input/ids-input.js
```diff
@@ -0,0 +1,97 @@
...
+    `;
+  }
+
+  /**
+   * Set the type of input
+   * @param {boolean} value [text, password, number, email]
+   */
+  set type(value) {
+    if (value) {
+      this.setAttribute(props.TYPE, value);
```
tmcconechy, line 60: Will setting the type change the render in the page for these attributes? Maybe add a test for that.

src/ids-input/ids-input.js
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+
+  get type() { return this.getAttribute(props.TYPE); }
+
+  /**
+   * Set if the input is tabbable
+   * @param {boolean} value True of false depending if the input is tabbable
+   */
+  set tabbable(value) {
```
tmcconechy, line 72: I think tabbable is always needed for inputs so this may not be needed. But for trigger field its whether or not you can tab to the button part so this should be just adding tabindexes to the button. (Two tab stops). that was the use case.

src/ids-input/ids-input.scss
```diff
@@ -0,0 +1,8 @@
+@import '../ids-base/ids-base';
+
+.ids-input-field {
+  @include px-8();
+  @include py-8();
+
+  width: 100%;
```
tmcconechy, line 7:
```suggestion
Will add a @w type for this (isnt one now)
```

src/ids-trigger-button/ids-trigger-button.scss
```diff
@@ -0,0 +1,8 @@
+@import '../ids-base/ids-base';
+
+.ids-trigger-button {
+  @include bg-transparent();
+  @include cursor-pointer();
+
+  border: none;
```
tmcconechy, line 7:
```suggestion
Maybe @include border-0 will work
```

src/ids-trigger-field/ids-trigger-field.d.ts
```diff
@@ -0,0 +1,18 @@
+// Ids is a JavaScript project, but we define TypeScript declarations so we can
+// confirm our code is type safe, and to support TypeScript users.
+
+interface nativeElement extends HTMLElement {
+  tabbable: boolean;
+  appearance: 'normal' | 'compact' | string;
```
tmcconechy, line 6:
```suggestion
Like this name for now. Might reconsider later (compactness/size ect?)
```

src/ids-trigger-field/ids-trigger-field.js
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  get appearance() { return this.getAttribute(props.APPEARANCE); }
+
+  /**
+   * Set if the button handles events
+   * @param {boolean} value True of false depending if the button handles events
+   */
+  set disableNativeEvents(value) {
+    const isDisabled = this.utilities.stringToBool(value);
```
tmcconechy, line 82: Not wired in yet right? Im not sure if we will need this so sort of wanted to see it in action

src/ids-trigger-field/ids-trigger-field.scss
```diff
@@ -0,0 +1,17 @@
+@import '../ids-base/ids-base';
+
+.ids-trigger-field {
+  position: relative;
```
tmcconechy, line 4: We can add more mixins for these

test/ids-trigger-field/ids-trigger-field-func-test.js
```diff
@@ -0,0 +1,43 @@
...
+
+  it('renders correctly', () => {
+    expect(triggerField.outerHTML).toMatchSnapshot();
+
+    triggerField.icon = 'close';
+    expect(triggerField.outerHTML).toMatchSnapshot();
+  });
+
+  it('fires triggerevent on trigger', () => {
+    triggerField.addEventListener('beforetagremoved', (e) => {
```
tmcconechy, line 36: Dont think is is right name?

src/ids-base/ids-constants.js
jmacaluso711: Was thinking here that if we have all the props that we know will be repeated in multiple components, for example `tabbable` it will be easier to manage. This way we avoid simple typos in the individual component and if we want to rename a property we only have to do it once. Does that make sense?

src/ids-trigger-field/ids-trigger-field.js
jmacaluso711: Yeah, this was just an idea. I'll wire it up just to show what I was thinking. Can make a note to discuss when we have a check in.

test/ids-trigger-field/ids-trigger-field-func-test.js
jmacaluso711: Oops

src/ids-trigger-field/ids-trigger-field.js
jmacaluso711: Wired this up. The thinking here was if a user wanted to use our component but attach their own events they could disable ours.

jmacaluso711:
@tmcconechy for the label, do you think we want to use the current `ids-label` component and add different types, for example, `type='input-label'` would render a label for an input, etc. Or do we want a specific `ids-input-label` component.

tmcconechy (commented):
Looks good for now.. We can continue on later with labels ect ... Maybe we just want to fix the console.log in the checks and the mixins (seperate PR?)

tmcconechy:
@jmacaluso711 re:

`do you think we want to use the current ids-label component and add different types, for example, type='input-label' would render a label for an input, etc. Or do we want a specific ids-input-label component.`

I was thinkin to use the existing ids-label component and then maye a type option that would change the template from a span to a label.

jmacaluso711:
@tmcconechy Removed the console.log... I'm down for calling this one good for now and tackling the label in a separate issue/pr.

tmcconechy:
Agreed we can always add more later on top.

tmcconechy (approved):

PR #11, opened 2020-10-09
2941 - Layout Grid

jmacaluso711:
Adding several settings to the grid layout component. This implementation makes use of CSS custom vars to set the amount of columns, rows, column/row span, start and end. In a previous iteration I was using class names that were generated in a loop. I think this new way will allow for more flexibility. I've also added a setting to adjust the grid gap, auto layout and margins. Additionally, there is test coverage for all settings and a sandbox page to test the implementation. Also proposing a name change from `ids-layout-column` to `ids-layout-panel` as it seems more fitting. This is a work in progress as of now, looking to get a little feedback.

closes #2941

- Pull and run branch
- Go to localhost:4300 and inspect the grid elements.
- Additionally you can go to http://localhost:4300/ids-layout-grid/test-sandbox and see more examples.

- [] A note to the change log.


app/ids-layout-grid/standalone-css.html
```diff
@@ -4,9 +4,9 @@
 <ids-layout-grid>
   <ids-label font-size="12">Layout Standalone CSS</ids-label>
 </ids-layout-grid>
-<div class="ids-layout-grid">
-  <div class="ids-layout-column ids-background-fill"><ids-label font-size="12">A</ids-label></div>
-  <div class="ids-layout-column ids-background-fill"><ids-label font-size="12">B</ids-label></div>
-  <div class="ids-layout-column ids-background-fill"><ids-label font-size="12">C</ids-label></div>
-  <div class="ids-layout-column ids-background-fill"><ids-label font-size="12">D</ids-label></div>
+<div class="ids-layout-grid ids-layout-cols-auto ids-grid-gap-md">
```
tmcconechy, line 7: Should ids-grid-gap-md be ids-layout-grid? Grids can get confused elsewise.

Also i noticed the class is ids-layout-cols but you changed to ids-layout-panel? Not really against panel vs column but usually see grids with columns vs panels. Panel could be like a modal or card.

src/ids-base/ids-constants.js
```diff
@@ -7,6 +7,19 @@ export const props = {
   APPEARANCE: 'appearance',
   ANIMATED: 'animated',
   VISIBLE: 'visible',
+  GRID_FIXED: 'fixed',
```
tmcconechy, line 10: Maybe these should be for example FIXED instead of GRID_FIXES as thats the ids right? These might be used in non grid places (these are a bit specific but stylistically)

src/ids-layout-grid/ids-layout-grid.scss
```diff
@@ -1,30 +1,155 @@
 /* Ids Layout Grid Component */
 @import '../../node_modules/ids-identity/dist/theme-uplift/tokens/web/theme-uplift-mixins.scss';
 
+$breakpoint-xs: 360px;
```
tmcconechy, line 4: Suggest we put these in the css properties as they are tokens.

src/ids-layout-grid/ids-layout-grid.scss
```diff
@@ -1,30 +1,155 @@
...
 }
 
 // Add Optional Background Fill
 ::slotted(.ids-background-fill),
-.ids-layout-column.ids-background-fill {
+.ids-layout-grid > .ids-background-fill {
   @include py-40();
   @include px-16();
 
-  background-color: #efeff0; // Get this one from a variable somehow
```
tmcconechy, line 22: I meant from a css variable... But this is fine for now.

webpack.config.js
```diff
@@ -167,6 +167,13 @@ module.exports = {
       chunks: ['ids-layout-grid/ids-layout-grid', 'ids-label/ids-label'],
       title: 'Layout - Standalone Css'
     }),
+    new HTMLWebpackPlugin({
```
tmcconechy, line 170: FYI: In my PR this will be automatic now but its not ready

app/ids-layout-grid/example.html
```diff
@@ -1,10 +1,39 @@
-<ids-layout-grid>
-  <ids-label font-size="12">Layout Grid</ids-label>
+<ids-layout-grid auto="true" gap="md">
+  <ids-label font-size="12" type="h1">Layout Cols</ids-label>
```
tmcconechy, line 2: Still Layout Grid Maybe? (Cols vs Panels)

tmcconechy (commented):
Looks good - just a few clarifications
- col - vs panel? just asking the explanation... If we stick with panel it still sats "Layout Col" in the example (and the class mentioned). maybe label it as Responsive Grid in the label or something?

- can we add a typings file and a README.md it would really help to document the config as it can get a bit confusing (and while fresh) and would help the understanding of the settings and examples
- can we get more test coverage for ids-layout-grid.js  ids-layout-panel.js? give it a try and maybe try an e2e test like tags with visual regressions

tmcconechy:
I like the row and colspan feature and the 2 up on mobile a lot 👍

src/ids-layout-grid/ids-layout-grid.scss
jmacaluso711: I meant to add a setting for this. It can probably come from there. Will add that

app/ids-layout-grid/standalone-css.html
jmacaluso711: Yeah makes sense. Noting here that after discussing we decided on `ids-layout-grid-cell` instead of `ids-layout-panel`

tmcconechy:
Maybe just a md and remove WIP and should be good for now 👍

EdwardCoyle:
Looks good to me.  Very flexible compared to what's in EP

tmcconechy (approved):

EdwardCoyle (approved):

PR #12, opened 2020-10-14
4390 - Implement IDS Button / Toggle Button

EdwardCoyle:
This PR Implements the foundation for the IDS Button component:
- A base set of button types/styles is implemented
- Additional button prototypes (Toggle Button / Trigger Button) are now extending the `IdsButton` class.
- Basic states are shared by all button types (disabled/focusable/focused/clicked), with Toggle Button having its own "pressed" state on top.
- Lots of tests

This PR still doesn't address the prototype of the Menu Button.  There are also some parts of the IdsButton not covered by tests yet because it's currently not possible to invoke the component with JS settings (need to discuss if this is necessary or not).

Closes infor-design/enterprise#4390

- Pull this branch, build, run the app
- Open kitchen sink: http://localhost:4300
- Test all the buttons, make sure they work when clicked, all should be focusable
- Click the Toggle button and test the pressed state (on/off)
- In web inspector, pick any button and add a `disabled="true"` attribute to make the button disabled (all states on all variants should match what we currently expect from the Uplift/Vibrant theme).
- In web inspector, pick any button and add a `tabindex="-1"` attribute.  This makes it impossible to focus the button via keyboard (clicking with mouse / touch is still possible). Make it focusable again with `tabindex="0"`.  Note that this attribute is passed into the inner button, and will not exist on the custom element once its set.
- Open http://localhost:4300/ids-button/test-fallback-slot and make sure the buttons that exist there appear to work.
- Open http://localhost:4300/ids-toggle-button/index.html and make sure the toggle button works.
- Check all documentation for both IDS Button and IDS Toggle Button.


app/ids-button/index.js
```diff
@@ -0,0 +1,16 @@
...
+// Supporting components
+import IdsIcon from '../../src/ids-icon/ids-icon';
+import IdsLabel from '../../src/ids-label/ids-label';
+import IdsLayoutGrid from '../../src/ids-layout-grid/ids-layout-grid';
+
+document.addEventListener('DOMContentLoaded', () => {
+  // Add an event listener to test clickable links
+  document.querySelectorAll('ids-toggle-button').forEach((idsButton) => {
+    idsButton.addEventListener('click', (e) => {
+      e.target.pressed = !e.target.pressed;
```
tmcconechy, line 13: What does this do?

src/ids-base/ids-mixins.scss
```diff
@@ -0,0 +1,48 @@
+/**
+ * Mixins in this file may want to eventually be moved to the Design Tokens.
+ */
+
+// All vendor prefixes
+$prefixes: -webkit-, -moz-, -o-, -ms-, '';
```
tmcconechy, line 6: I was going to use https://medium.com/@realfrancisyan/scss-autoprefixer-with-webpack-9f8456ae751b (autoprefixer) for these. But it got moved out of the web pack config during some changes and i forgot. Perhaps you can look at setting this up instead?

webpack.config.js
```diff
@@ -119,6 +120,27 @@ module.exports = {
       title: 'IDS Enterprise Web Components',
       chunks: ['index']
     }),
+    new HTMLWebpackPlugin({
```
tmcconechy, line 123: I will merge this out when i merge my PR FYI - as i have it working without adding config now

tmcconechy (commented):
Looking very good. Some initial comments:

- what would we use focusable="false" for?
- do we want to get rid of the "no_focus" buttons? I think maybe we can try as it is a bit non standard
- does a button need a named slot? or can you do this?
```
    <ids-button id="test-button-primary" type="primary">Primary Button</ids-button>
```
- maybe start a new md file similar to tags readme.md?
- seems like a bit of misalignment in chrome on the icon + text a) the text sits a bit low and is a bit close to the icon (try 8px between icon and text or 4?)


Not much else at the moment this looks good

EdwardCoyle:
@tmcconechy I'll try to fix these, but just addressing some of your comments:

> - what would we use focusable="false" for?
> - do we want to get rid of the "no_focus" buttons? I think maybe we can try as it is a bit non standard, by try i mean dont implement it until we ever get asked 👍

I'm realizing now that `focusable` probably isn't needed.  Originally I thought that the encapsulated Button element could still be focused even with a `tabindex="-1"` on the custom element (thought I saw this in testing) but it appears that's not the case on any browser.  I'll take that out to make it cleaner.

The cases I had in mind were:
- Do what @jmacaluso711 was doing to disable focus on the trigger buttons at the button level instead of the field level
- This would also eventually help the Toolbar case along ("overflowed" buttons need to be made non-focusable when they hidden)

However, since `tabindex` works we should just use that.

> - does a button need a named slot? 

Currently it does.  I'm still trying to figure out how I could handle a case where you either have just the one un-named slot (text only) vs. two named slots (icon + text), and how I could remove the icon and convert to just a single slot.  Also, for toggle buttons, the icon is required, so the case of what would be in the unnamed slot would change.  With the current design the slots' contents also determine the look of the button to some degree (not having text makes it an "icon" button style with round edges, fx).  I'm not sure its as easy as just passing in text and an icon into a single unnamed slot

> the text sits a bit low

Any ideas on how we could address this globally?  I can't seem to get these two browsers to line these things up evenly without introducing a "firefox only" or "chrome only" style to push around the text or icon.  I think it's an issue with the font.  The button styles are all using `inline-flex` on the `:host` as well.

app/ids-button/index.js
EdwardCoyle: This actually changes the Toggle button's pressed state.  It doesn't do this internally atm.

EdwardCoyle:
TODO:

- [ ] Remove focusable
- [ ] Figure out slotting
- [ ] Add `README.md` for buttons
- [ ] Fix Trigger button integration (needs to occur after removing focusable)
- [ ] On Toggle Buttons, actually move the `pressed` state change from its test page into an internal click event?
- [ ] There was a change to icons here that actually uses the `size` attribute to set `width`/`height` styles on the element.  Needed this for improved alignment within the button, but I'm just realizing that it broke the layout of the Tag icons, so I'll fix that as part of this PR too.
- [ ] Probably should have some actual examples of `disabled` buttons.

tmcconechy:
Cool Cool.

- ok if we need focusable="false" i think we can keep it. im not sure about the toolbar case. But i think the trigger button case feature could just be done with a tabindex = -1 , but the focusable vs non focusable trigger button is a case we want to add.
- maybe you can add both a named slot and an unnamed slot as a fallback like this example https://stackoverflow.com/a/53197524/852398 it would be nice to use it as text only and only have to use the names if both are used. Or other ideas to try might be making text always unnamed then only icon has a name? Or looking for a slot with no name and attaching the name property to it.
- found to center the text that this seems to work when added to the span slot ="text"

```
  padding-left: 4px; //add space mentioned
  vertical-align: bottom;
}
```

you can style it like this if you dont have a rule for it already?
```
slot[name=text]::slotted(*) {
  padding-left: 4px; //add space mentioned
  vertical-align: bottom;
}
```

EdwardCoyle:
I'm not convinced we need the `focusable` setting anymore, I'm open to just using `tabindex` directly (seems to work)

tmcconechy:
This done pretty much then @EdwardCoyle ? Just the merge conflicts and remove focusable?

EdwardCoyle:
@tmcconechy I'm going to address everything in that TODO list comment above, then we can call it good 👍

EdwardCoyle:
@tmcconechy calling this one done now.  One thing I didn't address "fully":

> On Toggle Buttons, actually move the pressed state change from its test page into an internal click event?

I did add a `toggle()` method for backwards compatibility with 4.x, but I didn't make any click events automatically trigger the toggle from within.  Wondering if we should leave that to developers to decide (when to make the press change), rather than add in the whole "veto" feature for this?

Other than that there are a couple API features we could add (listed them in TODOs), but I wouldn't mind calling this a save point and merging.

tmcconechy (approved):

jmacaluso711 (approved):
🍌

PR #13, opened 2020-10-15
4458 - Add field, trackdirty, validation and update input, labels

deep7102:
Added some basic components and mixin as below
- ids-field: wrapper for form fields
- ids-validation-message: wrapper for validation messages
- ids-trackdirty-mixin: to keep track user changes
- ids-validation-mixin: to check/set validation on form fields
- ids-icon: added new size `15`, use for validation message

Update some feature as:
ids-input: add attributes state, validation and dispatch basic events
ids-label: add attributes state, audible, required

Fixes infor-design/enterprise#4458

- Navigate to: http://localhost:4300
- See the styles for validation message (icons, color and audible text for each type)
- Test ids-fields as
   - First Name (required and trackdirty)
   - Last Name (trackdirty only)
   - Email (required, email and no trackdirty)
- Test enable/disable/readonly 


src/ids-base/ids-trackdirty-mixin.js
```diff
@@ -0,0 +1,137 @@
+/**
+ * Track changes on inputs elements and show a dirty indicator.
+ */
+const IdsTrackdirtyMixin = {
+  /**
+   * Should use trackdirty OR not, init false
+   * @private
+   */
+  useTrackdirty: false,
```
tmcconechy, line 9: Whats the idea around this? i dont follow

src/ids-input/ids-input.js
```diff
@@ -79,6 +99,131 @@ class IdsInput extends IdsElement {
...
+    this.removeAttribute(props.VALUE);
+  }
+
+  get value() { return this.getAttribute(props.VALUE); }
+
+  /**
+   * Set the `fieldState` attribute of input
+   * @param {string} val the value property
+   */
+  set fieldState(val) {
```
tmcconechy, line 122: What is field-state for? It seems a bit generic sounding i dont follow it.

src/ids-label/ids-label.js
```diff
@@ -49,34 +61,85 @@ class IdsLabel extends IdsElement {
...
+    this.rerender();
+  }
+
+  get audible() { return this.getAttribute(props.AUDIBLE); }
+
+  /**
+   * Set the `fieldState` attribute of input
+   * @param {string} val the value property
+   */
+  set fieldState(val) {
```
tmcconechy, line 131: Again im not sure about fieldState - maybe we need to be more specific with a name?

src/ids-label/ids-label.scss
```diff
@@ -6,6 +6,35 @@
...
+
+    content: '*';
+    display: inline-block;
+    font-size: 22px;
+    left: 3px;
+    line-height: 0;
+    top: 3px;
+  }
+
+  &.audible {
```
tmcconechy, line 26: I think Ed has this on His PR as well but i like having it in ids-label. Css @EdwardCoyle

tmcconechy (changes requested):
Just a few comments but looks good

- [ ] Some errors in the console on http://localhost:4300/ids-field
- [ ] The labels arent working (cant click them to focus the field). I tried adding a for and id and this didnt work. So what im thinking we try is add the label in with input. The means that we cant use the ids-label component in it. It would be like this in the template. Lets see if this works.

```
<label for="x">
</input id="x">
```
- [ ] Can we add more complete input styles from current uplift?
- [ ] Maybe ids-dirty-tracker sounds better than trackdirty
- [ ] Eventually tests and docs and a typings file(s)

src/ids-base/ids-trackdirty-mixin.js
deep7102: - I was doing something else, and this logic left in there. You right we do not need this anymore because now doing different way, so going to remove `useTrackdirty`

deep7102:
> Just a few comments but looks good
> 
>     * [1]  Some errors in the console on http://localhost:4300/ids-field
> 
>     * [2]  The labels arent working (cant click them to focus the field). I tried adding a for and id and this didnt work. So what im thinking we try is add the label in with input. The means that we cant use the ids-label component in it. It would be like this in the template. Lets see if this works.
> 
> 
> ```
> <label for="x">
> </input id="x">
> ```
> 
>     * [3]  Can we add more complete input styles from current uplift?
> 
>     * [4]  Maybe ids-dirty-tracker sounds better than trackdirty
> 
>     * [5]  Eventually tests and docs and a typings file(s)

1) will check for errors: http://localhost:4300/ids-field
2) yes, same way I thought we need both input + label under one shadow root. I was thinking `ids-input` to extend the `ids-label` this way we do not have to manage different set of styles/features as `ids-label` having we should do this for all low level elements which go with `label + element` like textarea, checkbox, radio etc.

3) yes, will check for input styles from current uplift
4) yes, I do like `ids-dirty-tracker` better than `trackdirty`
5) yes, was thinking to do after all final then write tests and docs about

src/ids-icon/ids-icon.js
```diff
@@ -6,7 +6,8 @@ const sizes = {
   large: 24,
   normal: 18,
   medium: 18,
-  small: 10
+  small: 10,
+  15: 15
```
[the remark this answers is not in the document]
deep7102: - sure, will try both `small and medium` sizes. I was trying to match with EP styles

src/ids-input/ids-input.js
deep7102: - `field-state` to keep track for `enabled/disabled/readonly`. Put this on `ids-filed`, because `ids-filed ` is main root level for input and see that as main input component which should hold more info about like dirty-tracker, error-msg or what type of input inside etc. Please let me know if you have other name for this property was can update

src/ids-label/ids-label.js
deep7102: Please let me know if you have other name for this property was can update

src/ids-icon/ids-icon.js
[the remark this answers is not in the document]
tmcconechy: Yeah please do - we want to remove this inconsistency in sizes.

src/ids-input/ids-input.js
tmcconechy: Does it work to just only use enabled/disabled/readonly each on their own?

tmcconechy:
For the label problem. Try adding the label in with ids-input for now and get it working. I think i can see maybe that we dont need ids-field then. As all elements like ids-input , ids-checkbox ect will all container their own label.

Then add a label attribute on the ids-input and then the label is in the template? Think that could work.

src/ids-label/ids-label.scss
EdwardCoyle: The use case on mine is similar, but it's specifically on "audible" spans inside of buttons.  Makes sense to have it in both.  Is there a way we could share it -- does it belong in the base styles?

tmcconechy: What i was thinking maybe is 
- if labels are in the form element components we dont need to support labels in ids-label
- rename ids-label to ids-text
- add an audible property to it
- use it in both places as ids-text to make the audible spans.

deep7102:
Close this, so can use other PR https://github.com/infor-design/enterprise-wc/pull/16

PR #19, opened 2020-11-11
4456 - Implement IDS RenderLoop

EdwardCoyle:
This PR adds the RenderLoop to the WebComponents system, which allows access to a single instance of a `requestAnimationFrame` loop tied to a callback queue:
- Implements/Cleans Up the 4.x RenderLoop/RenderLoopItem APIs
- Adds `IdsRenderLoopMixin` for access to the global instance on each component.
- Adds full documentation/typescript defs
- Adds many tests
- Replaces the `setTimeout` calls that were in the Ids Popup component with `IdsRenderLoopItem`s

Closes infor-design/enterprise#4456

- Pull this branch, build, run the demoapp
- Open http://localhost:4300/ids-render-loop.  Click the button and watch the count down.  It should get to "0" and say "DONE!".  This one is also present on the kitchen sink page (check that to make sure it works)
- Open http://localhost:4300/ids-render-loop/test-elapsed-time.  Click both "play/pause" buttons to watch the counts change.  When only using the "Play/Pause Item" button, the item count will stop, but the loop count will always continue.  When using "Play/Pause Loop", both will stop.  The counters (including the `totalStoppedTime` counters) will not change while the loop is stopped.
- Open http://localhost:4300/ids-render-loop/test-flying-popup.  Make sure there is a Popup flying around and switching colorful targets every 1000ms or so.
- Open http://localhost:4300/ids-popup and make sure it still works (setTimeout calls have been replaced with RenderLoop items)
- Read the documentation and make sure it makes sense.


tmcconechy:
This looks good. But just some cleanup.

1) the conflicts
2) `npm run lint`

EdwardCoyle:
I'll check out the conflicts/lints shortly!

The samples are all demonstrating slightly different functionality.  The "flying" popup sample isn't demonstrating timeouts as much as it's showing the `updateCallback`, which is more like a `setInterval()` with pause/destroy features.

The elapsed time test is trying to demonstrate that it's possible to stop/start the loop, as well as each individual loop item independently.  I had trouble demonstrating this feature but the idea is that these time controls are a bit more "exposed" now to components than they were in 4.x, so I wanted to show how you'd access/use them properly.  I guess we could do this with the tests instead but I was trying to demo it with a usable sandbox.

tmcconechy:
OK, i guess we will just have to go back on some of these examples later and clean them up when we have more styles and patterns available.

Also what about the naming of tests `test-xxx` and `example-xxx` i was going away from that and my idea was in general everything is an important example. And a test is done as a test. Hoping to reduce the number of examples that require maintenance than before. So thats why im thinking that way.  But im open.

EdwardCoyle:
@tmcconechy fixed those items

deep7102 (approved):

PR #28, opened 2020-12-28
4628 - Add Ids Popupmenu

EdwardCoyle:
This PR implements basic functionality for Popupmenus in a Webcomponent.

Closes infor-design/enterprise#4628

Pull this branch, build, and run the demoapp.
Then, check the following examples, testing selection/disabled/etc:

- http://localhost:4300/ids-menu
- http://localhost:4300/ids-popup-menu/selected-state
- http://localhost:4300/ids-popup-menu/standalone-css
- http://localhost:4300/ids-popup-menu/menu-button (Might need to make this an actual component)

- [] An e2e or functional test for the bug or feature.
- [] A note to the change log.


app/ids-popup-menu/index.js
```diff
@@ -0,0 +1,30 @@
...
+  const popupmenuEl = document.querySelector('ids-popup-menu');
+  const popupEl = popupmenuEl.popup;
+
+  // Preconfigure the Popup
+  popupEl.align = 'top, left';
+
+  // Add a console log for all the others on `selected`
+  const menuItems = document.querySelector('#popupmenu').items;
+  menuItems.forEach((item) => {
+    item.addEventListener('selected', (e) => {
```
tmcconechy, line 26: Do you need an event on each element to handle selection as i dont know if thats the best.

src/ids-base/ids-resize-mixin.js
```diff
@@ -78,6 +78,17 @@ const IdsResizeMixin = {
     return typeof ResizeObserver !== 'undefined' && this.ro instanceof ResizeObserver;
   },
 
+  /**
+   * @returns {HTMLElement} that should receive a resize observer
```
tmcconechy, line 82: Maybe a comment - unclear what this does

src/ids-icon/ids-icon.scss
```diff
@@ -3,5 +3,6 @@
 
 :host {
   @include inline-flex();
-  @include self-center();
```
tmcconechy, line 6: Whats wrong here? I think we have other code relying on it now

src/ids-menu/ids-menu-group.js
```diff
@@ -0,0 +1,137 @@
...
+    super();
+  }
+
+  /**
+   * Return the properties we handle as getters/setters
+   * @returns {Array} properties
+   */
+  static get properties() {
+    return [
+      'keep-open',
```
tmcconechy, line 32: Add to props?

tmcconechy (commented):
Looks really good just some minor comments

http://localhost:4300/ids-menu
- focus state is the browser default (all examples)

http://localhost:4300/ids-popup-menu
- should use left and right arrow not enter to toggle the sub menu
 
http://localhost:4300/ids-popup-menu/selected-states
- didnt open?

http://localhost:4300/ids-popup-menu/menu-button
- can we make this a icon+text button with down arrow instead?
- show filter row didnt seem to toggle

app/ids-popup-menu/index.js
EdwardCoyle: Probably not, I think I'd written this before the event bubbling fix.  I'll address it soon.

src/ids-icon/ids-icon.scss
EdwardCoyle: I had issues with icon alignment due to this rule but I must have fixed them. After putting this back, it's no longer an issue. 👍

src/ids-base/ids-resize-mixin.js
EdwardCoyle: I had an issue with the Popupmenu's nested menus where `Element.parentNode` would sometimes resolve to a `DocumentFragment` type instead of a node.  In general I think this will only happen when the templates for complex WebComponents contain other WebComponents.  If a DocumentFragment is in the page already (which in this case it is), you just have to get the `host` property of the fragment to get at the actual WebComponent "node".

I'll get a brief description of it in the notes.

EdwardCoyle:
@tmcconechy I still need to finish test coverage and docs but I believe I've addressed all the usability issues/states/etc (basically all your comments).  Also realized we still needed CSS-only for the Popupmenu, so that's there now too.

app/ids-menu/example.html
```diff
@@ -0,0 +1,56 @@
+<ids-layout-grid>
+  <ids-text font-size="12" type="h1">Menu</ids-text>
+</ids-layout-grid>
+
+<ids-layout-grid auto="true">
+  <ids-layout-grid-cell>
+
+    <ids-menu id="simple-menu">
```
tmcconechy, line 8: Just wondering but it works in or out of grid cells?

app/ids-menu/index.js
```diff
@@ -0,0 +1,20 @@
...
+  const nonSelectableItem = document.querySelector('#no-select');
+  nonSelectableItem.addEventListener('beforeselected', (e) => {
+    console.log('%c You cannot select this item', 'color: #ff0000;', e.detail.elem);
+    e.detail.response(false);
+  });
+
+  // Log to the console on `selected`
+  const menu = document.querySelector('#complex-menu');
+  menu.addEventListener('selected', (e) => {
+    console.log(`Item "${e.detail.elem.text}" was selected`);
```
tmcconechy, line 18: I think you can do `console.info` here or add or eslint-ignore or remove

tmcconechy (dismissed):
Looks nice 🍌  feel free to fix tests up and linting problems and i'll reapprove

app/ids-menu/example.html
EdwardCoyle: Yep it definitely works, but I put it inside the cells here to make the two-up (alongside one another instead of stacked)

tmcconechy (approved):
Looks great - nice work 🍌

tmcconechy (commented):
@EdwardCoyle I think we should push this in now as soon as the tests pass. But they seem not to be passing.

Looks like Percy is ready for the month now. But the only thing i see is perhaps the right click test is necessary as its just some text in the page... Or have to make the popup go?

PR #37, opened 2021-01-28
4727 - Add Ids Alert wc

ericangeles:

Implements IDS Alert for the web components. The Ids Alert component acts a wrapper that can be use for different alert messages. 

Closes https://github.com/infor-design/enterprise/issues/4727

- Go to http://localhost:4300/ids-alert
- Check the codes and review it


src/ids-alert/ids-alert.d.ts
```diff
@@ -0,0 +1,7 @@
+// Ids is a JavaScript project, but we define TypeScript declarations so we can
+// confirm our code is type safe, and to support TypeScript users.
+
+export class IdsAlert extends HTMLElement {
+  /* Types of alert */
+  type: 'error' | 'warning' | 'success' | 'info';
```
tmcconechy, line 6: I think there is a few more types?

app/ids-alert/example.html
```diff
@@ -0,0 +1,29 @@
+<ids-layout-grid auto="true">
+  <ids-layout-grid-cell>
+    <ids-text font-size="12" type="h1">Alerts</ids-text>
+  </ids-layout-grid-cell>
+</ids-layout-grid>
+
+<ids-layout-grid auto="true">
+  <ids-layout-grid-cell>
+    <ids-alert type="alert" icon="alert" size="normal"></ids-alert>
```
tmcconechy, line 9: Im thinking we dont need both a type and an icon. So maybe just type.

src/ids-alert/ids-alert.js
```diff
@@ -0,0 +1,110 @@
+import pathData from 'ids-identity/dist/theme-uplift/icons/standard/path-data.json';
```
tmcconechy, line 1: This file is very big so i think what we should do is just use the <ids-icon> instead of this. So try not to import this and use <ids-icon> in the template.

src/ids-alert/ids-alert.js
```diff
@@ -0,0 +1,110 @@
...
+    return pathData[this.icon];
+  }
+
+  /**
+   * Create the template for the contents
+   * @returns {string} The template
+   */
+  template() {
+    const size = sizes[this.size];
+    return `<svg xmlns="http://www.w3.org/2000/svg" stroke="currentColor"
```
tmcconechy, line 49: This is redundant as mentioned. I think we should use <ids-icon> in here directly instead of duplicating its code.
If any problem setting the alert color we can make a color setting on the ids-icon (if not already there)?

This should reduce code

tmcconechy (changes requested):
Try to use <ids-icon instead> so we dont need to the big import twice. Also try and do 100% coverage when done with that change.

Otherwise looks good 👍🏻 

Thanks

app/ids-alert/example.html
```diff
@@ -0,0 +1,28 @@
...
+    <ids-alert type="empty-circle"></ids-alert>
+    <ids-alert type="half-empty-circle"></ids-alert>
+    <ids-alert type="error"></ids-alert>
+    <ids-alert type="info"></ids-alert>
+    <ids-alert type="alert"></ids-alert>
+    <ids-alert type="new"></ids-alert>
+    <ids-alert type="in-progress"></ids-alert>
+    <ids-alert type="error"></ids-alert>
+    <br /><br />
+    <ids-alert type="success-solid"></ids-alert>
```
tmcconechy, line 19: Rethinking this setting name... If it is one to one with the icons in <ids-icon> maybe we should call the setting `icon` so its consistent

tmcconechy:
@ericangeles 

1. dont worry about the lcov error its fixed on the other PR
2. but the only thing left is please try and make the coverage 100%

```
_ids-alert | 80 | 50 | 83.33 | 80 |  
-- | -- | -- | -- | -- | --
_ ids-alert.js | 80 | 50 | 83.33 | 80 | 54-57
```

By adding more tests see coverage/index.html for details

src/ids-icon/ids-icon.scss
```diff
@@ -1,5 +1,6 @@
 @import '../../node_modules/ids-identity/dist/theme-uplift/tokens/web/theme-uplift-mixins.scss';
 @import '../ids-base/ids-base';
+@import '../ids-alert/ids-alert.scss';
```
tmcconechy, line 3: I think this isnt needed and perhaps whats causing the UI changes here https://percy.io/00aaa841/enterprise-wc/builds/8609550/changed/489246062?browser=chrome&browser_ids=9&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=new&width=1280&widths=375%2C1280

src/ids-alert/ids-alert.scss
```diff
@@ -0,0 +1,47 @@
+@import '../../node_modules/ids-identity/dist/theme-uplift/tokens/web/theme-uplift-mixins.scss';
+@import '../../node_modules/ids-identity/dist/theme-uplift/tokens/web/theme-uplift.scss';
+@import '../ids-base/ids-base';
+
+svg {
```
tmcconechy, line 5:
```suggestion
svg {
```

Probably better to do this on a class you add like ids-alert?

ericangeles: For some reason, it doesn't change the stylings unless I import it in the `ids-icon.scss`.

tmcconechy: I can have a look maybe. what are you trying to change with it?

ericangeles: @tmcconechy sure, you can take a look.

ericangeles: @tmcconechy i think i found the solution for this. Will push some updates later.

src/ids-alert/ids-alert.scss
```diff
@@ -0,0 +1,47 @@
+@import '../../node_modules/ids-identity/dist/theme-uplift/tokens/web/theme-uplift-mixins.scss';
+@import '../../node_modules/ids-identity/dist/theme-uplift/tokens/web/theme-uplift.scss';
+@import '../ids-base/ids-base';
+
+:host {
+  @include mr-20();
```
tmcconechy, line 6:
```suggestion
  @include mr-8();
```

Would less padding work ok? Because 1. We want to work in 4s now (2, 4, 8, 16) and i think 20 seems like a lot and is something we would probably have to negate somehow when using the alerts later.

tmcconechy (dismissed):
Looks good - just thinking maybe reduce the internal margins? I assume thats just to seperate them in the example?

Ideally it doesnt have any margin/padding at all in it and thats done outside?

ericangeles:
> Looks good - just thinking maybe reduce the internal margins? I assume thats just to seperate them in the example?
> 
> Ideally it doesnt have any margin/padding at all in it and thats done outside?

Sure, Tim. Give me a sec.

tmcconechy:
Maybe it would work if you use the way the icon page is built with the layout grid doing the separation instead of the internal margin?

tmcconechy (approved):

EdwardCoyle (approved):

PR #38, opened 2021-01-28
4390 - Minor IdsButton improvements

EdwardCoyle:
This PR changes a couple minor things in IdsButton:
- Replaced the `setTimeout` in the ripple effect method to use IdsRenderLoop instead
- Code cleanup for property names (all now point to prop names in `src/ids-base/ids-constants.js`)
- Standardized `tabIndex` to be camelCase like the Button base prop, and linked it to `domUtils.isTrueBooleanAttribute()` for null/empty check.  Added this in menu to handle boolean attributes and it does [per the guidelines here](https://github.com/infor-design/enterprise-wc/blob/master/doc/COMPONENTS.md#boolean-attributes)
- Adds an `iconAlign` property/attribute that can be used to swap the location of a button's icon from `start` to `end` (left/right).  Adds supporting examples, tests, and docs for this feature.

- Related to infor-design/enterprise#4390
- Follows Up: #12 

- All tests should pass.
- Open http://localhost:4300/ids-button, pick a button, and try setting `tabIndex`, `type`, `disabled`, and `id` props.  Everything should work the same.
- Click/Key "Enter" on the button and ensure it still ripples
- Smoke test any components that extend `IdsButton` to make sure it all still works.
- Try the Toggle Button sample as well: http://localhost:4300/ids-toggle-button/icon-align-right
- Test Standalone CSS: http://localhost:4300/ids-button/standalone-css

### DIFF
--- a/src/ids-button/ids-button.js
+++ b/src/ids-button/ids-button.js
@@ -5,12 +5,11 @@ import {
   scss
 } from '../ids-base/ids-element';
 
+import { IdsDomUtilsMixin as domUtils } from '../ids-base/ids-dom-utils-mixin';
 import { IdsEventsMixin } from '../ids-base/ids-events-mixin';
 import { IdsRenderLoopMixin, IdsRenderLoopItem } from '../ids-render-loop/ids-render-loop-mixin';
-
-// @ts-ignore
-import { IdsStringUtilsMixin as stringUtils } from '../ids-base/ids-string-utils-mixin';
 import { props } from '../ids-base/ids-constants';
+
 // @ts-ignore
 import styles from './ids-button.scss';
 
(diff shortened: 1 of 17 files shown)


src/ids-button/ids-button.js
```diff
@@ -64,7 +73,15 @@ class IdsButton extends IdsElement {
    */
   attributeChangedCallback(name, oldValue, newValue) {
     if (this.shouldUpdate) {
-      IdsElement.prototype.attributeChangedCallback.apply(this, [name, oldValue, newValue]);
+      switch (name) {
+        // Convert "tabindex" to "tabIndex"
```
tmcconechy, line 77: Is this for backwards compatibility? If so we dont really need to worry about that so you could just change it as we havent released anything yet? I maybe be misunderstanding this change

EdwardCoyle: Not for backwards compatibility, the `tabIndex` gets passed to the inner HTMLButtonElement, which is expecting the camelcase to set properly.  I was just transforming the attribute name here.

tmcconechy (approved):

ericangeles (approved):

PR #40, opened 2021-02-03
NA - Add improvements to types and build process

tmcconechy:

Added a few improvements to building/builds and the bundling sizes. This includes the following interesting changes:

1. The index.js in the demo app was putting code specified for the demos in the release bundle. To fix that i separated the index.js and example.js into two files. index.js should be ONLY the imports needed for webpack to build a usable component. example.js will contain the demo code now for example.html. All other examples by name are uneffected
2. Updated a few dependencies to the latest
3. Made a test release for prod and tested it over at https://github.com/infor-design/enterprise-wc-examples
4. Added a new prod build command `npm build:prod`
5. Fix a bug in virtual scroller that if you scroll up really fast to the top, it doesnt render

Bonus Unsolved Issue:
- go to https://github.com/infor-design/enterprise-wc-examples
```sh
cd typescript-ids-wc
npm i
npm run start
open http://127.0.0.1:8080 and check the console
```
Why is this error there? how do i fix

NA

- go to https://github.com/infor-design/enterprise-wc-examples and test each of the four projects (typescript one still has an issue loading in the browser)
- see that the examples still work as expected
- look at the size of the js files in dist
- then run `npm run prod:build` and check them again -> is now smaller
- note that in all the types corrected `export class IdsXxx` to export default class IdsXxx` this fixed some typing issues


EdwardCoyle (approved):
Still not sure what's up the WC-examples.  I tried changing the [`.tsconfig`](https://github.com/infor-design/enterprise-wc-examples/blob/master/typescript-ids-wc/tsconfig.json) settings to more closely match what's in this project but that didn't seem to help.  Maybe a webpack issue?

deep7102 (approved):

PR #43, opened 2021-02-08
4815 - Improve tabbable option

tmcconechy:

Looked over the tabbable options on clear an trigger fields. Also changed the validation icon to the newer icons as was intended but we never did it. It also looks better (i think anyways)

Adds some work on #4815 to WC

- go to http://localhost:4300/ids-input
- go to the clearable example and type something and then tab
- go to http://localhost:4300/ids-trigger-field
- tab through the fields - the first trigger is NOT tabbable the rest are so you should hit the trigger button on the last two fields

### DIFF
--- a/src/ids-trigger-button/ids-trigger-button.js
+++ b/src/ids-trigger-button/ids-trigger-button.js
@@ -5,6 +5,7 @@ import {
 
 // @ts-ignore
 import { IdsButton, BUTTON_PROPS } from '../ids-button/ids-button';
+import { IdsStringUtilsMixin as stringUtils } from '../ids-base/ids-string-utils-mixin';
 import { props } from '../ids-base/ids-constants';
 // @ts-ignore
 import styles from './ids-trigger-button.scss';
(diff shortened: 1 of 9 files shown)


src/ids-trigger-button/ids-trigger-button.js
```diff
@@ -27,8 +28,22 @@ class IdsTriggerButton extends IdsButton {
...
+  /**
+   * Set if the trigger field is tabbable
+   * @param {boolean|string} value True of false depending if the trigger field is tabbable
+   */
+  set tabbable(value) {
+    const isTabbable = stringUtils.stringToBool(value);
+    /** @type {any} */
+    const button = this.shadowRoot.querySelector('button');
+    this.setAttribute(props.TABBABLE, value.toString());
+    button.tabIndex = !isTabbable ? '-1' : '0';
```
EdwardCoyle, line 43: I was wondering about this setting before... is there any value to having this on top of just a standard `tabIndex`?  Was thinking since this is an extension of IDS Button, we could use the tabIndex prop there, which passes directly through to the [HTML button](https://github.com/infor-design/enterprise-wc/blob/master/src/ids-button/ids-button.js#L293-L319)

EdwardCoyle (commented):
Looks good to me, just had an implementation question below

src/ids-trigger-button/ids-trigger-button.js
tmcconechy: I'm not sure? I think the problem with that might mean we do have the attributeChangeEvents fire so we can only set a tabIndex on the root element with it. We cant go in the shadow DOM and set it on the shadowElement button.

EdwardCoyle: The button component's `tabIndex` does remove the attribute from the light DOM, and sets it on the Shadow DOM button underneath.  Focusing doesn't seem to be a problem with it from what I can tell.  Either way it's good, just wanted to make sure we didn't have competing functionality on the trigger buttons.

EdwardCoyle (approved):

deep7102 (approved):

PR #45, opened 2021-02-15
4718 - Add Ids Block Grid

ericangeles:

Implements responsive layout for IDS Block Grid Component. Some functionalities from the enterprise are not yet here like paging, single and mixed selection.

**Related GitHub/Jira issue (required)**:

Closes https://github.com/infor-design/enterprise/issues/4811

- Go to http://localhost:4300/ids-block-grid
- Go check on different breakpoints to test the layout.
- Check the code implementations, tests (functional, e2e) and docs.

~~- [ ] A note to the changelog.~~

### DIFF
--- a/src/ids-block-grid/ids-block-grid.scss
+++ b/src/ids-block-grid/ids-block-grid.scss
@@ -1,4 +1,3 @@
-@import '../../node_modules/ids-identity/dist/theme-uplift/tokens/web/theme-uplift-mixins.scss';
 @import '../ids-base/ids-base';
 
 ids-block-grid-item {
--- a/src/ids-block-grid/ids-block-grid.scss
+++ b/src/ids-block-grid/ids-block-grid.scss
@@ -8,11 +8,13 @@ ids-block-grid-item {
 }
 
 :host {
-  @include block();
   @include m-0();
 
-  max-width: 100%;
-  text-align: left;
+  align-items: center;
+  display: flex;
+  flex-wrap: wrap;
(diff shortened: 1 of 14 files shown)


tmcconechy:
Looks pretty good but a couple small comments...

1. If you make the browser wide - it looks like its not centered - can we make align="centered" the default.
2.Then after you do that.. You will see if you resize so 3 are on the first row. Wonder if there is a better way to lay it out? I see you used inline block but was thinking we would use css grid or flex in case of add numbered items I.E. can we do left align but also the whole thing centered within the page (but the tiles align left/right)

Visual image is ok but anticipating you will make a change so did not approve.

tmcconechy (dismissed):

src/ids-block-grid/ids-block-grid.scss
```diff
@@ -0,0 +1,38 @@
+@import '../../node_modules/ids-identity/dist/theme-uplift/tokens/web/theme-uplift-mixins.scss';
```
deep7102, line 1: Think you do not need these uplift mixins already imported in `src/ids-base/ids-base`.

deep7102 (commented):
Look good!,
are we going to do standalone-css example

tmcconechy (approved):
Looks great 👍🏻  We will need to merge https://github.com/infor-design/enterprise-wc/pull/47 with this. It might be easier if that goes in first and we merge it in.

I will help with that.

tmcconechy (approved):

PR #52, opened 2021-03-19
Implement Ids Toolbar / Enable usage of Ids Menu Button in shadow root

EdwardCoyle:
This PR makes a few additions:
- Fixes some issues in the construction of the `IdsPopupMenu` that prevented some of its components from communicating between each other when they were present inside a shadow root (infor-design/enterprise#4892).  A new library, `IdsDOMUtils`, has been added and includes some tools that enable traversal across Shadow DOM boundaries.
- Implements the `IdsToolbar`, `IdsToolbarSection`, and `IdsToolbarMoreActions` components. (infor-design/enterprise#4778)
- Adds most functionality from the Flex Toolbar into IdsToolbar (section types, alignments, API for accessing items/sections, disabled/tabbable on the Toolbar, keyboard handling).
- Tests/Examples/Docs/etc
- The Toolbar More Actions Button implementation demonstrates how a Menu Button works while in a shadow root

This PR doesn't yet implement the full functionality of the Overflow menu into the Toolbar's More Actions button.  We need to complete the work on IdsContainer, and add some additional improvements to the IdsPopup placement algorithm before that work can be completed.  The More Actions Button does support having pre-defined menu items, which is what all the samples will display for now.

- Closes infor-design/enterprise#4892
- Closes infor-design/enterprise#4778

- Pull, build, run
- Open http://localhost:4300/ids-toolbar
- Try navigation with the keyboard.  Arrow keys navigate.  Using Enter will open the Menu Buttons. The default example is exactly like the default 4.x, where only one item has a 0-or-more tabIndex at a time.  Also see http://localhost:4300/ids-toolbar/tabbable, which provides Tab/Shift+Tab as a way to navigate between items.
- Try using the More Actions button.  The base of the Menu Button structure exists in Shadow Root, but a slot is provided for passing in the contents of a Popup Menu.  It should not be possible to navigate the Toolbar while focus is on a Menu Button's menu item.
- See the disabled style:  http://localhost:4300/ids-toolbar/disabled - Toolbar contents shouldn't respond.  The text is currently not responding to changes in disabled state, but that should be fixed in an additional PR (@tmcconechy).
- http://localhost:4300/ids-toolbar/no-eyebrow shows a Toolbar sample without the additional "Eyebrow" text.
- http://localhost:4300/ids-toolbar/centered demonstrates a Toolbar with a center-aligned Buttonset element.
- http://localhost:4300/ids-toolbar/standalone-css shows the style only.

### DIFF
--- a/app/ids-toolbar/index.js
+++ b/app/ids-toolbar/index.js
@@ -1,10 +1,12 @@
 import IdsToolbar, {
-  IdsToolbarSection
+  IdsToolbarSection,
+  IdsToolbarMoreActions
 } from '../../src/ids-toolbar/ids-toolbar';
 
 // Supporting components
 import IdsButton from '../../src/ids-button/ids-button';
 import IdsInput from '../../src/ids-input/ids-input';
+import IdsMenuButton from '../../src/ids-menu-button/ids-menu-button';
 import IdsPopupMenu, {
   IdsMenuGroup,
   IdsMenuItem,
--- a/app/ids-toolbar/index.js
+++ b/app/ids-toolbar/index.js
@@ -2,15 +2,3 @@ import IdsToolbar, {
   IdsToolbarSection,
   IdsToolbarMoreActions
 } from '../../src/ids-toolbar/ids-toolbar';
-
-// Supporting components
-import IdsButton from '../../src/ids-button/ids-button';
-import IdsInput from '../../src/ids-input/ids-input';
-import IdsMenuButton from '../../src/ids-menu-button/ids-menu-button';
-import IdsPopupMenu, {
-  IdsMenuGroup,
-  IdsMenuItem,
-  IdsMenuHeader,
-  IdsSeparator
-} from '../../src/ids-popup-menu/ids-popup-menu';
-import IdsText from '../../src/ids-text/ids-text';
--- a/app/ids-toolbar/index.js
+++ b/app/ids-toolbar/index.js
@@ -11,3 +11,4 @@ import IdsPopupMenu, {
   IdsMenuHeader,
   IdsSeparator
 } from '../../src/ids-popup-menu/ids-popup-menu';
+import IdsText from '../../src/ids-text/ids-text';
--- a/app/ids-toolbar/index.js
+++ b/app/ids-toolbar/index.js
@@ -1,0 +1,13 @@
+import IdsToolbar, {
+  IdsToolbarSection
+} from '../../src/ids-toolbar/ids-toolbar';
+
+// Supporting components
+import IdsButton from '../../src/ids-button/ids-button';
+import IdsInput from '../../src/ids-input/ids-input';
+import IdsPopupMenu, {
+  IdsMenuGroup,
+  IdsMenuItem,
+  IdsMenuHeader,
+  IdsSeparator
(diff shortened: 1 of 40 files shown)


app/ids-toolbar/index.js
```diff
@@ -0,0 +1,16 @@
+import IdsToolbar, {
```
tmcconechy, line 1: This file should just have whats needed for IdsToolbar directly? Is all this needed? If not please put in example.js whats needed for the example

src/ids-toolbar/ids-toolbar-more-actions.js
```diff
@@ -0,0 +1,95 @@
...
+import IdsToolbarSection from './ids-toolbar-section';
+import IdsMenuButton from '../ids-menu-button/ids-menu-button';
+import IdsPopupMenu from '../ids-popup-menu/ids-popup-menu';
+
+/**
+ * IDS Toolbar Section Component
+ */
+@customElement('ids-toolbar-more-actions')
+@scss(styles)
+class IdsToolbarMoreActionsButton extends IdsElement {
```
tmcconechy, line 16:
```suggestion
class IdsToolbarMoreActionsButton extends IdsElement {
```
A bit verbose? Maybe IdsToolbarActions? Not sure i guess its fine if no one really uses this?

src/ids-toolbar/ids-toolbar-more-actions.js
```diff
@@ -0,0 +1,95 @@
...
+  constructor() {
+    super();
+  }
+
+  static get properties() {
```
tmcconechy, line 21: Possibly not needed

tmcconechy (commented):
Looks really great 👍🏻  Just some small comments is all i found. And one "issue" should it be expected that the menu opens the other way yet?

Screen Shot 2021-03-22 at 9 35 54 AM

src/ids-toolbar/ids-toolbar-more-actions.js
EdwardCoyle: I've opted for `IdsToolbarMoreActions` since the component itself isn't just a "button", and the actual "actions" are in the other Toolbar sections (this contains "more" actions).

src/ids-toolbar/ids-toolbar-more-actions.js
EdwardCoyle: I'll ditch it for now, this component will have props at some point though (was considering exposing the `data` property from the `Popup Menu` > `DataSource` here

app/ids-toolbar/index.js
EdwardCoyle: Yeah most of this is needed for the other components in the example, so I'll move those 👍

EdwardCoyle:
@tmcconechy I addressed those comments.  The submenu on the More Actions button can't be fixed until we add the additional placement logic to Popup

tmcconechy:
OK, do we want an issue for the placement thing then? Its what is preventing me from using it as a theme switcher.

tmcconechy (approved):

tmcconechy:
ok made https://github.com/infor-design/enterprise/issues/4990 -> ready to approve this now

deep7102 (approved):

PR #53, opened 2021-03-22
⬆️ Upgrade `@percy/puppeteer` to v2

Robdel12:
## What is this?

👋🏼 Hey-o! Percy SDK dev here. I've been cruising GitHub for OSS Percy projects to test our `@percy/migrate` package with. With that, I figured I'd PR the changes and get you up to the latest SDK toolchain :D (https://docs.percy.io/docs/migrating-to-percy-cli#js-sdk-migration)

(You'll likely want to check this branch out and push it to your repo, so CI runs with secrets)


tmcconechy (commented):
Thanks @Robdel12 - i actually had this in a branch thats coming very soon.
Looks like i did the same thing. So i will keep this for now and compare - and possibly close this one...

tmcconechy:
Actually i did have a question i cant resolve. With the new API it seems like its checking if the percy is available now and taking a few extra seconds. We only want to do it when we run one command.

Is there a way to tell the test not to even both to check for percy since we know for sure we dont want percy this time (we wont run it all the time for every dev for local tests as that counts too much on our snap shot count)

```
 PASS  test/ids-icon/ids-icon-func-test.js
[percy] Percy is not running, disabling snapshots
```

tmcconechy:
Actually will just close this - as i do have it covered in a coming branch looks pretty much the same. Thanks a lot of heping tho.

PR #54, opened 2021-03-25
Misc: Fix a looping issue on windows 

tmcconechy:
Fixes a looping issue on windows by only copying d.ts files on prod

NA

- npm run build -> then check the folders in the dist folder and there will be no d.ts files
- npm run build:prod -> then check the folders in the dist folder and there will be d.ts files
- pull this branch on windows
- then run `npm run start` -> the command should no longer constantly reload

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -138,7 +138,26 @@ module.exports = {
       chunks: ['index']
     }),
     // Show Style Lint Errors in the console and fail
-    new StylelintPlugin({})
+    new StylelintPlugin({}),
+    new CopyWebpackPlugin({
+      patterns: [
+        {
+          from: './src/**/*.scss',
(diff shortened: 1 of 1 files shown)


rob2d (approved):
@tmcconechy Copied from `enterprise` issue I had mis-replied to.

One thing, and may be superficial/subjective, is it might be easier to read webpack configs when they're more sequential; in this case I don't know if it's too hacky for this teams' taste but can also do something like:

```
module.exports = {
  entry: { ...someOptions },
  plugins: [
    PluginA,
    PluginB,
    ...(isProduction? [] : [CopyWebpackPlugin({ 
      opts 
    }])
  ]
}
```

deep7102 (approved):

PR #56, opened 2021-04-07
Ids text weight and default ellipsis overflow

rob2d:

- For `IdsText`, overflow text ellipsis was not possible as it is in a span which has a quirk in CSS. Added `overflow` property which defaults to `ellipsis` behavior. Specifying `clip` or `none` would remove this functionality.
- Also added a `font-weight` property as the component injects `font-weight: 400`. 

Aliased `extrabold` as `bolder` since `extrabold` can be confused with two words without hyphen but internally in our code does not have hyphen.

Needed to address infor-design/enterprise-wc#56

- pull branch `ids-text-weight-and-default-ellipsis-overflow` and run dev server locally via `npm run start`.
- visit http://localhost:4300/ and scroll to `TEXTS/TYPOGRAPHY`
- in Chrome Inspector, set `<ids-layout-grid-cell>` surrounding the text demo section to have `element.style` `max-width: 100px` and see that text ellipsis properly.
image
- edit an `ids-text` component to have font weight of `bold`. Notice that the text changes properly.
Kapture 2021-04-07 at 10 51 20

--------------------

+ also have a fix in this PR so that calling `this.render()` clears inner shadow DOM HTML if it existed already without duplicating markup. 

This allows us to use it predictably in lifecycle without having to add second `rerender()` methods when all we would like to do is modify the template.

Spoke with @EdwardCoyle on this change but if there's any problems it creates please let me know.

### DIFF
--- a/src/ids-base/ids-element.js
+++ b/src/ids-base/ids-element.js
@@ -80,6 +80,11 @@ class IdsElement extends HTMLElement {
 
     // Make template and shadow objects
     const template = document.createElement('template');
+
+    if (this.shadowRoot?.innerHTML) {
+      this.shadowRoot.innerHTML = '';
+    }
+
     if (!this.shadowRoot) {
       this.attachShadow({ mode: 'open' });
     }
--- a/src/ids-text/ids-text.js
+++ b/src/ids-text/ids-text.js
@@ -70,7 +70,6 @@ class IdsText extends IdsElement {
   set fontSize(value) {
     if (value) {
       this.setAttribute(props.FONT_SIZE, value);
-      this.container.classList.add(`ids-text-${value}`);
       this.render();
       return;
     }
--- a/src/ids-text/ids-text.js
+++ b/src/ids-text/ids-text.js
@@ -28,6 +28,7 @@ class IdsText extends IdsElement {
     return [
       props.TYPE,
       props.FONT_SIZE,
+      props.FONT_WEIGHT,
       props.AUDIBLE,
       'overflow'
     ];
@@ -40,20 +41,15 @@ class IdsText extends IdsElement {
   template() {
     const tag = this.type || 'span';
     let classList = 'ids-text';
-    classList += (
-      this.overflow == 'ellipsis' || !this.overflow
-    ) ? ' ellipsis' : '';
+    classList += (this.overflow === 'ellipsis') ? ' ellipsis' : '';
     classList += this.audible ? ' audible' : '';
     classList += this.fontSize ? ` ids-text-${this.fontSize}` : '';
 
     // @ts-ignore
     switch (this.fontWeight) {
-    case 'bold': {
-      classList += 'ids-text-bold';
-      break;
-    }
+    case 'bold':
     case 'bolder': {
-      classList += 'ids-text-bolder';
+      classList += ` ${this.fontWeight}`;
       break;
     }
     default: {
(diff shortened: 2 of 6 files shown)


src/ids-text/ids-text.js
```diff
@@ -61,30 +70,42 @@ class IdsText extends IdsElement {
...
 
   get fontSize() { return this.getAttribute(props.FONT_SIZE); }
 
+  /**
+   * Adjust font weight; can be either "normal", "bold" or "bolder"
+   * @param {string} [value='normal'] font weight
+   */
+  set fontWeight(value) {
+    this.setAttribute(props.FONT_WEIGHT, value || 'normal');
+    this.render();
```
tmcconechy, line 89: rerender/render is a bit more expensive. The best is if we can do this with a class change. Wouldn't that work here? Im thinking this is only needed if you need to change the node type. The rest you can toggle a class or something simpler?

src/ids-text/ids-text.js
```diff
@@ -96,14 +117,27 @@ class IdsText extends IdsElement {
...
+    return this.getAttribute('overflow') || 'ellipsis';
+  }
+
+  /**
+   * Set how content overflows; can specify 'ellipsis' or 'clip'
+   * @param {string} [value='ellipsis'] how content is overflow
+   */
+  set overflow(value) {
+    this.setAttribute('overflow', value || 'ellipsis');
+    this.render();
```
tmcconechy, line 139: Seem here maybe we can just toggle the class?

tmcconechy (changes requested):
Some small edits

- unsure these new props need a full rerender
- in the example maybe just leave the two examples out and put them in a test instead https://github.com/infor-design/enterprise-wc/blob/main/test/ids-text/ids-label-func-test.js
- also noticed the test is called label (old name) can rename that to text
- the PR will eventually likely report coverage dropped so need a test for the new settings 

Otherwise looks good.

src/ids-base/ids-element.js
```diff
@@ -80,6 +80,11 @@ class IdsElement extends HTMLElement {
 
     // Make template and shadow objects
     const template = document.createElement('template');
+
+    if (this.shadowRoot?.innerHTML) {
```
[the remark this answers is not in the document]
rob2d: @tmcconechy yeah actually -- the innerHtml on `shadowRoot` is for current element, so it exists and creates duplicates in lifecycle in many cases. Added this since it was creating issues on the WIP `ids-wizard` branch.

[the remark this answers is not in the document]
tmcconechy: Ok dont think it would hurt... But if its related to using the full re render method to update we want to try and minimize that for cases like you need to change ALL the markup. (comments below)

src/ids-text/ids-text.js
rob2d: will look into this -- was trying to follow the 80/20 rule since the pattern of manually editing classes seems error-prone vs the payoff in this sort of thing (since the browser still needs to repaint the component... I think may be more expensive than our template method if the logic isn't crazy). But will try to just change the class list on the element directly in this csae.

tmcconechy: Dont see any reason adding and removing classes should be error prone. But that was what i was thinking is that you really dont need to remove all the html and add it again. When you can just toggle a class.

Rule is do the minimum you can to make the property active and its a pattern we have been following (maybe in the end its about the same but it seems logical that toggling a class/attribute would be simpler to me then removing it all and adding it again).

rob2d:
Closing to push onto new branch to clean the merges from commit log -- have fixes for the test that was breaking so that we can work with jest mock a little more safely also.

PR #57, opened 2021-04-08
IdsElement ::host support and mock test decoupling

rob2d:

There were potential issues with `::host` (vs `:host`) not tested that make IdsElement unpredictable in some use cases, since this is a criteria. So added support for that small variant + de-hardcoded/coupled stylemock from some related tests that were also having issue there on another branch.

Pushing this here separately as not to further bloat upcoming PRs.

Related to work on enterprise/issues/5026 -- but can help others not figure out what went wrong on tests in future scenarios.

- pull `ids-element-host-css-and-mock-iteration`
- `npm run test -- ids-base-func-test`

Should work as expected.

### DIFF
--- a/src/ids-base/ids-element.js
+++ b/src/ids-base/ids-element.js
@@ -112,8 +112,10 @@ class IdsElement extends HTMLElement {
       const style = document.createElement('style');
       // @ts-ignore
       style.textContent = this.cssStyles;
-      if (style.textContent?.indexOf(':host') === 0) {
-        style.textContent = style.textContent.replace(':host', `.${this.name}`);
+      // @ts-ignore
+      if (/^:?:host/.test(style.textContent)) {
+        // @ts-ignore
+        style.textContent = style.textContent.replace(/^:?:host/, `.${this.name}`);
       }
       style.setAttribute('nonce', '0a59a005'); // TODO: Make this a setting
       this.shadowRoot?.appendChild(style);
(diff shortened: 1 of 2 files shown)


rob2d:
Example of why code-coverage == 100% is not a good barometer for builds passing in this PR: 

simple change for quality of test increased => coveraged decreased.

rob2d:
Things I tried: 

- double-checked logic
- debugged tests, 
- ran locally. 
- changed logic for regex it fails on to match "absolutely anything or nothing @ IdsElement.js => LN 116 " (`[\s\S]*`)

It seems to be a bug in coveralls reporting here  @ `IdsElement.js` since `:host`, and `::host` are set on the `ids-tag` and clearly converted to `.ids-tag`. 

So it's not running through the conditional found in `IdsElement.appendStyles` for some reason or another even though it clearly replaces content, and then does not report that it went through this.

Seems unideal/unoptimal but I'll just do as we did before PR and `.replace()` with tag name on the entire stylesheet string without conditionals.

rob2d:
With some help from Ed to understand the coverage tool, got it solved: 

Seems that the Jest environment was already mimicking Firefox behavior with (missing) support for some W3C things vs the inverse as we may have thought. 

After setting cssStyles to end-result before re-render() manually, coverage is 100% now.

tmcconechy (dismissed):
Looks good - assuming checks pass...

tmcconechy (dismissed):

tmcconechy (dismissed):

tmcconechy (approved):

deep7102 (approved):

PR #70, opened 2021-04-12
ids-events-mixin: fix leak when adding events

rob2d:

- When adding events to the `ids-elements-mixin`, we were adding a potential memory leak. This is because in `connectedCallback`, we call `handleEvents`, but only detaches events on `disconnectedCallback`. If we change the elements in ShadowDOM and handle another event but never remove the element from the DOM, it's possible to keep adding subs before the element finally disconnects, even though the only accessible sub is the one on the event namespace. If an object persists on the page and does this it can keep adding up.

Discovered this adding events to wizard component which showed this behavior. It is possible to work around by always calling `offEvent` before subbing via `onEvent` but safe to not attach events that may be copied over until element is removed.

Other minor changes along the way of troubleshooting this issue:
- since there's a map `handledEvents`, we can actually access things via `has`/`get` which is O(1), so minor performance bump on `Map.set` to filter for the event name when there may be a lot of listeners.
- Proposal: `detachEventName` => `detachEventsByName`. This is just because if we ever support multiple events (using `on`/`off` paradigm in jQuery which it seems we may be going for), then this wouldn't become an issue to refactor the codebase.  We may also need to support multiple events as there's no way to do that now, and it wouldn't create another memory issue or consideration later (not in scope of this bug fix, just was considering it as it was tiny). [I can revert depending on what any reviewers think of course, too]

Unfortunately there are no scenarios directly in this behavior right now and a bit out of the way of direct tickets to work on -- we do not have calls for render() outside of `ids-text`, but it is a paradigm we use and also used in `ids-wizard` as the markup was complex and needed a `render()` call to refresh which we briefly mentioned in team chat today. Should hopefully be simple enough to see the behavior as this is a minimal change to fix.


rob2d:
built off of wrong branch -- will re-open issue

PR #71, opened 2021-04-12
ids-events-mixin: fix potential leak and minor changes described in PR

rob2d:

When adding events to the `ids-elements-mixin`, there was a small potential memory leak for an element that stays active but re-`render()` ing. This is because in `connectedCallback,` we call `handleEvents`, but only detaches events on `disconnectedCallback` in base `IdsElement`. If we change the elements in ShadowDOM and handle another event but never remove the element from the DOM, it's possible to keep adding subs before the element finally disconnects, even though the only accessible sub is the one on the event namespace. If an object persists on the page and does this it can keep adding up.

Discovered this adding events to wizard component which showed this behavior. It is possible to work around by always calling offEvent before subbing via onEvent but safe to not attach events that may be copied over until element is removed.

Other minor changes along the way of troubleshooting this issue:

- since there's a map `handledEvents`, we can actually access things via has/get which is O(1), so minor performance bump on Map.set to filter for the event name when there may be a lot of listeners.
- Proposal: `detachEventName` => `detachEventsByName`. This is just because if we ever support multiple events (using on/off paradigm in jQuery which it seems we may be going for), then this wouldn't become an issue to refactor the codebase. We may also need to support multiple events as there's no way to do that now, and it wouldn't create another memory issue or consideration later (not in scope of this bug fix, just was considering it as it was tiny). [I can revert depending on what any reviewers think of course, too]

Relates to infor-design/enterprise#5026

Unfortunately there are no scenarios directly in this behavior right now and a bit out of the way of direct tickets to go back to and trigger it -- we do not have calls for `render()` outside of ids-text, but it is a paradigm we use and also used in ids-wizard as the markup was complex and needed a `render()` call to refresh which we briefly mentioned in team chat today. Should hopefully be simple enough to see the code as it is minimal (if code coverage passes).


rob2d:
calling `offEvent` before `onEvent` seems to cause some issue with some mixin code -- this is a bit odd as it seems to be synchronous, so I'm guessing there may be some quirk in the way we're adding native event handles in our mixins (e.g. clearables, etc). Started to debug to rule out it was code added, which it does not seem to be -- but its a bit of a rabbit hole so will try to get to other things for the rest of the evening and early tomorrow, and continue.

Otherwise: if anyone else is more familiar with this in memory and has an idea feel free to chime in.

rob2d:
Just leaving a note: one possibility is it's not very natural to add then manually trigger/dispatch an event immediately in same scope (since we're just synthesizing it after all). 

Could try adding a setTimeout to test this otherwise when revisiting, and hopefully that is acceptable since it's in not-too-practical test code, whereas in real browser scenario there shouldn't be a simulation of dispatching events like this.

rob2d:
Found the issue: some functions were called that belong to the class that werent binded to "this".  So when we called in a specific scope, was having issues after trying to call offEvent() then onEvent().

There's a nice syntax to solve this, and in some particular components where we want `this` to be immutable such as event mixins particularly, could be good practice.
```js
class X {
     someCallback = () => {
     };
}
```

which is just syntactic sugar for:
```
class X {
     constructor() {
          this.someCallback = this.someCallback.bind(this);
     }

     someCallback() {
     }
}
```

src/ids-base/ids-events-mixin.js
```diff
@@ -18,58 +18,59 @@ const IdsEventsMixin = (superclass) => class extends superclass {
...
     const event = new CustomEvent(eventName.split('.')[0], options);
     target.dispatchEvent(event);
-  }
+  };
 
   /**
    * Detach all event handlers
    */
-  detachAllEvents() {
+  detachAllEvents = () => {
```
tmcconechy, line 54:
```suggestion
  detachAllEvents = () => {
```

Lets keep the original syntax. I think that way it will be more consistent
ets k

tmcconechy (commented):
My suggestion/vote is we keep the current format of the functions for consistency in the code base.

src/ids-base/ids-events-mixin.js
rob2d: put a note on why the change in the comments. Basically: on event listener mixins specifically we're always aiming to subscribe to `this` (on Element) but calling the function from other scopes than the object (or even possibly anonymous) was causing issues in some conditions. It's what was causing the test to fail after the small logical change, so that's why I had adjusted.

rob2d: adjusted to use bind in constructor way of doing that

tmcconechy (approved):

deep7102 (approved):

EdwardCoyle (approved):
I was sort of able to to test this one:

- Open dev tools.  In the console use `$('ids-popup-menu').handledEvents`.  It should show 6 entries -- the last one is `click.trigger`
- In the elements panel, find the first menu button and change its `trigger` attribute to `contextmenu`.
- Now try and open the menu button.  It won't open with left click, but will open with right click (this is a glitchy scenario, but it shows the events have been removed/rebound)
- Check `$('ids-popup-menu').handledEvents`.  It should still have 6 entries, but the last one is now `contextmenu.trigger`

Can confirm, looks like it's removing events as expected.  Thanks for this @rob2d!  :banana:

rob2d:
@EdwardCoyle nice thanks for the thorough testing/practical scenario Ed! 🍌

Have another small fix where we just check to call `offEvent` before `onEvent` as it's not expensive to do so; but this is just for the sake of automatically cleaning any leaks if we `render()` to refresh. That is only until we support multi-subs to one event (solution is trivial as we discussed but out of scope). 

Pushed ^that 2-liner-fix on ids-wizard branch though since it (theoretically) shouldn't affect any current components.

PR #82, opened 2021-04-14
4973 - Add ids-counts Component

chrisakinrinade:
Adding IDS Counts component to the web component repo. Currently supports hyperlinked, compact, and disabled counts.

Fixes https://github.com/infor-design/enterprise/issues/4973

1. Pull this branch, build, and run the app
2. Go to http://localhost:4300/ids-counts
3. Check the code implementations and functional tests.

- [x An e2e or functional test for the bug or feature.


app/ids-counts/example-default.html
```diff
@@ -0,0 +1,35 @@
...
+  </ids-layout-grid-cell>
+  <ids-layout-grid-cell>
+    <ids-counts color="success" href="#">
+      <div slot="value">2</div>
+      <div slot="text">Open <br /> Incidents</div>
+    </ids-counts>
+  </ids-layout-grid-cell>
+  <ids-layout-grid-cell>
+    <ids-counts color="caution" href="#">
+      <div slot="value">4</div>
```
tmcconechy, line 19: Suggestion: Would a span be a better choice here that a div for the two slot elements?

app/ids-counts/example.html
```diff
@@ -0,0 +1,3 @@
+{{> example-default.html }}
+{{> example-not-actionable.html }}
```
tmcconechy, line 2: Can probably just show the default one here and leave the others for sub pages (for example simplicity). This might mean removing example-default and just putting in in example.

app/ids-counts/example-not-actionable.html
```diff
@@ -0,0 +1,35 @@
+<ids-layout-grid>
+  <ids-text font-size="12" type="h1">Counts: Disabled</ids-text>
+</ids-layout-grid>
+<ids-layout-grid cols="5" gap="md">
+  <ids-layout-grid-cell>
+    <ids-counts disabled="true">
```
tmcconechy, line 6:
```suggestion
    <ids-counts disabled="true">
```

i think maybe disabled is the wrong word here. Its not disabled looking its just not linked. Maybe we can do this with just the presence of a `href` or not instead of making a disabled property as that would imply to me it looks disabled..

app/ids-counts/example-short.html
```diff
@@ -0,0 +1,35 @@
+<ids-layout-grid>
+  <ids-text font-size="12" type="h1">Counts: Short</ids-text>
+</ids-layout-grid>
+<ids-layout-grid cols="5" gap="md">
+  <ids-layout-grid-cell>
+    <ids-counts short="true">
```
tmcconechy, line 6: Was going to start calling this compact https://github.com/infor-design/enterprise/issues/5077#issue-857302801
Could we change the short setting to compact to match up?

app/index.html
```diff
@@ -1,4 +1,5 @@
 {{> layouts/head.html title=htmlWebpackPlugin.options.title }}
+{{> ids-counts/example.html }}
```
tmcconechy, line 2: Can you put this down at the bottom instead of the first one (and as above we can just show the default)

app/index.js
```diff
@@ -1,4 +1,5 @@
 // Import All Components Used in the Page
+import IdsCounts from '../src/ids-counts/ids-counts';
```
tmcconechy, line 2: Can also move this down to the bottom - but that doesn't effect the the kitchen sink example. So not that important.

src/ids-counts/ids-counts.d.ts
```diff
@@ -0,0 +1,9 @@
+// Ids is a JavaScript project, but we define TypeScript declarations so we can
+// confirm our code is type safe, and to support TypeScript users.
+
+import { IdsElement } from '../ids-base/ids-element';
+
+export default class IdsCounts extends IdsElement {
+  /** Set the tag type/color */
```
tmcconechy, line 7: Seems to be missing the new settings in the component (looks like just copied from the tag)

src/ids-counts/ids-counts.js
```diff
@@ -0,0 +1,103 @@
...
+@customElement('ids-counts')
+@scss(styles)
+class IdsCounts extends IdsElement {
+  constructor() {
+    super();
+  }
+
+  /**
+   * Template helper function for creating ShadowDOM tags
+   * @returns {number} the font size
```
tmcconechy, line 36:
```suggestion
   * @private
   * @returns {number} the font size
```

tmcconechy: If this is private? Sometimes we use the full wording on these for example numberSize

src/ids-counts/ids-counts.js
```diff
@@ -0,0 +1,103 @@
...
+  get numSize() {
+    return this.getAttribute('short') !== 'false' ? 32 : 40;
+  }
+
+  /**
+   * Return the properties we handle as getters/setters
+   * @returns {Array} The properties in an array
+   */
+  static get properties() {
+    return [props.COLOR, props.DISABLED];
```
tmcconechy, line 47: Add these both  to d.ts?

src/ids-counts/ids-counts.js
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  /**
+   * Template helper function for creating ShadowDOM tags
+   * @returns {object} Object containing the outer HTML to be rendered
```
tmcconechy, line 52: Private?

src/ids-counts/ids-counts.js
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  /**
+   * Inner template contents
+   * @returns {string} The template
+   */
+  template() {
+    const { openingTag, closingTag, innerTags } = this.generateTags();
+
+    return `
```
tmcconechy, line 72: This is ok - but in many cases we just had the string template as a one expression with Template Literals all in the template. https://github.com/infor-design/enterprise-wc/blob/main/src/ids-accordion/ids-accordion.js#L31

Might simplify it a bit as this is rather simple...

src/ids-counts/ids-counts.js
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  /**
+   * Set the color of the tag
+   * @param {string} value The color value, this can be not provided,
+   * secondary (white), error, success, danger, caution or a hex code with the #
+   */
+  set color(value) {
+    const prop = value[0] === '#' ? value : `var(--ids-color-status-${value === 'error' ? 'danger' : value})`;
+    this.container.style.color = prop || '';
```
tmcconechy, line 86: As a test case - what happens if they clear the color? Did not test it but something to think about

src/ids-counts/ids-counts.js
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  /**
+   * Sets to disabled
+   * @param {boolean|string?} value If true will set `disabled` attribute
+   */
+  set disabled(value) {
```
tmcconechy, line 93: As per above i think disabled isnt the correct name for what this is doing. Maybe we can try doing it off of having a href or not

tmcconechy (changes requested):
Looks really great for a first PR 🍌  Nice Job.
Im just adding a few suggestions. Also you could add an e2e test like the others?

Im soon merging a PR where it will mean we have to add themes here as well. But thats not merged so dont worry about that until it is and i'll let you know. So just a heads up for now

app/ids-counts/example-default.html
chrisakinrinade: Possibly, what's your thought behind that? I wanted to ensure they wound up on different lines @tmcconechy

app/ids-counts/example-not-actionable.html
chrisakinrinade: I don't know why I thought the issue description explicitly mentioned "disabled" when, because it didn't. My mistake, I was actually thinking the same thing, the presence of an href is all you need.

app/ids-counts/example-default.html
tmcconechy: Yeah thats why i wondered - if you needed the div to make it different lines. 

Originally, my thinking is that divs are for sections/divisions in the document but spans are more for text.

But thinking again when its in the slot the could technically add anything... For example maybe they put a span or ids-text or ids-icon in there for some use case. So maybe we should make it be as flexible as we can - which is possibly more important then the span vs div. So making each slot a seperate line no matter what it is could be the answer there.

If we dont let them add data in the slot and "lock it down" we could use settings instead but having some options like you did it seems more flexible

app/ids-counts/example-short.html
chrisakinrinade: sure thing!

app/ids-counts/example.html
chrisakinrinade: will do

app/index.html
chrisakinrinade: yup! do all new components go at the bottom by default? wasn't sure where to put it.

app/index.js
chrisakinrinade: will do

app/index.html
tmcconechy: Yeah pretty much but its sort of free form for example i might put a new button next to the other buttons or the checkbox next to the radio ect. So more like where ever it fits with something similar if anything not then bottom. 

I'm thinking we wont keep this page for long as it gets a bit much. Probably replace it with a list like https://master-enterprise.demo.design.infor.com/components/list with a description, link ect eventually but looking for a design first

EdwardCoyle (commented):
Looks good so far @chrisakinrinade.  I've got nothing to add to Tim's comments.  Will approve once you post changes 👍

app/ids-counts/example-default.html
chrisakinrinade: Good point about the user injecting additional elements into the slots!

app/index.html
chrisakinrinade: Got it, I'm remembering now that you mentioned earlier about converting to a list.

deep7102:
This looks very nice, only saw
- the value slot `normal` and `compact` both are on 32px.
- are we adding `compact` and `href` to `properties()` array and setter so live change, currently saw only use by attributes
- will coming back to add  the `standalone-css` and `README.md` (no docs)

tmcconechy:
@chrisakinrinade - I merged in a big PR so you might have to fix a conflict when you apply main on.
Also it adds another task for this issue. Since i implemented a theme mixin you can now make this component have the themes

https://github.com/infor-design/enterprise-wc/blob/main/src/ids-base/README.md#ids-theme-mixin <- try and follow that and look at the components. And ask me if any questions or that doesnt make any sense

chrisakinrinade:
> This looks very nice, only saw
> 
> * the value slot `normal` and `compact` both are on 32px.
> * are we adding `compact` and `href` to `properties()` array and setter so live change, currently saw only use by attributes
> * be good if you could take a first pass at adding the `standalone-css` example and `README.md` (similar to the other components with DOCS)

Question on the compact and href attributes -- should those be added to ids-constants.js?

tmcconechy:
@chrisakinrinade - yes generally we want to list all possible settings in the consts. So if it has a setter/getter and its used in properties should just add it in the consts..

src/ids-counts/ids-counts.d.ts
chrisakinrinade: Ok, I've added these now

tmcconechy:
@chrisakinrinade only a couple more things that should be easy to fix.

1) for the tests i split it in https://github.com/infor-design/enterprise-wc/tree/main/test/ids-badge two. ids-counts-percy-test.js for just the percy tests and other e2e tests in ids-counts-e2e.js
1) notice in the percy test the import changed - this is why your tests are failing
1) did you add the themes? im not seeing that working when i change the theme switcher to dark mode an

tmcconechy:
@chrisakinrinade is this done and ready for re-review? If so i think just some tests need to be updated..

chrisakinrinade:
> @chrisakinrinade is this done and ready for re-review? If so i think just some tests need to be updated..

@tmcconechy I was waiting to see if it passed the checks. I noticed the axe test keeps timing and I'm not sure what's wrong with it. Would you happen to know if there's something simple it's missing? The error it gives is that the async call never gets a response.

tmcconechy (changes requested):
Tried to fix a few things for you but didnt get it all. I think the following should be fixed

1. The focus state on the default counts
2. The linkable counts should use the azure color (same as ids-hyperlink and you could try using that directly).
3. The non linkable should be the text color as you have it note that color="base" redundant instead should be just 2/3
4. the color setting is no longer working (worked at one point) see http://localhost:4300/ids-counts/compact
5. the standalone css example isnt 100% http://localhost:4300/ids-counts/standalone-css
6. Make it so you can reorder the two slot items to make the text vs count on top/bottom. Didnt realize this was a feature

I'm looking at the axe test failure so will PR that... separate and you can pull it in

tmcconechy:
Closing this for now - but when its done you can repopen the PR

PR #90, opened 2021-04-20
5112 - Implement Shared Mixin Props

EdwardCoyle:
This PR is an example of how we could solve the issue of sharing specific properties from the mixins up the prototype chain, better separating out the concerns of the mixin from the base classes.  In this PR, the `IdsThemeMixin`'s properties have been moved into the theme mixin code, and are shared via concatenation to the `IdsButton` base class, which looks for `super.properties` and concatenates its own properties there.

This reduced case is basically what I showed in our 4/20/2021 call, but applied to something already in the codebase.  We can discuss/change/etc here.

Closes infor-design/enterprise#5112

- Pull, build, run
- Open http://localhost:4300/ids-button
- Open dev tools and inspect the first button element.  Change its `mode` property to `contrast`.  the button color should change, responding to the altering the property defined in the mixin.  Also change `version` to `classic` to see that property change.

*For checking performance: Open http://localhost:4300/ids-button/performance*
This page has 1000 buttons loaded into the DOM after a `DOMContentLoaded` event is fired.  After the button render completes, a console message displays with a reading from `window.performance.now()`.  I was also testing this with the performance tab in Dev Tools, but both are showing on average that the spread syntax is slightly faster here (maybe on average 300-500ms).

Note that the theme mixin still works throughout other components here because all the other components still contain `mode` and `version` in their own properties arrays, and will need to be modified to actually accept the shared properties if we choose this route.

cc @lucacolumbu


src/ids-base/ids-theme-mixin.js
```diff
@@ -8,6 +10,23 @@ const IdsThemeMixin = (superclass) => class extends superclass {
     super();
   }
 
+  /*
+  static get properties() {
```
tmcconechy, line 14: Should we remove this?

src/ids-base/ids-theme-mixin.js
```diff
@@ -8,6 +10,23 @@ const IdsThemeMixin = (superclass) => class extends superclass {
...
+    return super.properties.concat([
+      props.MODE,
+      props.VERSION
+    ]);
+  }
+  */
+
+  static get properties() {
+    return [
+      ...super.properties,
```
tmcconechy, line 24: Like this a lot!

tmcconechy (commented):
Looks perfect. Just wondering if:

- we want to add the IdsEventsMixin to inherit? if we do i wonder if they imports should be removed in the components that only need themes and dont have events themeselves.. Realize that could be out of scrop
- when should we change the rest?

EdwardCoyle:
@tmcconechy 

- let's maybe change the events part outside of this, and scope this specifically to changing the way props are shared.
- If we get consensus, maybe change all the existing mixins in this PR to make the system uniform

tmcconechy (approved):

deep7102 (approved):

PR #98, opened 2021-04-22
NA - Fix e2e / percy test

tmcconechy:

Trying to fix a test instability and speed by:
- releasing a cache
-  removing byPassCSP
- adding `npm ci`
- adding a simple express server to serve the tests

NA


rob2d:
(1) only actual-change-suggestion: on the Node server, we should probably keep the `port` as a variable via env as we did before with fallback to `4444` e.g.
```
const port = process.env.PORT || 4444;
```

If a user doesn't set their environment variable in their terminal session, then they would start on 4444 so running `node server.js` then works the same in CI w/o config or extra params (can see screen of terminal verifying this)

image

(2) just a note that we may only need `networkidle2` in the *-percy.test.js` files (vs `networkidle0`) and not everywhere, but the failings were intermittent anyway so we'll just have to find out eventually either way and this is passing so LGTM.

Besides that these look like really nice changes (especially the fact that CI time goes down) 🍌

tmcconechy:
@rob2d good suggestion. thanks i added the ability to set the port on the command line (although the tests are running on 4444 it could still come in handy...)

PR #101, opened 2021-04-22
5049 - Typescript Typings Support iteration/cleanup

rob2d:

In our local development environment for WC, we are using both Typescript "classes" and Javascript "classes" within the same classname/namespace. We are also performing strict typescript checking in pure JavaScript and to prevent issues working around namespace/var resolution via `@ts-ignore` comments.

Since there's a fundamental compatibility between using the same d.ts class and javascript with native ES7 classes that share names, we have to disable strict type-checking within JS specifically. This does not interfere with detection of `d.ts` externally (demonstrated below).

This does not fully resolve infor-design/enterprise#5049 , but it allows us to test our repo externally with the other test projects non-theoretically + iteratively.

Partially addresses infor-design/enterprise#5049.

Will also attempt to create an external demo/code sandbox to test our React and Angular projects, as well as vet out further issues on another branch since this would be much easier once we have this change pushed to the repo. This likely is also the only opportunity to do this while other apps are not consuming it in production.

- pull `5049-ts-typings-support-iteration` locally
- open VSCode (and with ESLint plug and other typical dev env set up)
- Type Cmd + P (Mac) or Cntrl + P (Windows) and look up `ids-checkbox.js` @ `src/ids-checkbox/ids-checkbox.js`
- Copy and paste the following in lines 19-20
```
const idsText = new IdsText();
idsText.type = 'h1';
```
(1) hover over IdsText in import @ ln 17 and see the content is taken from the `.d.ts` and not the `.js` file.
(2) while hovering on IdsText, press Cmd and notice the source code is the `ids-text.d.ts` and not `ids-text.js`.
(3) hover over `idsText.type` and see the preview is from the `ids-text.d.ts` file and not the JSDoc in the `.js` file.

(and of course that all test pass here)

typescript-in-local-dev

### DIFF
--- a/src/ids-text/ids-text.d.ts
+++ b/src/ids-text/ids-text.d.ts
@@ -1,6 +1,9 @@
 // Ids is a JavaScript project, but we define TypeScript declarations so we can
 // confirm our code is type safe, and to support TypeScript users.
 
+/**
+ * a segment of text with standardized settings, theming and fonts
+ */
 export default class IdsText extends HTMLElement {
   /** Set the type of element it is (h1-h6, span (default)) */
   type: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'span' | string | null;
--- a/src/ids-text/ids-text.d.ts
+++ b/src/ids-text/ids-text.d.ts
@@ -1,9 +1,7 @@
 // Ids is a JavaScript project, but we define TypeScript declarations so we can
 // confirm our code is type safe, and to support TypeScript users.
 
-/** A segment of text with standardized settings, theming and fonts*/
- * a segment of text with standardized settings, theming and fonts
- */
+/** A segment of text with standardized settings, theming and fonts */
 export default class IdsText extends HTMLElement {
   /** Set the type of element it is (h1-h6, span (default)) */
   type: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'span' | string | null;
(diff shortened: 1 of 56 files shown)


src/ids-text/ids-text.d.ts
```diff
@@ -1,6 +1,9 @@
 // Ids is a JavaScript project, but we define TypeScript declarations so we can
 // confirm our code is type safe, and to support TypeScript users.
 
+/**
```
tmcconechy, line 4:
```suggestion
/** A segment of text with standardized settings, theming and fonts*/
```

Been doing these as oneliners...

tmcconechy (commented):
Just one comment but want to try it out tomorrow...

src/ids-text/ids-text.d.ts
rob2d: wasn't aware that it was possible but since JSDoc/VSCode seem to pick it up then makes sense!

tmcconechy:
@rob2d this looks good to me 👍🏻  ready to approve but there is one test failing/ error /parse error in here: https://github.com/infor-design/enterprise-wc/pull/101/checks?check_run_id=2419976543

rob2d:
@tmcconechy just noticed that and fixed thanks -- seemed to be from a merge conflict mistake at some point on GH's part with the suggestion.

tmcconechy (approved):

EdwardCoyle (approved):
Looks great!  The IDE hints will be really helpful here, thanks for this!

PR #102, opened 2021-04-27
4972 - ids rating wc 

danielortiz1982:
Porting and converting the enterprise jQuery rating component into an IDS Web Component. Creating a custom element <ids-rating> web component with options to customize the stars, readonly, and full/half values. When writable as the stars are clicked the value setAttribute will change to reflect the updated star count. 

Closes https://github.com/infor-design/enterprise/issues/4972

- Checkout the feature branch - _**git checkout 4972-ids-rating-wc**_
- Pull the latest changes - _**git pull -f**_
- Create a sample html document inside of the ids-rating folder - _**./app/ids-rating**_
- Create the custom element by using the markup - _**`<ids-rating></ids-rating>`**_
- Navigate to the ratings wc url - _**http://localhost:4300/ids-rating**_
- Click on the given stars and check for the update value in the web inspect.
Screen Shot 2021-04-28 at 8 59 22 AM
Screen Shot 2021-04-28 at 9 00 34 AM
Screen Shot 2021-04-28 at 9 01 36 AM

**Pending completion in this Pull Request**:
- e2e fn testing - current coverage 26%
- Percy testing
- ReadMe Docs


tmcconechy:
Before we review can you
- update the PR template (make sure to include what isnt done yet and anything you need us to help look at)
- can also start by fixing the checks...

BTW converted to WIP

app/ids-rating/example.html
```diff
@@ -0,0 +1,8 @@
+<ids-layout-grid>
+    <ids-layout-grid-cell>
+        <ids-text><h2>Ratings</h2></ids-text>
```
tmcconechy, line 3:
```suggestion
        <ids-text><h2>Ratings</h2></ids-text>
```

This should be ` <ids-text font-size="12" type="h1">Alerts</ids-text>` as the label/heading isnt styled right like other examples

src/ids-rating/README.MD
```diff
@@ -0,0 +1 @@
+# ids rating wc
```
tmcconechy, line 1: Example for when you do this https://github.com/infor-design/enterprise-wc/blob/main/src/ids-hyperlink/README.MD

Also some snippets from the old site
```
Displays a scale of selectable values in ascending order. A user can select exactly one value at a time. Best for showing a user's current selection relative to the lower and upper limits of a scale.
```

and https://design.infor.com/product/components/rating

src/ids-rating/ids-rating.js
```diff
@@ -0,0 +1,209 @@
...
+import styles from './ids-rating.scss';
+
+/**
+ * IDS Rating Component
+ * @type {IdsRating}
+ * @inherits IdsElement
+ * @mixes IdsEventsMixin
+ * @mixes IdsKeyboardMixin
+ */
+
```
tmcconechy, line 21:
```suggestion
```

Extra space?

src/ids-rating/ids-rating.js
```diff
@@ -0,0 +1,209 @@
...
+ @customElement('ids-rating')
+ @scss(styles)
+
+class IdsRating extends mix(IdsElement).with(IdsEventsMixin, IdsKeyboardMixin) {
+   constructor() {
+     super();
+   }
+
+    ratingsConfig = {
```
tmcconechy, line 30: For consistency we have been using `this.state.blah` so instead of config use this.state

src/ids-rating/ids-rating.js
```diff
@@ -0,0 +1,209 @@
...
+        color: this.getAttribute('color'),
+        size: this.getAttribute('size'),
+        value: this.getAttribute('value'),
+        readonly: this.getAttribute('readonly'),
+        clickable: this.getAttribute('clickable'),
+        compact: this.getAttribute('compact')
+      }
+    }
+
+    connectedCallback() {
```
tmcconechy, line 43: All methods should have jsdoc and be either private or public. See examples .. (some might not have done that for this method but its a general rule)

src/ids-rating/ids-rating.js
```diff
@@ -0,0 +1,209 @@
...
+      this.buildDOM();
+      console.log(this.ratingsConfig)
```
tmcconechy, line 45: Some lint errors to fix

src/ids-rating/ids-rating.js
```diff
@@ -0,0 +1,209 @@
...
+    template() {
+      return `<div id="rating"></div>`;
+    }
+
+    /**
+    * @returns {Array<string>} this component's observable properties
+    */
+    static get properties() {
+      return [
+        'value',
```
tmcconechy, line 61: Please use props - see https://github.com/infor-design/enterprise-wc/blob/main/src/ids-alert/ids-alert.js#L41 for an example

src/ids-rating/ids-rating.js
```diff
@@ -0,0 +1,209 @@
...
+        'stars',
+        'readonly'
+      ];
+    }
+
+    /**
+    * @param {any} v
+    */
+    set value(v) {
```
tmcconechy, line 70: Please correct jsdoc and maybe use same names as others for consistency https://github.com/infor-design/enterprise-wc/blob/main/src/ids-input/ids-input.js#L569

src/ids-rating/ids-rating.js
```diff
@@ -0,0 +1,209 @@
...
+
+    get value() {
+      return this.getAttribute('value')
+    }
+
+    /**
+    * @param {any} v
+    */
+    set stars(v) {
+      if(v) {
```
tmcconechy, line 84: For these props add the case where the attribute is empty and it clears it see https://github.com/infor-design/enterprise-wc/blob/main/src/ids-hyperlink/ids-hyperlink.js#L77

src/ids-rating/ids-rating.js
```diff
@@ -0,0 +1,209 @@
...
+        this.setAttribute('stars', v.toString());
+      }
+    }
+
+    get stars() {
+      return this.getAttribute('stars')
```
tmcconechy, line 90:
```suggestion
      return this.getAttribute('stars') || 5;
```

For the defaults you did above you can just do this....

src/ids-rating/ids-rating.js
```diff
@@ -0,0 +1,209 @@
...
+        this.addRemoveClass();
+      }
+      this.udpateValue()
+    }
+
+    /**
+    * @param {HTMLDivElement} element
+    * @param {number} index
+    */
+    buildRatingStar(element, index) {
```
tmcconechy, line 122: For these build** methods i think its simpler if you just put it in the template string instead of creating objects here.. That would remove maybe need for 2/3 methods... Many are like that https://github.com/infor-design/enterprise-wc/blob/main/src/ids-textarea/ids-textarea.js#L144

We dont usually use createElement just put all the logic in the template function

src/ids-rating/ids-rating.js
```diff
@@ -0,0 +1,209 @@
...
+      div.innerHTML = starFilled;
+      div.innerHTML += starHalf;
+      ratingItem.appendChild(div);
+      element.appendChild(ratingItem);
+    }
+
+    /**
+    * @param {HTMLDivElement} el
+    */
+    ratingBuilder(el, stars) {
```
tmcconechy, line 144: Are all these methods needed? they seem to not follow the patterns. I think that the template () comment might clean it up

src/ids-rating/ids-rating.js
```diff
@@ -0,0 +1,209 @@
...
+        ratingChildren[i].classList.add('active');
+      }
+      if (value < roundValue) {
+        const activeArr = ratingChildren.filter((act) => act.classList.contains('active'));
+        const lastItem = activeArr[activeArr.length - 1];
+        lastItem.classList.add('is-half');
+      }
+    }
+
+    udpateValue() {
```
tmcconechy, line 199: Spelled wrong but maybe i tihnk for all these updates we do it view the setter...

src/ids-rating/ids-rating.scss
```diff
@@ -0,0 +1,49 @@
...
+  cursor: pointer;
+  display: block;
+}
+
+.rating-outer:hover {
+  color: orange;
+}
+
+.rating-inner {
+  color: orange;
```
tmcconechy, line 24: This isnt a valid color.. Please use the mixins...

test/ids-rating/ids-rating-func-test.js
```diff
@@ -0,0 +1,41 @@
...
+        window.document.body.appendChild(elem)
+        elem.remove();
+        expect(document.querySelectorAll('ids-rating').length).toEqual(1);
+        expect(errors).not.toHaveBeenCalled();
+    });
+
+    it('renders correctly', () => {
+        const elem = new IdsRating();
+        window.document.body.appendChild(elem);
+        elem.template();
```
tmcconechy, line 29: Do we need to call template here? Also i've changed these to use elem.shadowDOM.innerHTML as of now you can see the snap shot is empty. its more interesting to use the innerHTML of the shadow DOM

tmcconechy (changes requested):
Note some of these comments may reply to the other PR

1. dont include an empty example.js ...
2. on the example page it looks like its readonly but i can still hover it and get a hover state and it looks like its clickable
3. complete types see example https://github.com/infor-design/enterprise-wc/blob/main/src/ids-hyperlink/ids-hyperlink.d.ts
4. new since you started this but please add themes as well see https://github.com/infor-design/enterprise-wc/tree/main/src/ids-base#ids-theme-mixin for notes
5. use the template and getters / setters more for consistency... 
6. can you use "value/stars" and update?
7. do we need a clickable example? i.e. in the PR try to add the examples and not make the reviewer add code to see its working, that should be in the examples already (or tests)

src/ids-rating/ids-rating.js
danielortiz1982: Hi Tim,
Can you please point me to a re-render example. I put this into it's own fn to create stars based on the stars attribute value being passed in. Thnx!

src/ids-rating/ids-rating.js
danielortiz1982: Thanks! Didn't catch this.

src/ids-rating/ids-rating.js
danielortiz1982: Thanks, I will keep this in mind as I work through it. I was trying to write some defensive logic in case the attribute was not added to the wc. The setter and getter do not fire when the attribute is not present. That was my thinking behind it.

src/ids-rating/ids-rating.js
tmcconechy: There is one example in list view https://github.com/infor-design/enterprise-wc/blob/main/src/ids-list-view/ids-list-view.js#L97-L102 you can make a method like that...

I also am considering a quick PR with a rerender in the pass that just reapplies the template. Would that be usable?

Also think you could fix this by just replacing the star count directly in the setter code.

danielortiz1982: Thanks! Will do, I will try this out.

tmcconechy:
Closing this as agreed until its ready. Comments will still be here for reference

PR #109, opened 2021-05-05
Run GH Actions build step only on PRs without "skip ci tests" label

rob2d:

This patch is for considerations with resources on CI infra; we are currently running CI on all PR changes. Gives devs the ability to turn off auto-builds while they are addressing feedback without needing to obfuscate how they want to commit or track history in their work + save on resources as we tend to get close to our build limits.

N/A

Once pushed/existing here on GH, this pull request should: 
- not run a `build` step on Github Actions with a `skip ci tests` label once it is submitted.
- auto run when `skip ci tests` label is removed.
- not fail because of `ci.yml` syntax errors.

To test:
- add a label to this PR of "skip ci tests"
- see that the new job triggered in Actions tab on this repo doesn't build or pass a new job 
(this step *might* not be testable without adding test commits, but see history for ref otherwise)
- remove label of "skip ci tests"
- see that a job got triggered
- cancel running workflow since theres' no `ci.yml` `build` code changes here besides minor CI config checks.
image

See screens which show example history @ https://github.com/infor-design/enterprise-wc/actions tab 
(note: label checked was changed from `WIP` to `skip ci tests` promptly after these were taken)

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ on:
 jobs:
   build:
 
+    if: "!contains(github.event.pull_request.labels.*.name, 'WIP')"
     runs-on: ubuntu-latest
 
     strategy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,7 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+    types: [ opened, reopened, synchronize, unlabeled ]
 
 jobs:
   build:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ on:
 jobs:
   build:
 
-    if: "!contains(github.event.pull_request.labels.*.name, 'WIP')"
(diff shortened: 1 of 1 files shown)


EdwardCoyle:
@rob2d @tmcconechy any idea why things hung here?

tmcconechy:
Yeah i guess since the build is cancelled the other checks are just waiting for status. Maybe we need to conditionally disable the checks? Not sure if thats possible as its a github setting

tmcconechy:
I think i can approve this but wondering if you do want to look at the state where its appearing hung?

rob2d:
It seems like GH is not re-validating the job on the conditional once it got skipped when the label was removed the first time. I am not sure that it's something we can control and PR was just to see if there was a way in the ci.yml can skip based on labels but will try to make time for digging into it.

rob2d:
or well, actually -- sorry, looking over again today it looks fine and there's nothing hanging 😅 I cancelled the last run intentionally. Listed it in the instructions for testing (since if you can see it running jobs, there was no code changes outside of simple ci.yml patches to drain resources with).

tmcconechy (approved):

EdwardCoyle (approved):

tmcconechy (approved):

PR #112, opened 2021-05-07
Misc - Render / Rendered Pattern

tmcconechy:

Based on feedback from @danielortiz1982 this pr

- fixes the render method so it can be recaled to re-render a component it can be used like this..

```
render() {
  super.render();
  // Do other stuff you need to like re-add event handlers or anything specific
}

rendered() {
 // After Painting
}
```

- also fixed a console warning about requestAnimationframe taking too long
- Also it was noted that setting item-count in the DOM did not work. This is because you can't really do that as you need to reapply the matching data set so i removed the visibility of this property

- test http://localhost:4300/ids-virtual-scroll
- test http://localhost:4300/ids-list-view 
- when scrolling check the console for errors
- check the DOM and notice there is no item-count inline property
- test http://localhost:4300/ids-data-grid/virtual-scroll


EdwardCoyle:
@tmcconechy @rob2d I'm still trying to figure this out after our talk yesterday.  My understanding up to now was that visual jank can get pretty bad if we have multiple `requestAnimationFrame` calls resolve their callbacks on the same frame.  Maybe I'm misunderstanding this, but if we're just calling it on every component's render step, coupled with potentially having thousands of components on the page, wouldn't that set up the scenario for jank?

The render loop uses a [`requestAnimationFrame` loop internally](https://github.com/infor-design/enterprise-wc/blob/main/src/ids-render-loop/ids-render-loop.js#L127).  Its main intention is to replace our use of `setTimeout` for operations that need staggering over a fixed number of frames, or for handling animation timelines.  It could be used to handle this problem, but since we don't really need the "loop" aspect here, it might just be "extra"... however if there are thousands of them potentially going off simultaneously... it may help prevent that?

The render loop queue executes everything attached to it on each frame, so it solves the multiple callback problem, but you're still executing that many "renders" per frame if everything ties into that one run.  I'm not sure how that will affect performance overall... seems like it could be nominal but 🤷‍♂️ -- never tried it with this many things at once.

I've read conflicting ideas about the performance, and it seems the one StackOverflow thread that had a perf test attached isn't currently available: https://stackoverflow.com/a/55326296/4024149.  It seems to suggest performance is negligible if you try and limit the number of calls, so it's got me wondering if `requestAnimationFrame` has gotten better over time and maybe doesn't require the queuing.

TL;DR - not sure, needs testing, preferably with tons of components needing rendering in tandem.

If you want to use render loop for this, we can bypass the mixin and just attach directly to the global instance:

```js
import { renderLoop } from '../ids-render-loop/ids-render-loop-global';
import { IdsRenderLoopItem } from '../ids-render-loop/ids-render-loop-item';

// ...

if (this.rendered) {
  renderLoop.register(new IdsRenderLoopItem({
    duration: 1,
    timeoutCallback: () => {
      this.rendered(); 
    }
  }));
}
```

EdwardCoyle:
Some additional reading on the subject (some of it explains needing to "manage" the queuing in some cases):

- https://medium.com/@paul_irish/requestanimationframe-scheduling-for-nerds-9c57f7438ef4
- https://maqentaer.com/devopera-static-backup/http/dev.opera.com/articles/view/better-performance-with-requestanimationframe/index.html
- https://gamedev.stackexchange.com/questions/125977/dealing-with-multiple-requestanimationframe-calls

tmcconechy:
Yeah so maybe we just make it use the mixin. Can i use the loop not as a mixin (as i'll need to do it in this scenario) or perhaps you could add to the PR?

EdwardCoyle:
@tmcconechy Added to the PR directly.  I guess let's just test it and see how it performs.  We can always pull it out relatively quick if it doesn't work.

EdwardCoyle (dismissed):

deep7102 (approved):
🍌

Issue #120, opened 2021-04-01
IdsTabs: Add new web component


katieinfor:
**Describe the bug**

Bring over the current tabs component.  See if we can simplify and improve the code at all while improving as we go. Current tab component https://master-enterprise.demo.design.infor.com/components/tabs

Markup (not fixed in stone at all) - but taking a lot of ideas from https://vaadin.com/components/vaadin-tabs/html-examples/tabs-basic-demos#vertical-tabs

We'll start with just the tab and list, and leave the content containers to the developer. Possibly adding a "scrollview" in the future with swipe. We'll just fire a selected event and they can do what they like

```
<ids-tabs orientation="vertical">
  <ids-tab selected="true"><ids-text>Tab One (not linkable)</ids-text></ids-tab>
  <ids-tab><ids-hyperlink>Tab Two (linkable)</ids-hyperlink></ids-tab>
</ids-tab>
```

**Functionality to bring over**
Phase 1
- Supports Standard Tabs
- Supports Vertical Tabs
- Supports Counts
- Tabs can be linkable (can be accessed by hash url) or not
- Accessible markup (see https://googlechromelabs.github.io/howto-components/howto-tabs/#demo) 
- show events so the user can use them to set tab contents - should be beforeselected, selected, afterselected and beforeSelected (async) see https://github.com/infor-design/enterprise-wc/blob/main/src/ids-tooltip/ids-tooltip.js#L223-L254

Phase 2
- Don't bring over animation (design in progress)
- Supports Overflow (we'll try and change this to scrollable with left and right arrows)  (design in progress) - will need to think about dismissible and add/remove buttons when we do this (Phase 2)

**Additional context**
- 100% test coverage
- DOCS, CHANGE log ect. 
- https://github.com/infor-design/enterprise-wc/blob/master/doc/CHECKLIST.MD
- themes 

See completed components like card, tag, badge, hyperlink as reference

PR #188, opened 2021-05-15
120 (5033) - ids-tabs wc Phase 1

rob2d:

Adds `IdsTabs`/`IdsTab` components in addition to supporting and iterating to modular import/export support where this branch has had to pull from.

Addresses issue #120  (phase  1). 

- pull this repo
- visit http://localhost:4300/ids-tabs and the example page
- verify that functionality outside of that listed in `src/ids-tabs/TODO.md` is covered, and accessible.


app/ids-base/index.js
```diff
@@ -0,0 +1,2 @@
+import IdsEventsMixin from './ids-events-mixin';
```
tmcconechy, line 1: Im wondering why this is needed as these imports should be in any of the individual components that need them?

app/ids-tabs/index.js
```diff
@@ -0,0 +1,14 @@
+import IdsTabs, { IdsTab, IdsTabsDivider } from '../../src/ids-tabs';
+
+const booth = document.querySelector('#test-ids-tabs');
+const tabContainers = [...booth.children];
+
+for (const c of tabContainers) {
+  if (!c.matches(':first-child')) {
+    c.style.marginTop = '8px';
```
tmcconechy, line 8: Is this some sort of hack that should be done in the component itself?

app/ids-tabs/index.js
```diff
@@ -0,0 +1,14 @@
+import IdsTabs, { IdsTab, IdsTabsDivider } from '../../src/ids-tabs';
+
+const booth = document.querySelector('#test-ids-tabs');
```
tmcconechy, line 3: Both? Booth?

src/ids-base/ids-element.js
```diff
@@ -105,7 +105,13 @@ class IdsElement extends HTMLElement {
     // Give implementing-component an opportunity to react to
     // new render and replacement of container node
 
-    this.rendered?.();
+    // note that it must run on the next paint to be sure rendered()
+    // was binded to the component reliably; otherwise it doesn't
+    // run every time on the first paint due to inheritance
+
+    if (this.rendered) {
```
tmcconechy, line 112: See https://github.com/infor-design/enterprise-wc/pull/112 -> added it with the mixin already so this can go once thats merged

app/ids-tabs/example.html
```diff
@@ -0,0 +1,33 @@
+<ids-layout-grid>
+  <ids-text font-size="12" type="h1">Tabs</ids-text>
+</ids-layout-grid>
+<div id="test-ids-tabs">
```
tmcconechy, line 4: Can we put each example on a different page and just show the "default" one on an initial page. Suggest page names vertical.html and counts.html

src/ids-base/ids-theme-mixin.js
```diff
@@ -68,3 +68,4 @@ const IdsThemeMixin = (superclass) => class extends superclass {
 };
 
 export { IdsThemeMixin };
+export default IdsThemeMixin;
```
tmcconechy, line 71: Why do we change all this stuff exactly? Seems like it was fine... Also this appears like now it has two exports?

src/ids-base/index.js
```diff
@@ -0,0 +1,4 @@
+export { default as IdsThemeMixin } from './ids-theme-mixin';
```
tmcconechy, line 1: Why do we need this in index.js?

src/ids-tabs/README.md
```diff
@@ -0,0 +1,74 @@
+# Ids Tabs Component
+
+## Description
+
+This component styles tabs which serve as navigation points and
```
tmcconechy, line 5:
```suggestion
Tabs are a navigation element used to allow users to easily access different areas of a site or different parts of an individual page. A user can navigate between page sections with a tab. Best for communicating different views of the data, or offering navigation between related data.
```

src/ids-tabs/README.md
```diff
@@ -0,0 +1,74 @@
...
+way to highlight which section/content in a page you are currently
+focused on.
+
+## Use Cases
+
+- When you need to add a series of navigable headers and vertical space is a priority.
```
tmcconechy, line 11:
```suggestion
- Used when need information to be highly scannable and to simplify navigation.
- Used when a page needs a more organized structure.
```

src/ids-tabs/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Features (With Code Examples)
+
+A normal text element used as a web component in size 16px.
```
tmcconechy, line 15:
```suggestion
A normal default horizontal tab component
```

src/ids-tabs/README.md
```diff
@@ -0,0 +1,74 @@
...
+    <ids-tabs>
+        <ids-tab>Example One</ids-tab>
+    </ids-tab>
+</ids-tabs>
+```
+
+URI-Hashing-Enabled tab
+
+```html
+<ids-tabs uri-hashing="true">Disabled Text</ids-text>
```
tmcconechy, line 28: Is Uri Hashing really the right name? Maybe its me but i havent heard this term

src/ids-tabs/README.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+TODO: phase i examples
+
+## Settings and Attributes
+
+`IdsTabs`
```
tmcconechy, line 35:
```suggestion
```

src/ids-tabs/README.md
```diff
@@ -0,0 +1,74 @@
...
+- `disabled` {boolean} disables all tabs.
+- `value` {string} set which tab is currently selected. If tab children
+do not have a value, will fall back to being a 0-based index. Otherwise, it can
+also be any string as long as there are relevant matches for the values.
+- `orientation` {'horizontal' | 'vertical'} defaults to horizontal; controls
+the direction/axis tabs are flowed on.
+
+`IdsTab`
```
tmcconechy, line 43:
```suggestion
```

src/ids-tabs/ids-tab.js
```diff
@@ -0,0 +1,225 @@
...
+  mix
+} from '../ids-base/ids-element';
+import { IdsEventsMixin } from '../ids-base/ids-events-mixin';
+import { stringToBool, buildClassAttrib } from '../ids-base/ids-string-utils';
+import IdsText from '../ids-text/ids-text';
+import styles from './ids-tab.scss';
+
+/**
+ * IDS Tab Component
+ *
```
tmcconechy, line 15:
```suggestion
```

src/ids-tabs/ids-tab.js
```diff
@@ -0,0 +1,225 @@
+import {
+  IdsElement,
+  customElement,
+  props,
+  scss,
+  mix
+} from '../ids-base/ids-element';
+import { IdsEventsMixin } from '../ids-base/ids-events-mixin';
+import { stringToBool, buildClassAttrib } from '../ids-base/ids-string-utils';
```
tmcconechy, line 9: This can now by imported from 

```
import {  IdsElement,  customElement,  props,  scss,  mix, stringUtils} from '../ids-base/ids-element';
```

src/ids-tabs/ids-tab.js
```diff
@@ -0,0 +1,225 @@
...
+import IdsText from '../ids-text/ids-text';
+import styles from './ids-tab.scss';
+
+/**
+ * IDS Tab Component
+ *
+ * @type {IdsTab}
+ * @inherits IdsElement
+ * @part container - the tab container itself
```
tmcconechy, line 18: Can add @mixes for the events mixin

src/ids-tabs/ids-tab.js
```diff
@@ -0,0 +1,225 @@
...
+  constructor() {
+    super();
+  }
+
+  /**
+   * Return the properties we handle as getters/setters
+   * @returns {Array} The properties in an array
+   */
+  static get properties() {
+    return [props.VALUE, props.SELECTED, props.ORIENTATION, props.COUNT];
```
tmcconechy, line 32: Have been alphabetizing but not sure that's an official thing we all agreed on

src/ids-tabs/ids-tab.js
```diff
@@ -0,0 +1,225 @@
...
+    return idsTextEls.map((textEl) => {
+      const slotNode = textEl.querySelector('slot')?.assignedNodes?.()?.[0];
+      return slotNode?.textContent || textEl.textContent;
+    }).join(' ');
+  }
+
+  /**
+   * Sets the data-text of nested ids-text to it's slot
+   * text content. Fixes issue with bold moving around
+   * when we edit content.
```
tmcconechy, line 208: @private

src/ids-tabs/ids-tabs-divider.js
```diff
@@ -0,0 +1,37 @@
...
+  constructor() {
+    super();
+  }
+
+  /**
+   * Create the Template for the contents
+   * @returns {string} the template to render
+   */
+  template() {
+    return (`<div class="ids-tabs-divider"></div>`);
```
tmcconechy, line 29: Shouldn't this be a string? Or does this work too?

src/ids-tabs/ids-tabs.js
```diff
@@ -0,0 +1,397 @@
...
+  props,
+  scss,
+  mix
+} from '../ids-base/ids-element';
+import {
+  IdsKeyboardMixin,
+  IdsEventsMixin,
+  IdsThemeMixin,
+  IdsStringUtils
+} from '../ids-base';
```
tmcconechy, line 13: Maybe i see what your doing now... I guess this looks ok but the old way i dont think was a problem was it?

tmcconechy (commented):
This looks pretty good. Nothing major some minor comments.
But wondering about the export / mixin changes. Its probably fine but wasnt expecting it.

tmcconechy:
Another comment i was thinking is should we show "how" to change content? Im not sure about that since we did want it broken out but thats also the point so maybe we need to show some contents changing...

src/ids-tabs/ids-tabs-divider.js
rob2d: good point it can be just a string. It works/compiles fine, evals to just an str in an expression. Just sort of a habit I have from React to do that in any templating -- since sometimes they span multiple lines and most of the time they are multi-line.

src/ids-tabs/ids-tabs.js
rob2d: it was a little bit non idiomatic, since we had mentioned we were going to index.js pattern and also defaults, and it makes it easier to import -- so I was only doing it with things my branches touch as not to create a super huge branch that might interfere with other people's work more than they would like. Just figured it we started doing it incrementally it would be less likely to never get updated, but if we aren't doing that I can revert. It could have been total miscommunication/understanding on this also.

tmcconechy: I think this makes sense. I was thinking maybe to export them out of IdsElement as well like the others but not sure if that means that just imports them into EVERYTHING.

The other side of it is we just do it in one place it sort of just makes it immediately inconsistent.

In the future this sort of thing probably should be a separate PR and discussion as there are a few ways we can do it but it's there now, lets see what @EdwardCoyle thinks... I'm ok with it.

src/ids-tabs/ids-tabs-divider.js
tmcconechy: cool yeah lets just a use a string.. didnt know this worked so it looked wierd but that will keep us more consistent

app/ids-tabs/example.html
rob2d: ok sounds good -- just wasn't aware we had the option on it

app/ids-base/index.js
rob2d: I think I added this while troubleshooting why certain things were not working in tests (there is a weird quirk where for at least web components, we need to import them all explicitly from at least one import -- figured could be related). Will try to remove now since it shouldn't be needed).

rob2d: actually looking and the folder has all of these; so was just making a default import e.g. adding index.js.

(So I'm not quite sure why they were needed in the modules... just added to centralize those imports when they were)

app/ids-tabs/index.js
rob2d: this was not related to the component code; it's the containers around the tabs (the tabs are sizes to fit the containers -- so I added code to style the containers since we don't have CSS here). Saw some similar things in example setup so that's why I added here.

src/ids-base/index.js
rob2d: if index.js exists, then pulling from /ids-theme-mixin will look here first. It is the place to centralize imports.

(this may answer some other similar questions on it in feedback here)

src/ids-tabs/README.md
rob2d: I was using the idea of "URI" and the fact that the "#" is hash tag. Hashing is just a general term for making something accessible in a HashMap/dict which is how we're doing this internally. 

The term for the "#" is just "URI Fragment" ( https://en.wikipedia.org/wiki/URI_fragment ) but it would seem weird to say "uri-fragmenting".

So I'm not sure what to call it if that's not good. Maybe auto-anchor-tags?

tmcconechy: Before we called it url hash? So maybe im wrong here https://main-enterprise.demo.design.infor.com/components/tabs/example-url-hash-change.html#tab-number-two see what @EdwardCoyle thinks

I guess url-hashing ? https://danielmiessler.com/study/difference-between-uri-url/#:~:text=The%20terms%20%E2%80%9CURI%E2%80%9D%20and%20%E2%80%9C,as%20HTTPs%20%2C%20FTP%20%2C%20etc. is it a url or a uri? Seems like a url

app/ids-tabs/index.js
tmcconechy: Hmm still not following? What is it for? Do the developers need to add it? Do we even need it at all?

src/ids-tabs/README.md
rob2d: I think this suggestion might have been a mistake? just had 2 lists to differ "IdsTabs" from "IdsTab" since they both are different components; was not sure that pattern was alright or not though.

src/ids-tabs/README.md
rob2d: (same as above)

src/ids-tabs/README.md
tmcconechy: Oh i see. It seemed like a typo. Then maybe can be a bit more explicit then like `Individual Tabs Settings (ids-tab)` and `Tab Container Settings (ids-tabs)`? Something along those lines

I did something similar here with normal (grid) settings vs column settings https://github.com/infor-design/enterprise-wc/blob/main/src/ids-data-grid/README.md#settings-and-attributes

src/ids-tabs/README.md
rob2d: I think it's a URL hash. URI is something that points to a specific resource -- in this case the tab. I have seen both interpretation but just from the definition online. The link says that but is a little vague.

The reason it's confusing is because: a URL is something that includes http://xyz.yzx

But in this case we only point to a specific resource in the URL which is a URI e.g. #. A URI can be kind of both a superset (uses any protocol besides web) and subset (uses specific resource). Looked into it a bit for Wizard.

In our case URL would be:

http://wherever-user-is-visiting-on-app

URI is

#tab-converted-to-identifier

image

The article only uses some examples, which make the whole point. I can see his point, but some other things show it's not correct just from Base API if we think about it. For example `encodeURI` is named `encodeURI` vs `URL` because we don't have to convert the URL: it just takes any subset of it e.g. `encodeURI(theHashItself)` and not `encodeURL(theHashItself)` (since URL would imply there is some guaranteed ftp:// or http://

rob2d: https://stackoverflow.com/questions/176264/what-is-the-difference-between-a-uri-a-url-and-a-urn

So there's another convo here on it. The thing I think might sum it up and doesn't contradict the article is:
> So all URLs are URIs, and all URNs are URIs - but URNs and URLs are different, so you can't say that all URIs are URLs.

And in the case of what the component is doing, it's not really messing with the URL of the website; only the identifier on the page/document itself -- which is the URI fragment: 

https://en.wikipedia.org/wiki/URI_fragment

If we thought to let a user use the URL itself, that would actually break the tabs functionality since a URL is a full address that a browser can visit vs just the part itself.

So another example/idea on it: tabs component can actually be used on an SPA (say for React or Angular), and the URL/routes can transform. But in the meantime the only part we really care about is the URI fragment (#some-tab-auto-addr) that will keep the user on same page and just change the tab focused.

Also taken from that on the discussion @ SO link:

> A Uniform Resource Locator is a URI that, in addition to identifying a resource, provides a means of locating the resource by describing its primary access mechanism [RFC3986]. As there is no exact definition of URL by means of a set of Schemes, "URL is a useful but informal concept", usually referring to a subset of URIs that do not contain URNs [RFC3305].

The key I think is the **in addition to identifying a resource...provides the means of locating** there.

tmcconechy: ok lets go with your uri-hashing as is then https://www.w3.org/wiki/HashURI seemed wierd to me at first but seems like its the right term

src/ids-tabs/README.md
rob2d: ok that's a good idea (thanks!)

src/ids-base/ids-theme-mixin.js
rob2d: (just waiting on the meeting today to resolve this particular one like you mentioned earlier at this point; putting a note on it)

src/ids-tabs/ids-tab.js
```diff
@@ -0,0 +1,238 @@
...
+
+  get selected() {
+    return this.getAttribute(props.SELECTED);
+  }
+
+  /**
+   * @param {string} value value which becomes selected by tabs component
+   */
+  set value(value) {
+    /* istanbul ignore else */
```
tmcconechy, line 149:
```suggestion
    /* istanbul ignore next */
```

src/ids-base/index.js
tmcconechy: I have a PR coming with all these changes so you could pull this out for now but if you dont it should merge ok as its close to this.

.eslintrc.js
```diff
@@ -53,6 +53,16 @@ module.exports = {
     // we aren't doing special math with binary/hex radix numbers often,
     // so this removes need for parseInt(number, 10);
     radix: 0,
+    // allow ability to move forward to index.js
```
tmcconechy, line 56: This will also be in my PR and may merge conflict

rob2d: ok thanks for the heads up. Will just keep the PR open until then in that case as it works and then sort the conflicts when it's ready so that I can chip away on Spinbox as well.

app/ids-tabs/index.js
rob2d: it's nothing fancy it was just styling the components since we can't do that manually on our test examples do to the security limitations/etc. This was just sandbox code with simple query selector; unless I'm mistaken I was told that the `index.js` was for setting up sandboxes, but also tabs were to fill their parents.

```
<div:first-child>
  <ids-tabs>...</ids-tabs>
</div>
<div>
  <ids-tabs>...</ids-tabs>
</div>
```

(not there anymore though/outdated)

tmcconechy: Oh i see so its "demo css". Actually you can use css you just have to put it in a file so it this is an example:

1) Put it in a file (i would use example.scss not index.scss) https://github.com/infor-design/enterprise-wc/blob/main/app/ids-popup-menu/index.scss
2) Ref it in example.js for the example.html which will also run https://github.com/infor-design/enterprise-wc/blob/main/app/ids-popup-menu/example.js#L1 (can also go in several examples if need be)

rob2d: oh wow... did not know. Thanks! Shall do this next time.

tmcconechy (approved):
OK good for me for now then 👍🏻  Nice work!

deep7102:
Looks very nice! just saw few
- In count tabs `separator` did not see in example page, are we going to add separator in next phase
- First time focus/selected state different click vs keyboard (with regular, count and vertical tabs)

image
image

PR #189, opened 2021-05-18
118 - Implement IDS Modal

### LINKED ISSUE #118: IdsModal: Add new web component
**Describe the bug**

Bring over the current modal component.  See if we can simplify and improve the code at all while improving as we go. Current modal component https://master-enterprise.demo.design.infor.com/components/modal

```
Markup TBD needs further consideration and work
```

- Supports an overlay
- Supports nested modals
- Supports keyboard 
- Use ids popup for modal 

**Additional context**
- 100% test coverage
- DOCS, CHANGE log ect. 
- themes 

EdwardCoyle:
This PR implements the Ids Modal component with basic functionality:
- show/hide
- overlay with opacity
- nesting capability
- keyboard handling

Closes #118

- Pull, Build, Run!
- Open http://localhost:4300/ids-modal
- Use the buttons to show/hide the modal.  They should work by click and by keyboard (focus of the correct elements should occur)
- Check that visibility can be set and accounted for on page load - http://localhost:4300/ids-modal/visible
- Check that nesting works: http://localhost:4300/ids-modal/nested


src/ids-modal/ids-modal-z-counter.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Provides a global way to display
+ */
+class IdsModalZCounter {
+  constructor() {
+    this.zCounter = 1000;
```
tmcconechy, line 6:
```suggestion
    this.zCounter = 1020;
```

Minor but the z-indexes i worked out start modals at 1020

tmcconechy (changes requested):
Not sure if this is "make changes" worthy but some small things i spotted. But the one thing i wanted to ask is that the style looks a bit different from https://main-enterprise.demo.design.infor.com/components/modal/example-index.html

- border radius
- overlay is a bit lighter (but if this is on purpose i think it might be better this way and solve some issues with "seeing behind")
- not sure if the box shadow on the outside is right?

Also not sure if its a TODO but if i tab around the modal it "jumps out" to the parent page instead of staying in the modal

EdwardCoyle:
@tmcconechy I addressed some of your comments, with some notes:
- I added a "Modal" style to the popup but I was finding the only differences between what was there and what the old modal had were the padding and the border color.  Everything else should be ok (border radius and box shadow included)
- I did lessen the opacity on purpose, but it's easily adjusted if we want to put it at 0.7
- I have the keyboard nav wrapping on the [TODO list for a phase 2](https://github.com/infor-design/enterprise-wc/pull/189/files#diff-ebd832bb38d33dfe08609c2ead5665cee6cce355e09755278d82ab1963567281R4)

tmcconechy (dismissed):

deep7102 (approved):
Looks nice 🍌
do not have solution but just wonder about nested stuff as deep we going overlay will be looking more darker

PR #195, opened 2021-05-22
139 ids-spinbox wc

### LINKED ISSUE #139: IdsSpinBox: Add new web component

**Phase 1 Features (MVP)**
- Settings min/mask and step
- keyboard/ accesibility

**Future Features (Not Now)**
- Autocorrect on blur
- Attributes/Internal IDs

**Suggested Internal Markup**
```js
```

**Suggested Dependencies**
- ids-trigger-field
- ids-input
- ids-mask-mixin

**Settings**
- step (string / live)
- max (string / live)
- min (string / live)

**Events**
- change

**Other**

rob2d:

Misc: also fixes:
- an issue with `ids-button` disabled prop persisting (`ids-buton` => disabled setter) 
- a small potential issue with `ids-keyboard-mixin` where we sub on an array as the key vs the events (since we do not actually group key combos or use it in this manner in other components -- the fix was 2 lines so figured small enough to include here)

NOTE: there is some things that are still a WIP. Pushing in case anyone would like to review it since the code is now all functionally there. 

There is still a need for documentation in the README/CHANGELOG, and should add simple tests for dirty tracker and disabled states; outside of that all is ready for feedback.

Closes #139 

- pull this PR branch locally and run environment (`139-ids-spinbox`)
- visit http://localhost:4300/ids-spinbox/
- ensure functionality listed at #139 for Phase 1 is implemented.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -52,7 +52,6 @@ module.exports = {
       exports: 'never',
       functions: 'never'
     }],
-    'no-await-in-loop': ['off', { }],
     // we aren't doing special math with binary/hex radix numbers often,
     // so this removes need for parseInt(number, 10);
     radix: 0,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -52,6 +52,7 @@ module.exports = {
       exports: 'never',
       functions: 'never'
     }],
+    'no-await-in-loop': ['off', { }],
     // we aren't doing special math with binary/hex radix numbers often,
     // so this removes need for parseInt(number, 10);
     radix: 0,
--- a/src/ids-spinbox/ids-spinbox.js
+++ b/src/ids-spinbox/ids-spinbox.js
@@ -57,6 +57,13 @@ class IdsSpinbox extends mix(IdsElement).with(IdsEventsMixin, IdsKeyboardMixin) 
     };
 
     this.setAttribute('tabindex', 0);
+    this.onEvent('click.decrement', this.container.children[0], () => {
+      this.#onDecrement();
+    });
+
+    this.onEvent('click.increment', this.container.children[2], () => {
+      this.#onIncrement();
+    });
 
     this.listen(['ArrowUp', 'ArrowDown'], this.input, (e) => {
       const key = e.key;
--- a/src/ids-spinbox/ids-spinbox.js
+++ b/src/ids-spinbox/ids-spinbox.js
@@ -3,13 +3,17 @@ import {
   customElement,
   scss,
   props,
+  stringUtils,
   mix
 } from '../ids-base/ids-element';
 import IdsButton from '../ids-button/ids-button';
 import IdsInput from '../ids-input/ids-input';
 import { IdsEventsMixin } from '../ids-base/ids-events-mixin';
 import { IdsKeyboardMixin } from '../ids-base/ids-keyboard-mixin';
+import { IdsDirtyTrackerMixin } from '../ids-base/ids-dirty-tracker-mixin';
 import styles from './ids-spinbox.scss';
+
+const { stringToBool } = stringUtils;
 
 /**
  * IDS Spinbox Component
@@ -18,7 +22,11 @@ import styles from './ids-spinbox.scss';
  */
 @customElement('ids-spinbox')
 @scss(styles)
-class IdsSpinbox extends mix(IdsElement).with(IdsEventsMixin, IdsKeyboardMixin) {
+class IdsSpinbox extends mix(IdsElement).with(
+    IdsEventsMixin,
+    IdsKeyboardMixin,
+    IdsDirtyTrackerMixin
+  ) {
   constructor() {
     super();
   }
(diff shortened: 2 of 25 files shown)


.eslintrc.js
```diff
@@ -52,6 +52,7 @@ module.exports = {
       exports: 'never',
       functions: 'never'
     }],
+    'no-await-in-loop': ['off', { }],
```
tmcconechy, line 55: Super brief comment as to why we need this off?

EdwardCoyle (commented):
@rob2d everything on this looks great so far!  I had two small comments, but other than these I'd be ready to approve:

- When you first open http://localhost:4300/ids-spinbox/, the input portion of the disabled spinbox is focused somehow (shouldn't be possible)
- When you change the dirty example, the dirty indicator doesn't appear to be its normal yellow color.  It may just be a matter of importing the [Dirty Tracker mixin styles](https://github.com/infor-design/enterprise-wc/blob/main/src/ids-base/ids-dirty-tracker-mixin.scss) into spinbox (not sure)

Screen Shot 2021-05-24 at 10 09 04 AM
Screen Shot 2021-05-24 at 10 12 15 AM

src/ids-base/ids-keyboard-mixin.js
```diff
@@ -42,13 +42,17 @@ const IdsKeyboardMixin = (superclass) => class extends superclass {
...
   /**
-   * Add a listener for a key or key code combination
+   * Add a listener for a key or set of keys
    * @param {Array|string} keycode An array of all matchinng keycodes
    * @param {HTMLElement} elem The object with the listener attached
    * @param {Function} callback The call back when this combination is met
    */
   listen(keycode, elem, callback) {
-    this.hotkeys.set(`${keycode}`, callback);
+    const keycodes = Array.isArray(keycode) ? keycode : [keycode];
```
tmcconechy, line 51: This already did work with arrays? https://github.com/infor-design/enterprise-wc/blob/main/src/ids-tag/ids-tag.js#L195

src/ids-spinbox/ids-spinbox.js
```diff
@@ -0,0 +1,307 @@
...
+  set min(value) {
+    if (this.getAttribute(props.MIN) !== value) {
+      this.setAttribute(props.MIN, value);
+
+      this.#updateIncrementDisabled();
+      this.#updateDecrementDisabled();
+    }
+  }
+
+  get min() {
```
tmcconechy, line 173: Can you add more comments to each method?

tmcconechy (commented):
I saw the same things as ED but would call out it also needs
- some docs README.MD i can help with this
- some brief comments in the code methods
- themes (we forgot that on tabs? and wizard?) see https://github.com/infor-design/enterprise-wc/tree/main/src/ids-base#ids-theme-mixin for notes

But overall this is looking really good.

.eslintrc.js
rob2d: I actually had this initially for some tests -- it seemed like there wasn't a logical limitation to have and people just preferred not to who wrote the airbnb, but since we're using `async () =>` in tests it seemed appropriate to be able to `for await` or use `await` in `while` loops.

Can revert for the sake of it just not used yet in this branch though so far.

rob2d: (actually had forgotten to remove also so good point -- the reasoning was more an explanation of why the rule made sense in general)

src/ids-base/ids-keyboard-mixin.js
rob2d: @tmcconechy it worked with arrays but can't be removed predictably so could cause issues in test, since the array always has a new ref. 

e.g. `['ArrowUp', 'ArrowDown']` => trying to remove on `['ArrowUp', 'ArrowDown']` or do any logic with it would not match since array match is shallow. 

If we were doing combo keys (I think the original intention, which it didn't end up being):  it would make sense since its diff listeners but seems we're just doing "both arrow up and arrow down" in a sense.

(I could be missing something here but did try to understand it here since I was finagling with stuff for tests)

src/ids-spinbox/ids-spinbox.js
rob2d: yeah I was sort of considering this with "documentation"; just pushed the branch since the functionality was mostly done in case something was bad. Will do that as soon as I get the other stuff Deepak mentioned and that we were just talking about w `label for` this AM.

tmcconechy: No biggie i figured it was WIP so just a reminder

rob2d:
Thanks for the helpful feedback on this guys. Besides doc, I addressed most of it. I'm going to close this temporarily though just until I ran through a mini circus with accessibility + the labels as I'm noticing what @deep7102 was concerned about this AM. 

Workaround seems really simple (just need to be able to assign `id` on inner input of `ids-input`), but that may also need another smaller PR depending on what the solution here looks like. That could end up being a good idea anyway to add some small tests for the label-as-optional-change since `ids-input` is critical for other components and I def don't want to break things by throwing things out of scope globbed in here.

PR #196, opened 2021-05-25
139 - ids-button: accessibility options

rob2d:

This branch is mainly to add accessibility options so that both Axe tests and Vox tool works properly with WC for some compound components (in this case for the IdsSpinbox). 

(1) Because of some quirks in the shadowDOM, tabIndex has to persist for -1 to the WC as well for disabling things.
(2) There was also some inconsistency with disabled not applying to the button when toggling enabled => disabled that (discovered during dev on the spinbox component for #139).
(3) remove setting role="button" on LightDOM due to axe tools errors (button itself can handle it) [added after PR pushed]

Supports #139 but does not have a direct ticket on JIRA.

- pull this branch (`139-ids-button-accessibility-options`)
- do a quick sanity check that buttons do not behave abnormally/regressively with tabbing on test page.
- see that tests pass on PR

### DIFF
--- a/src/ids-button/ids-button.js
+++ b/src/ids-button/ids-button.js
@@ -12,6 +12,8 @@ import { IdsRenderLoopMixin, IdsRenderLoopItem } from '../ids-mixins/ids-render-
 
 import styles from './ids-button.scss';
 
+const { stringToBool } = stringUtils;
+
 // Button Styles
 const BUTTON_TYPES = [
   'default',
@@ -106,7 +108,10 @@ class IdsButton extends mix(IdsElement).with(
     this.setIconAlignment();
     this.shouldUpdate = true;
     super.connectedCallback();
-    this.setAttribute('role', 'button');
+
+    if (!this.hasAttribute('role')) {
+      this.setAttribute('role', 'button');
+    }
   }
 
   /**
@@ -286,16 +291,19 @@ class IdsButton extends mix(IdsElement).with(
    * @param {boolean|string} val true if the button will be disabled
    */
   set disabled(val) {
+    const isValueTruthy = stringToBool(val);
     this.shouldUpdate = false;
-    this.removeAttribute(props.DISABLED);
-    this.shouldUpdate = true;
+    if (isValueTruthy) {
+      this.setAttribute(props.DISABLED, '');
+    } else {
+      this.removeAttribute(props.DISABLED);
+    }
 
-    const trueVal = stringUtils.stringToBool(val);
-    this.state.disabled = trueVal;
+    this.state.disabled = isValueTruthy;
 
     /* istanbul ignore next */
     if (this.button) {
-      this.button.disabled = trueVal;
+      this.button.disabled = isValueTruthy;
     }
   }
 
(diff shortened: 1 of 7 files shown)


tmcconechy (approved):

src/ids-button/ids-button.js
```diff
@@ -106,7 +108,10 @@ class IdsButton extends mix(IdsElement).with(
     this.setIconAlignment();
     this.shouldUpdate = true;
     super.connectedCallback();
-    this.setAttribute('role', 'button');
+
+    if (!this.hasAttribute('role')) {
```
tmcconechy, line 112: Suggest removing this entirely it works in Axe without it. So no role for now. This causes an issue in chrome vox where it doesnt read "button" but this works in voice over and other screen readers so this will probably get fixed at some point

```
  if (!this.hasAttribute('role')) {
    this.setAttribute('role', 'button');
   }
```

tmcconechy (changes requested):
Lets remove the role, lets just test it works on your end

tmcconechy (approved):

deep7102 (dismissed):

src/ids-button/ids-button.js
```diff
@@ -286,16 +287,19 @@ class IdsButton extends mix(IdsElement).with(
    * @param {boolean|string} val true if the button will be disabled
    */
   set disabled(val) {
+    const isValueTruthy = stringToBool(val);
     this.shouldUpdate = false;
-    this.removeAttribute(props.DISABLED);
-    this.shouldUpdate = true;
```
EdwardCoyle, line 291: This line is still needed if future updates to properties are to occur (a check for `shouldUpdate` wraps the property updates in this component).  Maybe move this to L297?

src/ids-button/ids-button.js
```diff
@@ -309,17 +313,16 @@ class IdsButton extends mix(IdsElement).with(
    * @returns {void}
    */
   set tabIndex(val) {
-    // Remove the webcomponent tabIndex
-    this.shouldUpdate = false;
-    this.removeAttribute(props.TABINDEX);
     this.shouldUpdate = true;
```
EdwardCoyle, line 316: This one can go as well

EdwardCoyle (changes requested):
@rob2d just a minor thing regarding the attribute changes here, but other than that this should be OK

rob2d:
@EdwardCoyle adjustments made to a reasonable extent here -- besides that there's no noticeable performance/functional issues that I observed and don't have workarounds for WC framework causing the setter issues. If you would like to spend time to address this I am perfectly ok with it but it isn't something I can be working on further this sprint and was a blocker for #139 that was a little out of scope here.

tmcconechy (commented):
Looks good to me if @EdwardCoyle is happy with it now

EdwardCoyle:
@rob2d Looks good to me now - just removed that one extraneous change to the `shouldUpdate` flag 👍

EdwardCoyle (approved):

tmcconechy (approved):

rob2d:
cool thanks @EdwardCoyle

PR #197, opened 2021-05-25
139 - ids input: optional labels and unique ids for WC accessibility

rob2d:

Adding some flexibility to the `ids-input` component with regards to labels and `id` assignment; mainly for the sake of WC + axe and screen readers to work naturally together with the hassle of ShadowDOM (at least for now, in 2021 🤞...).

Changes for the sake of flexibility and WC accessibility quirks on `ids-input` in this PR: 

(1) no `label` attrib removes `<label>` elem from shadowDom

(2) `id` prop is set to be incremented/unique by default. Input will also receive `${id}-input`.
--  future facing since puppeteer tests can be written with `pierce/#unique-id` for the overall document or any sub document.
-- tames axe issues when you want to link inputs using hierarchical aria rules or `label for`
-- There is a gotcha that `this.id` had to be set in the `template` vs in `connectedCallback` as an exceptional rule before first render -- otherwise we have to be very defensive in when we do things with the attrib as I learned in the `ids-spinbox`. There might be a better way to handle this but it was just the conclusion of going in circles previously on the issue.

(3) added a `label-hidden` attrib: setting this adds `aria-label` to the inner `<input>`. This also satisfies screen readers and axe tools if we don't want the label to be explicitly on an input in an app (say it is self explanatory or for some group/etc, or with IdsSpinbox + ChromeVox)

(4) renamed `label-text` class to `ids-label-text` + added `setLabelElement` prop which allows external label to get required asterisk/validation styling. 

Related to #39 and most of the changes help to satisfy making `ids-spinbox` accessible without axe issues (in addition to `ids-button` PR)

- pull this branch (`139-ids-input-optional-label`) and visit http://localhost:4300 and view input changes.
- check the examples for input and see that there's no regressive changes.
- verify that no `id`s on the `ids-input` components nor the inner `input` components have collisions in overall document scope when inspecting the main demo page markup.
image

- if the changes in PR aren't straightforward, another WIP branch which this is a blocker on is `139-ids-spinbox`. Pulling that and going to `http://localhost:4000/ids-spinbox` can show how it prevented Axe errors/simplified the label in a bit of a niche pattern with the WCs we haven't had to use yet.

### DIFF
--- a/app/ids-input/standalone-css.html
+++ b/app/ids-input/standalone-css.html
@@ -14,49 +14,49 @@
   <div class="ids-layout-grid-cell">
 
     <div class="ids-input">
-      <label for="input-first-name" class="ids-text label-text">First Name</label>
+      <label for="input-first-name" class="ids-text ids-label-text">First Name</label>
       <div class="field-container">
         <input class="ids-input-field" id="input-first-name" name="input-first-name" placeholder="Normal text Field"/>
       </div>
     </div>
 
     <div class="ids-input disabled">
-      <label for="input-disabled" class="ids-text label-text">Disabled</label>
+      <label for="input-disabled" class="ids-text ids-label-text">Disabled</label>
       <div class="field-container">
         <input class="ids-input-field" disabled id="input-disabled" name="input-disabled" value="Field not editable"/>
       </div>
     </div>
 
     <div class="ids-input readonly">
-      <label for="input-readonly" class="ids-text label-text">Readonly</label>
+      <label for="input-readonly" class="ids-text ids-label-text">Readonly</label>
       <div class="field-container">
         <input class="ids-input-field" readonly id="input-readonly" name="input-readonly" value="02006"/>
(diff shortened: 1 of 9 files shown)


tmcconechy (commented):
https://coveralls.io/builds/40009579/source?filename=src%2Fids-input%2Fids-input.js#L262 toss in a test to cover this? 
And then it looks good. Not to add to the scope but its going to be possible to add IDs in the future like this https://github.com/infor-design/enterprise-wc/blob/main/src/ids-tooltip/ids-tooltip.js#L32 so we might need to take this into account

rob2d:
after reviewing with Deepak there was some accessibility concerns with the `labels` (some nuance with WCs + shadowDOM) so am going to address that + tests here before re-opening.

tmcconechy (approved):

deep7102 (approved):
Looks good 🍌

PR #206, opened 2021-06-01
139 - ids-spinbox wc

### LINKED ISSUE #139: IdsSpinBox: Add new web component

rob2d:

Misc: also fixes:

a small potential issue with ids-keyboard-mixin where we sub on an array as the key vs the events (since we do not actually group key combos or use it in this manner in other components -- the fix was 2 lines so figured small enough to include here)

**Related github/jira issue (required):**

Closes #139

Steps necessary to review your pull request (required):

pull this PR branch locally and run environment (139-ids-spinbox)
visit http://localhost:4300/ids-spinbox/
ensure functionality listed at #139 for Phase 1 is implemented.

**Included in this Pull Request:**


tmcconechy (changes requested):
Just a couple things that could be improved (some could be moved to TODO).

1. Can you make sure it works in all themes see https://github.com/infor-design/enterprise-wc/tree/main/src/ids-mixins#ids-theme-mixin for details (all the components you did are missing this so should get in the habit)
2. The focus state is looking a bit strange
Screen Shot 2021-06-02 at 11 07 19 AM
3. We have a feature i just remembered where you can hold and press the spinbox 
- go to https://main-enterprise.demo.design.infor.com/components/spinbox/example-index.html
- hold down on the + button with the mouse 
- this is the code https://github.com/infor-design/enterprise/blob/main/src/components/spinbox/spinbox.js#L735 and general logic

My suggestion would be fix 1 and 2 and add 3 to the TODO for now but feel free to add

app/ids-spinbox/demo.scss
```diff
@@ -0,0 +1,13 @@
+#ids-spinbox-demo {
```
deep7102, line 1: Why not use `ids-layout-grid`?

app/ids-spinbox/example.html
```diff
@@ -0,0 +1,36 @@
+<ids-layout-grid>
+  <ids-text font-size="12" type="h1">Spinbox</ids-text>
+</ids-layout-grid>
+<div id="ids-spinbox-demo">
```
deep7102, line 4: Same about `ids-layout-grid`  not use.

app/ids-spinbox/example.html
```diff
@@ -0,0 +1,36 @@
...
+  ></ids-spinbox>
+  <ids-spinbox
+    value="0"
+    min="-25"
+    max="25"
+    step="5"
+    label="Required"
+    placeholder="N/A"
+    validate="required"
+    label-required
```
deep7102, line 33: does this `label-required` need? by default it should be `true`

deep7102 (commented):
@rob2d looking very nice, few things found 
- Why not used the use `ids-layout-grid` for demo example
- Missing read-only example
- Initial load focus was on last spin-input
- Input is fine, but each button around looks bit wider
- Type in `10` then click on ant button +/- value got reset
- In 0-to-5 spin box, it let enter other values like `60` 
- Missing getter/setter for `label`, `label-required`, `readonly` and `placeholder` props
- Do we need `label-required` in demo example? by default it should be true

app/ids-spinbox/demo.scss
rob2d: Was trying to make the demo responsive; was `ids-layout-grid` able to go to 2/3 based on media-query? if so can use that, was not familiar with it but didn't see an immediate/obvious way to do that.

app/ids-spinbox/index.js
```diff
@@ -0,0 +1,2 @@
+import IdsSpinbox from '../../src/ids-spinbox';
+import './demo.scss';
```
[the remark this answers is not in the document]
rob2d: not sure if this is user error here, but tried to put this into `app/ids-spinbox/example.js` and it doesn't actually reflect when visiting the main app index.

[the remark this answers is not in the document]
tmcconechy: I did it here https://github.com/infor-design/enterprise-wc/tree/main/app/ids-tooltip but not sure about the main index either. You maybe need to import it index.js but sounds like this is for layout? If so i would just remove it and use ids-layout grid?

app/ids-spinbox/demo.scss
rob2d: switched to ids-grid, thanks

app/ids-spinbox/example.html
rob2d: it seemed like input wasn't label-required by default, so was just doing this for consistency sake

rob2d:
@deep7102 thanks for the feedback.

Re:  `Type in 10 then click on ant button +/- value got reset`

I saw in the issue it said handling blur/etc was not for Phase I, so I was aware of that but just pushing with functionality as it is since that's a related edge case.

rob2d:
@deep7102 seems placeholder and label setters are there

for

> Initial load focus was on last spin-input

I believe this is a separate issue/slightly out of scope and on `ids-input` at the moment since I'm not adding code for managing focus state beyond when user has retrieved it. Can see this when visiting http://localhost:4300 (document.activeState is on "active focus" component but that shouldn't be auto binding)

PR #211, opened 2021-06-03
139 - ids spinbox wc

### LINKED ISSUE #139: IdsSpinBox: Add new web component

rob2d:

Misc: also fixes:

a small potential issue with ids-keyboard-mixin where we sub on an array as the key vs the events (since we do not actually group key combos or use it in this manner in other components -- the fix was 2 lines so figured small enough to include here)

**Related github/jira issue (required):**

Closes #139

Steps necessary to review your pull request (required):

pull this PR branch locally and run environment (139-ids-spinbox)
visit http://localhost:4300/ids-spinbox/
ensure functionality listed at #139 for Phase 1 is implemented.

**Included in this Pull Request:**


rob2d:
@tmcconechy  & @deep7102 thanks for the useful feedback on the previous PR. Sorry to not realize this but if you edit a closed PR in a certain way, it can't be re-opened. So just for reference the review on previous PR can be found @ #206 .

Tried to be thorough to address all comments, except for:
> @deep7102 - Type in 10 then click on and button +/- value got reset

In this case I need to add some custom key-down listener/possibly blur, but I did fix a bunch of other edge cases on value. Based on suggestion of scope on the issue I hadn't done this yet just to get another round.

And:

> @tmcconechy hold and press a spinbox ... hold down on the + button with the mouse

On this feature I put a `TODO.md` note here.

Will see if both of these are feasible in current but we're also using different patterns + WC so may be muddy. Will be addressing while this is open again here time-boxed to this evening (after investigating failing CI build).

rob2d:
@deep7102 @tmcconechy should be ok to look at again. 

Was able to get value cycling and also made the timing speed up a little as you hold down the mouse on the increment/decrement buttons, but lot of gotchas for mousedown=>up tests and they needed rework to get any useful coverage; figured since adding helpers to finally move the other time-related helpers as so we can DRY other stuff later but was getting out of scope to continue it.

@deep7102 addressed some things on rounding the value to the step and other edge cases but the specific key down on numeric vs just handling change before button isn't there since above required some time (also a little hesitant on the consideration on this since if it's auto validating it may be frustrating for a user e.g. typing "35" w/ step == 5. Would jump from 3 to 5... so not impossible just may take time to not add frustrating side effects if done straightforwardly).

tmcconechy (approved):

EdwardCoyle (approved):
@rob2d Looks great! Very nice work 👍

tmcconechy (dismissed):

deep7102 (approved):
Thanks! @rob2d
looks great 🍌

tmcconechy:
Since every one approved im merging

PR #214, opened 2021-06-14
5254 - Swipe container

tmcconechy:

Adds a swipe component with actions on the left and right side. And some supporting features including

- a setting in cards to make them the height of content `autoHeight`
- a setting in grid cells to make the content them "float right" `justIfy`
- a setting in button to hide the ripple

The loading experience of the reveal style action is shows the container slide over on page load but couldnt solve this in a reasonable amount of time.

TODO: Port this back once reviewed

Fixes #5254

- go to http://localhost:4300/ids-swipe-action/
- in firefox you can just use the actions button (not wired)
- in chrome you can use the actions button (not wired) or try and swipe left and right with the magic mouse
- Main Point: Try this in ios and android and you can swipe left or right and click the button on the reveal example
- Main Point: Try this in ios and android and you can swipe left or right and the action will be taken on the continuous example


rob2d:
Haven't fully dove into the code but some things I notice on the UX for the last (continuous) setting example:

- it seems the opposite side is revealed to what is selected. Maybe when there is a continuous mode set, the left and right should position themselves flipped? or it could be an inverse drag? (e.g. swipe left to drift right, similar to default MacOS + mousewheel)... otherwise it's a little bit confusing (I do get that you're swiping the direction that usually is revealed, just you don't see the action selected here).
- something that adds to the above: when you first swipe without letting the gesture go, it auto-chooses something, but it seems this should happen on touchend vs while touch is down. So if a user lets go of touch that selection should be made just in case a user is just looking to see what options are available.

Codewise only thing that is standing out is the component isn't using the `index.js` pattern for it's module, though I thought we were moving to it after discussion a while ago.

Outside of that I haven't gotten a chance to look through exhaustively but component itself looks aesthetic and seems to work very nicely!

tmcconechy:
Thanks @rob2d 
- fixed the direction issue - i missed that good catch
- fixed the index.js pattern issue - also good catch
- im inclined to leave the touchend issue for now, as i see it working to me like other apps like outlook or gmail on my phone. but lets keep this in mind in case others have same feedback

rob2d:
Ok thanks @tmcconechy looked over this AM and LGTM!

rob2d (approved):

deep7102 (approved):
Looks good 🍌

Issue #140, opened 2021-05-04
Misc: Find performance improvements


tmcconechy:
**Describe the bug**
Look for web component improvements and make some tests.

**Tasks**

- [x] Document the tooltip perf test once merged
- [x] Add a datagrid perf test
- [x] Change from Constructed style sheets to links and test
- [x] Try adding https://developer.mozilla.org/en-US/docs/Web/CSS/contain https://developers.google.com/web/fundamentals/web-components/shadowdom#containment and testing as in

```
:host(ids-tooltip) {
  contain: content;
} 
```
- [x] See warnings on animation frame on http://localhost:4300/ids-list-view
- [x] Test load in fire fox as well

**Expected behavior**
Faster rendering

**Additional context**
Team Call...
See: https://blog.webpagetest.org/posts/benchmarking-javascript-memory-usage/

PR #215, opened 2021-06-14
Quick Optimization: Memoize prop-name conversions for IdsElement attribChangeCallback

rob2d:

When setting any prop in IdsElement, we were going through `number-of-chars` ^ 2 or 3 before being able to run through normal setter logic as we were converting `attrib-name` to `attribName` for property assignment. 

Through some benchmarking, found that using memoization could get that down a bit. In some sample test code (shared in MSTeams), running a similar operation gave these results on my work PC:

```
ITERATIONS: 10
withoutMemo: 0.207ms
withMemo: 0.037ms
change: ~540%

For 100 iterations:

ITERATIONS: 100
withoutMemo: 0.331ms
withMemo: 0.057ms
change: ~580%

ITERATIONS: 3000
withoutMemo: 4.322ms
withMemo: 0.226ms
change: ~1900% (!)
```

Note that these tests are much more simple and repeating one property, real world scenario should be better.

When loading `http://localhost:4300`, only on initial render we had *3138 cache hits* (times that we could re-load the var without running `camelCase` op), and for larger production apps this will likely be larger. So we are saving some time in interactions or load up (especially on slower PCs or with multitasking) which may be useful to help us maintain easy 60fps or not worry about less critical things.

Not listed on #140  but related. Most similar is to how we can optimize style loading on each component instance.

- N/A, app should run as it normally does and just need to confirm tests pass. Change is minor/mainly for perf with real app usage.


src/ids-base/ids-element.js
```diff
@@ -11,6 +11,15 @@ import renderLoop from '../ids-render-loop/ids-render-loop-global';
 import IdsRenderLoopItem from '../ids-render-loop/ids-render-loop-item';
 import { stringUtils } from './ids-string-utils';
 
+/**
```
tmcconechy, line 14: This is minor but we could probably be more concise here..
```
/** Dictionary used to cache attribute names */
```

tmcconechy (dismissed):

tmcconechy (approved):

deep7102 (approved):

Issue #209, opened 2021-06-03
IdsInput: Focuses the last input on the page


tmcconechy:
**Describe the bug**
This may be a newly introduced issue but the last input on the page is focused.

**To Reproduce**
1. Go to http://localhost:4300/ids-input
2. Scroll down if needed
3. See that the last input on the page is focused

**Expected behavior**
Nothing is focused by default

**Platform**
 - Chrome or any
 
**Additional context**
Noticed reviewing https://github.com/infor-design/enterprise-wc/pull/206

PR #216, opened 2021-06-14
209 - fix ids-input grabbing focus on each value set

rob2d:

When `ids-input` gets any value set (including when it first is loaded), it attempts to blur, then re-grab focus on itself. This is causing interaction issues on certain components as a default behavior and also causing inputs to grab focus when components mount.

Fixed this by removing the `focus` => `blur` => `focus` toggling. Seems to have no side effects on typical inputs, and that possibly components which it was used for should probably do this vs rely on `ids-input` to handle this behavior.

Pushed this particularly because it was also interfering with any behavior on `ids-pager` component; when clicking on any nav button it was forcing focus back to `ids-input` which seemed not to be an intentional default as input sometimes only is also used to reflect value vs set value.

Fixes https://github.com/infor-design/enterprise-wc/issues/209

- visit http://localhost:4300
- see that no ids-input is not automatically focused on.


tmcconechy (approved):
Nice, i couldnt figure this one out. Linked the issue we had for this bug

rob2d:
@tmcconechy whoops sorry, had the issue number off there. Edited.

deep7102 (approved):

PR #226, opened 2021-06-25
Misc: Rename instances of events from past-to-present tense for consistency

### LINKED ISSUE #213: General: Remove `ed` from all event names
**Describe the bug**
Because we have a mix of events with ed and not ed at the end. We settled on not having the `ed` because some native events dont use it and this will make it more consistent

**To Reproduce**
1. Search for `triggerEvent`
2. Anyone with `ed` at the end should be refactored
3. Make sure the tests work

**Additional context**
Ms teams convo

rob2d:

Since the HTMLDOM events use present tense e.g. `change` `click`, etc, we decided after a team conversation to make events consistently present tense e.g. `activecellchange` vs `activecellchanged`.

Closes #213 

- tests should pass

N/A

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -90,7 +90,7 @@
   - Markup has changed to a custom element `<ids-tabs value=${selection}></ids-tabs>`
   - content within the tabs are specified as `<ids-tab value=${selection-value}>`Tab Label/Content`</ids-tab>`
 - `[Tags]` The Tag component has been changed to a web component.
-  - If using events events are now plain JS events. beforetagremoved, tagremoved, aftertagremoved
+  - If using events events are now plain JS events. beforetagremove, tagremove, aftertagremove
   - If using properties/settings these are now attributes: dismissible, color
   - Markup has changed to a custom element `<ids-tag color="error">Text</ids-tag>`
   - Can now be imported as a single JS file and used with encapsulated styles
--- a/src/ids-data-grid/ids-data-grid.d.ts
+++ b/src/ids-data-grid/ids-data-grid.d.ts
@@ -2,7 +2,7 @@
 // confirm our code is type safe, and to support TypeScript users.
 import { IdsElement } from '../ids-base';
 
-interface sorted extends Event {
+interface sort extends Event {
   detail: {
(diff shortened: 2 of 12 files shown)


tmcconechy (approved):

EdwardCoyle:
The changes in this PR look great!  I have a question about another event we should probably normalize though...

IdsMenuItem will currently [trigger `selected` and `deselected` events](https://github.com/infor-design/enterprise-wc/blob/main/src/ids-menu/ids-menu-item.js#L555) from menu items, based on the scenario:  
- In single-select situations, only the "clicked" item will triggered a `selected` event, and any previously-selected items will trigger a `deselected` event.  
- In multi-select situations, the same occurs, while also triggering `deselected` event on menu items that are "clicked" and also previously selected.
- Additionally, there are `before` and `after`-prefixed events for both of these, similar to `beforetagremove`, `aftertagremove`, etc.

I could see us normalizing these event names in the same way... however, I am realizing there's another DOM event that goes by the same name for a different purpose -- [text selection on Input fields](https://developer.mozilla.org/en-US/docs/Web/API/Element/select_event#selection_logger). Maybe it's not a big deal if we had an event by the same name, but wanted to bring this to your attention so that we:
- cover the case of renaming everything uniformly
- make sure having a `select` event coming from our Menu Items would be an issue if it's a completely "different" type of event than the built-in one.

I'm indifferent about what the event name is as long as it makes sense for what it's doing... just want to see what we all think?

tmcconechy:
@EdwardCoyle My opinion would be to normalize that one as well to select/deselect on ids-menu-items. I think it will be ok as along as we dont need a select event on input and textarea elements since from the docs:

> For example, in HTML, select events can be dispatched only on form <input type="text"> and <textarea> elements.

So think we would only run into an issue there.

rob2d:
@EdwardCoyle that seems like a good consideration to make. Just in my opinion on this subject, it seems that using `select` is ok for an event listener interface -- especially if it's the standard. I think in that case a user can check the event details and possibly propagating that with some meta data if needed for diffing/etc.

(I noticed we already merged just figured to chime in on that -- thanks for the review guys)

Issue #131, opened 2021-03-15
IdsColorPicker: Add new web component


tmcconechy:
**Describe the bug**
Add the colorpicker  http://localhost:4000/components/colorpicker component to the WC components. 

**Features to Bring Over**
- custom labels setting
- show label setting
- ability to set colors (will make a simple color component that shows a background color in the palate or any hex)
- all normal input states (see input components i.e. dirty, validation, enable, disable ect)

**Suggested Markup**

```
<ids-colorpicker value="#737373">
 <ids-color hex="800000"></ids-color>
 <ids-color hex="800001"></ids-color>
 ....
</ids-upload>
```

**Additional context**
- 100% test coverage
- DOCS, CHANGE log ect. 
- https://github.com/infor-design/enterprise-wc/blob/master/doc/CHECKLIST.MD
- themes 

See completed components like card, tag, badge, hyperlink as reference

tmcconechy:
Closing this for round one

PR #228, opened 2021-06-28
131 ids colorpicker wc

danielortiz1982:
Creating a ids-color-picker web component. The color picker web component can be used by adding a custom element <ids-color-picker></ids-color-picker> on to any webpage or frontend application.

#131 

Include:
- npm checkout 131-ids-colorpicker-wc
- npm ci
- npm run build
- npm start
- Navigate to the ids-color-picker example page (http://localhost:4300/ids-color-picker/)

**test scenarios**:
- Click on click on the color swatch to view advance options to choose a color. Select a given color value to update the color picker value.
- Click on the pointer icon on the right side of input to open and close the color option panel. Select a given color option value to update the color picker value.
- Type a given value inside the input to update the value and change the color of the swatch.

**Loom Video**:
https://www.loom.com/share/1a8a9509e25643cda39e35c96aa453e9

**Screenshots**
Screen Shot 2021-06-30 at 10 12 28 AM
Screen Shot 2021-06-30 at 10 14 10 AM
Screen Shot 2021-06-30 at 10 15 46 AM


src/ids-color-picker/TODO.md
```diff
@@ -0,0 +1,13 @@
...
+- [ ] Fix focus state around input and swatch (needs trigger button issue)
+- [ ] Fix the keyboard (arrow down , keys around the colors ect)
+- [ ] Click the swatch to select / option for the "advanced" picker
+- [ ] Fix arrow position
+- [ ] Dirty tracker
+- [ ] Empty State
+- [ ] Clear Swatch
+- [ ] Input should be smaller size
+- [ ] Fix all the states [See example](https://main-enterprise.demo.design.infor.com/components/colorpicker/test-states.html)
+- [ ] Click out close (popup issue)
```
EdwardCoyle, line 13:
```suggestion
- [ ] Click out close (popup issue -- will be fixed in #225)
```

src/ids-color-picker/TODO.md
```diff
@@ -0,0 +1,13 @@
+# Ids Color Picker TODO's
+
+- [ ] Tooltip on the colors
+- [ ] Fix focus state around input and swatch (needs trigger button issue)
+- [ ] Fix the keyboard (arrow down , keys around the colors ect)
+- [ ] Click the swatch to select / option for the "advanced" picker
+- [ ] Fix arrow position
```
EdwardCoyle, line 7: - [ ] Enable/Disable standard HTML colorpicker with boolean attribute
- [ ] Standardize/Abstract out Popup/Picker/"return value" interface (to be used in this and other pickers)

EdwardCoyle (commented):
Looks really nice so far!  Functionality and style are both there.  I think it just needs a few things (will approve after these are done):

- [ ] Need to add JSDoc comments for everything that's missing
- [ ] Fix lints

Thinking about what comes next, these two things I'd like to maybe discuss for future PRs (will approve without these):

- [ ] (Possibly a phase 2) There are use cases for Colorpicker where we want users to be forced to pick an available swatch (as opposed to choosing any random color from the native picker).  If we are going to allow access via the native picker, we may also want to have a boolean setting so it can be enabled/disabled independently of the rest of the component.  I added this in a suggestion to the TODO list below
- [ ] (Possibly for phase 2) We might want to figure out how to abstract out the structure of linking an IdsPopup to an IdsTriggerButton/Field, and returning a value from the IdsPopup's contents.  The same structure here in the Colorpicker will also apply to Time/Datepicker (and potentially others), and it would be nice if all of these pickers operated in the same predictable manner.

src/ids-color-picker/TODO.md
danielortiz1982: Thanks for your feedback @EdwardCoyle. @tmcconechy has a similar idea. I will see what type effort is needed to achieve this for a round 2 PR.

tmcconechy (commented):
in addition to Ed's comments if you just fix coverage (or skip) on https://coveralls.io/builds/41013226/source?filename=src%2Fids-color-picker%2Fids-color-picker.js#L63

should be good enough for now for phase 1

deep7102 (changes requested):
Hi @danielortiz1982 this looks good found some points as:
- Addition to @EdwardCoyle and @tmcconechy comments
- Some text hard coded in component, should give setting/slots to user change those text strings (saw in template())
- Styles not match with 4.x (popup container) - swatch/hover/spacing etc [see example here](https://main-enterprise.demo.design.infor.com/components/colorpicker/example-index.html)
- Did not see the tooltip
- Why container need hard coded width/height (saw 360px/250px in css)
- Markup should be simpler, default color list should load from our theme palette
- `<ids-color-picker label="Default Colors"></ids-color-picker>`
- Custom color list can pass separately as json style dataset (can set dynamically or through a DataSource)
- `[{ label: 'Slate', number: '08', value: '383838', border: 'dark', arrowClass: 'dark' }]`
- Check mark color should change based on swatch background (check mark color can get from colorlist -arrowClass)
- Should be able to show `color-name` azure06 or `color-hax` #ffffff [see example here](https://main-enterprise.demo.design.infor.com/components/colorpicker/example-custom-labels.html)
- Did not see other examples are we going to add them later or in this phase (validation, readonly) [see example here](https://main-enterprise.demo.design.infor.com/components/colorpicker/test-states.html)

tmcconechy (commented):
Actually too more things that are probably important.

1) Can you test themes (or did you) might need some styles here
2) Can you add a test that the change event fires when you change it? It might because thats in input already but thats rather important. Shoud be a test there.

Also still failing on builds/coverage so just add a skip

tmcconechy:
I added some more TODO's based on the comments. I guess im ok with approving this for now but we will need to circle back as there are a lot of important TODO's

tmcconechy (approved):

danielortiz1982:
> I added some more TODO's based on the comments. I guess im ok with approving this for now but we will need to circle back as there are a lot of important TODO's

Thanks Tim.

EdwardCoyle (approved):
I'm approving as well, acknowledging we still have quite a bit to finish here

deep7102 (approved):
I'm approving too, hope will get thru most important TODO's next circle

PR #232, opened 2021-06-29
214 - Adds fixes for swipe container and more

tmcconechy:

This PR contains:

- fix / improvements to timeouts on block grid tests 
- adds option to flip the actions button
- fixes non workingnoRipple setting
- removes unneeded render in ids-text
- comments out console.errors as the are effecting tests so we should decide on direction here
- changes to swipe container
   - close on click
   -  flip the icons
   - hide the ripple

Relates to #214
Started from https://github.com/infor-design/enterprise/pull/5374

- go to http://localhost:4300/ids-block-grid/ should work as before
- go to http://localhost:4300/ids-swipe-action
- swipe the first box and click -> should now close
- icons should be flipped
- clicking should not show a ripple


rob2d (commented):
@tmcconechy looks good. 

Only thing is for flags e.g. `vertical`, it seems there's some cases where it would accidentally trigger issues e.g. if I set the flag with an empty string it would `|| false` (ln 94 on `ids-icon.js`). Looked over at HTML spec @ https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes
and there's one thing particularly they say (which `aria` attributes are just misleading for):

> The values "true" and "false" are not allowed on boolean attributes. To represent a false value, the attribute has to be omitted altogether.

I found recently that dealing with flags are a little easier via `hasAttribute` and then setter setting to `''` has helped in those cases personally so far after noticing that, but if that's too drastic to change then just `stringToBool` in the setter/getter (ln 97 @ ids-icon) at least prevents unintended checks for falsy coming as truthy in some scenarios. Mainly mentioning since I bumped into some cases where component attributes didn't work as flags though even they were boolean e.g. `ids-layout-grid auto`.

rob2d:
Looked at the failing tests and not quite sure what is up with that... it doesn't look like any obvious changes on this branch would trigger this. Maybe has to do with Axe changing what it expects to satisfy the region role issue?

tmcconechy:
Thanks @rob2d i fixed the tests and applied the stringToBoolean , thats a good catch. I didnt want to change it though the other way as i think this might be easier to bind to in NG ect and its only veery going to be `vertical="true"` or not there.

rob2d (approved):
thanks @tmcconechy LGTM 🍌

tmcconechy:
Yeah we could remove them. I added this as a topic for next call. I think the validation is a nice idea but it was causing errors when running the tests so wanted to map out a strategy and discuss how much validation and when we do it.

EdwardCoyle (approved):
👍  Looks really nice

Issue #201, opened 2021-05-26
IdsLoadingIndicator: Design Enhancements and new Features


tmcconechy:
Based on the design at https://www.figma.com/file/PwwLkUtTxpxLzRJzm5kACM/IDS-Progress-Indicators?node-id=0%3A1 we want to build out some of the elements and port some of it as noted.
See https://material.io/components/progress-indicators#usage for inspiration

We want to also port the design changes back to 4.x . I.E. whats "design changed" and the line loader styles (phase 1 items).
 
**Loader (Phase 1)**
- Use these designs and update ids-loader
- port this back to 4.x as a new loader component
- show under the header and on cards

**Line Loader (Phase 1)**
- Compare with what is there now.
- Update 4.x with these designs

**Spinners (Phase 1)**
- Will call this circular progress indicator (ids-progress-indicator type="circular") - P3
- Indeterminate vs with a progress count - P5 for progress count (indeterminate="false" value="40%")
- size (small for fields / vs normal)
- show on a card and show on an input
- Update 4.x with these designs

**Other Tasks (Phase 1)**
- Lets rename `ids-progress` to `ids-progress-bar` for clarity

**Progress Buttons (Future)**
- Skip for now

**Skeleton (Future)**
- skip for now 

**Suggested Internal Markup**

```js
<ids-loading-indicator percent="10" type="circular" determinate="true"></ids-loading-indicator>
<!-- Centered in the Card -->
<ids-loading-indicator type="linear"></ids-loading-indicator>
<ids-loading-indicator type="circular"></ids-loading-indicator>
<!-- Top Page Loader on a Card or Page/Container or Datagrid-->
<ids-loading-indicator type="bar"></ids-loading-indicator>

<!-- Future -->
<ids-loader type="skeleton"></ids-loader>
```

or
Loader:
- Skeleton Loader
- Circle  Loader
- Line  Loader
- Etc Loader

**Suggested Dependencies**
NA

**Settings**
- indeterminate
- type

**Design**
- https://www.figma.com/file/PwwLkUtTxpxLzRJzm5kACM/IDS-Progress-Indicators?node-id=0%3A1

**Events**
- updated (name)
- finished
- started

**Other**
Pass the following checklist ordered in priority (some may be left for later) https://github.com/infor-design/enterprise-wc/blob/main/doc/CHECKLIST.md

rob2d:
Thinking on it a little while trying to create example markup, I have a few ideas for possible iteration:

(1) type

[...]

PR #237, opened 2021-07-02
201 - ids-loading indicator

rob2d:

This PR adds the `ids-loading-indicator` design changes on the WC side.

Partially Addresses #201
There's also a redesign portion for existing `enterprise` repo so it doesn't close out the issue.

- after pulling the branch, visit http://localhost:4300 or http://localhost:4300/ids-loading-indicator
- make sure all tests pass and that the functionality is satisfactory


test/ids-loading-indicator/ids-loading-indicator-func-test.js
```diff
@@ -0,0 +1,177 @@
...
+  expect(elem[camelCasedAttrib]).toEqual(true);
+
+  elem[camelCasedAttrib] = false;
+  expect(elem.hasAttribute(attribute)).toEqual(false);
+};
+
+describe('IdsLoadingIndicator Component', () => {
+  let elem;
+
+  const createElemViaTemplate = async (innerHTML) => {
```
EdwardCoyle, line 40: @tmcconechy (maybe for a future PR) This seems like something we should do in any tests where we insert markup instead of using `new`... This should probably end up being moved to a helper function at some point.

EdwardCoyle (approved):
Looks really good!

deep7102 (approved):
🍌 This is nice addition, Thanks! @rob2d

tmcconechy (changes requested):
The only thing i'm wondering about in the example is why the 5 right most ones aren't doing anything?

https://drive.google.com/file/d/1Ju5_bOIEqtDXPUUjQisWQEye2utqaxal/view

Can we make those start up too? Everything else looks wonderful. 

Do we need any TODO's for things like attaching it to other components?
I'd also like to see an example with it at the very top of the page, in a card and in a datagrid for that.

rob2d:
@tmcconechy the ones on the right are determinate e.g. just representing a loader with specific percentage complete, so in a real app they would be moving... I couldn't tell from the figma spec since they don't animate but I wasn't aware of any functionality on it. 

Would it help to add some buttons to change the % for each of these? 

For the top of the page I didn't see we had a central index so I wasn't sure on adding lots of example pages since people wouldn't find it but I can add one.

tmcconechy:
yeah maybe a button that loops through and runs them all? and i guess skip the card/datagrid/page examples for now but we probably want them eventually.

rob2d:
@tmcconechy added some spinboxes to the examples. 

Needed to add a `label-hidden` attribute to `ids-spinbox`, just because things looked super off-balance with it in this scenario even though it seemed like an idiomatic way to adjust them (until we have full sliders).

image

tmcconechy:
@rob2d I just checked with design and the fact that the labels(35% / 72% ) are graphite 03 is a mistake. I noticed this since they are hard to see.

This was made before we got rid of graphite as a general rule. So can you make these Slate 60 instead? (the default text color)

rob2d:
@tmcconechy thanks for the catch, adjusted the palette.

tmcconechy:
This issue was previously approved so merging

PR #248, opened 2021-07-09
242 ids container padding

### LINKED ISSUE #242: IdsContainer: Remove Padding and make it a setting
**Describe the bug**
The general idea is to remove the internal padding from ids-container so things like ids-header that have 100% height and width will stretch to the outside. This should be the default

**Tasks**

- [x] Make this padding a setting https://github.com/infor-design/enterprise-wc/blob/main/src/ids-container/ids-container.scss#L16 and remove
- [x] Add new setting (suggest simply padding) https://github.com/infor-design/enterprise-wc/blob/main/src/ids-container/ids-container.js#L60 that is an number. But generally would use 8, 16, 32, default is zero/none
- [x] Add doc for it
- [x] Add test for it
- [x] Confirm ids-header works as expected

**Expected behavior**
N/A

danielortiz1982:
Adding a new padding setting to the ids-container wc.  This setting will have a new setter/getter that will set a style padding with the given value. 

closes #242 

Include:
- git checkout 242-IdsContainer-padding
- npm ci
- npm run build
- npm start
- Navigate to the main page http://localhost:4300/
- open up the inspect chrome dev tools
- find any <ids-container> custom element 
- add or change the padding attribute example: padding="8"
Screen Shot 2021-07-12 at 4 03 44 PM
Screen Shot 2021-07-12 at 4 03 24 PM


src/ids-container/ids-container.js
```diff
@@ -35,14 +35,16 @@ class IdsContainer extends mix(IdsElement).with(IdsEventsMixin, IdsThemeMixin) {
    */
   connectedCallback() {
     super.connectedCallback();
+    // eslint-disable-next-line no-self-assign
+    this.padding = this.padding;
```
tmcconechy, line 39: Why do we need to do this again? Didnt have to for scrollable option

src/ids-container/README.MD
```diff
@@ -24,14 +24,23 @@ A normal container used as a web component.
 </ids-container>
 ```
 
+A normal container with a set padding.
+```html
+<ids-container scrollable="true" padding="8px">
```
tmcconechy, line 29:
```suggestion
<ids-container scrollable="true" padding="8">
```

this is up for discussion but wondering if we should just assume always pixels? but maybe we might want other options?

src/ids-container/ids-container.js
```diff
@@ -53,6 +55,18 @@ class IdsContainer extends mix(IdsElement).with(IdsEventsMixin, IdsThemeMixin) {
     return `<div class="ids-container" part="container"${this.scrollable === 'true' ? ' tabindex="0"' : ''}><slot></slot></div>`;
   }
 
+  /**
+   * If set to number the container will have padding added
+   * @param {string} value sets the padding to the container
+   */
+  set padding(value) {
+    this.container.style.padding = value;
```
tmcconechy, line 63: I the case of zero could just not set the inline style. (its one less hit on the DOM)

tmcconechy (commented):
Some thoughts... Looks correct tho

src/ids-container/README.MD
danielortiz1982: I tried both and landed with a string as a value. Let's have a side chat about that. I would to show you my findings on it.

src/ids-container/ids-container.js
danielortiz1982: Let me see how I can make that happen, I think it's being set by the setter.

src/ids-container/ids-container.js
danielortiz1982: Not sure where the issue lies, but the setter and getter do not work as expected until  you do this.

src/ids-container/README.MD
tmcconechy: Sure lets chat i just to clarify mean the "px" part...Unless we might need `em` `rem` ect

tmcconechy:
@danielortiz1982 is this ready? If so please add PR information for starts and take it out of draft?

src/ids-container/README.MD
```diff
@@ -24,14 +24,23 @@ A normal container used as a web component.
...
+```html
+<ids-container scrollable="true" padding="8">
+  <!-- ids components -->
+</ids-container>
+```
+
 ## Settings and Attributes
 
 - `scrollable` {boolean} Turns on the overflow to allow for scroll
+- `Padding` {number} Sets the padding in px around the Ids Container.
```
tmcconechy, line 37:
```suggestion
- `padding` {number} Sets the padding in px around the Ids Container.
```

tmcconechy (changes requested):
This isn't quite right. I think we need the width to be

1) width: calc(100% - padding)
2) Also you are no longer negating the browser 8px margin so we need to keep that part as well.

So i think it needs

a) the margins back
b) set the width to calc(100% - 8px + padding)
c) also the same with the height

Also if padding is zero (default) i dont think we should need to set anything

tmcconechy:
@danielortiz1982 ok will take i look i might have gotten the logic im saying wrong

tmcconechy:
@danielortiz1982 added the margin and defaults back. I think this way the percy tests should be the same as before and we have the reset and width set

tmcconechy (approved):
Made necessary changes.

PR #255, opened 2021-07-14
121 - Add locale functionality and locale mixin

### LINKED ISSUE #121: IdsLocale: Add new mixin
**Describe the bug**

Bring over the current locale component.  See if we can simplify and improve the code at all while improving as we go. Current locale component https://master-enterprise.demo.design.infor.com/components/locale

```
<ids-locale-switcher locale="de-DE" language="se"></ids-locale-switcher>
<ids-data-grid locale="de-DE" language="se"></ids-datagrid>
```

- Supports numbers
- Supports dates
- Supports translation
- Supports right to left 
- Figure out how to dynamically load the scripts

**Additional context**
- 100% test coverage
- DOCS, CHANGE log ect. 
- themes 

tmcconechy:

This PR adds locale from 4.x but makes some improvements namely:

- works with async
- a lot less code and used more browser apis (number.toLocaleString and Intl.DateTimeFormat)

In addition added some locale features to some but not all components:

- ids-text can translate
- ids-data-grid can format cells
- some RTL conversion
- ids-icon can flip directional icons

This should have most of what we need to move forward.

In addition:
- found more places in text to change properties to attributes

Closes #121 

**Translations**
- go to http://localhost:4300/ids-text/translations
- on the ids-container change the language to 'ar' 
- see that it goes RTL and translates the text
- on one ids-text element change the language to 'de' 
- see that it can translate independently
- go to http://localhost:4300/ids-locale
- on the ids-container change the language to 'fi'  and see that it translates the text

**Misc**
- one simple test page http://localhost:4300/ids-locale/numbers (not live for changing numbers)
- Thousands of lines of tests pass

**Datagrid Formatters**
- go to http://localhost:4300/ids-data-grid
- on the container or ids-data-grid change locale to `de-DE` and see that the numbers and dates change in the grid
- on the container or ids-data-grid change language to `ar` and see that it works RTL

**RTL**
- go to http://localhost:4300/ids-accordion
- on the container or ids-accordion change language to `ar` and see that it works RTL
- go to http://localhost:4300/ids-icon
- on the container or ids-data-grid change language to `ar` and see that it works RTL and notice in particular that directional icons change direction

- [x ] A note to the change log.


src/ids-locale/TODO.md
```diff
@@ -0,0 +1,23 @@
+# TODO For Locales
+
+- [x] Promises/Await
```
rob2d, line 3: not sure since these are X'd off should we keep them here? 

I think it is good to see a list of functionality as a check-off but I noticed most components just remove from the TODO when in so that's the only source of consideration with making this comment here.

test/ids-locale/ids-locale-func-test.js
```diff
@@ -0,0 +1,1783 @@
+/**
+ * @jest-environment jsdom
```
rob2d, line 2: for first iteration these tests are really nice/simple/readable. Just for consideration though: it seems like it's quite a *lot* of repeating similar calls, so was thinking it could be good in future iteration to have a lookup map if that isn't convenient now?  e.g. maps of `Map<String, Set<String>>` with `Map<'en' => {'AU', 'GB','IN', 'NZ', 'US', ...}`>`

Could make it a little easier to add to or change as well but this also depends on the frequency that clients request changes to this sort of thing here (which I'm not familiar enough to weigh in on).

rob2d (commented):
@tmcconechy this was a really large PR but from what I can tell the patterns/approach/etc in the context of what we have are really nice! 

Hopefully I did not miss anything in reviewing but I put some comments that are mostly just minor where I could.

Again though looks really great for a first PR in the WCs with this and this is a pretty big feature handled this extensively at once 👏

src/ids-mixins/README.md
```diff
@@ -168,4 +168,66 @@ get providedAttributes() {
...
+1. Add IdsLocaleMixin to the @mixes list
+
+When using it access the locale with `this.locale`.
+
+1. Test it by adding for example `tooltip="Additional Information"` on the component.
+1. Add tests for your locale use cases to your component tests.
+1. If you need to respond and change things on language or locale change then you may need to add a combination of the following to handle events
+
+```js
+// Respond to parent changing language
```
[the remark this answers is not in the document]
tmcconechy: good catch

test/ids-locale/ids-locale-func-test.js
tmcconechy: I might have to add this for "validating its only in the accepted list in the future. But i think i'll punt on this for now. I sort of tried to have it "open". Although that might not totally make sense in the future. WIll keep this one in mind.

tmcconechy:
@rob2d made changes for all your comments. Good catches. (except i skipping the last one for now)

rob2d (approved):
Changes look good @tmcconechy ! (approved)

EdwardCoyle (approved):
Looks really nice @tmcconechy!

PR #259, opened 2021-07-15
ids-text: fix right-click crashing Chrome

rob2d:

Fixes an issue with IdsText where when it is arranged in certain patterns and right clicking on the non-whitespace characters, would crash Chrome browsers. 

Since IDSText is meant to simply be a content tag type with styling and a user defined slot, a workaround for now was to render `innerHTML` vs a `slot` in the shadowDOM directly.

N/A, was a known issue but we did not have a tracker for it yet. Fixed while working on `ids-draggable` since it was affecting efficient debuggability with something.

- visit the local demo app @ http://localhost:4300 in Chrome latest (important; otherwise worked on FF)
- right click on the "ICON" or "ONLY" text specifically -- not on whitespace characters on the header for "ICON ONLY BUTTONS"
- Chrome should no longer crash and visit an "Aw Snap" and "Error code: 5" page
image

### DIFF
--- a/src/ids-text/ids-text.js
+++ b/src/ids-text/ids-text.js
@@ -45,19 +45,18 @@ class IdsText extends mix(IdsElement).with(IdsEventsMixin, IdsThemeMixin, IdsToo
    */
   static get attributes() {
     return [
-      attributes.TYPE,
-      attributes.FONT_SIZE,
       attributes.AUDIBLE,
+      attributes.COLOR,
       attributes.DISABLED,
       attributes.DISPLAY,
       attributes.ERROR,
-      attributes.MODE,
-      attributes.VERSION,
-      attributes.LABEL,
+      attributes.FONT_SIZE,
       attributes.FONT_WEIGHT,
-      attributes.AUDIBLE,
+      attributes.LABEL,
+      attributes.MODE,
       attributes.OVERFLOW,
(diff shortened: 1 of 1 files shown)


tmcconechy (commented):
This definitely fixes the issue but the problem with it is that now you cant update the text as it wont reflect back in the shadow DOM.

Maybe we can use the slotchange event to solve that? Given that i'd call this a temporary fix.

rob2d:
@tmcconechy oh good point -- was having a bit of tunnel vision with the crashing there.

The thing on the `slotchange` is it needs a `slot` to fire. We can also use a MutationObserver but I think for both `slotchange` or `mutationobserver` it would be the same overhead.... might not be ideal to have for text nodes since they're displayed so much so I guess we'll have to come up with something else.

tmcconechy:
Yeah lets leave this open for discussion tho. I think that might be too heavy for a text item and this feels like a chrome bug we can get around by open the dev tools first and clicking the pointer. So maybe not worth making the text non updatable

rob2d:
Closing PR just because I exhausted a few other options this AM and seems outside of a `MutationObserver` that would may add some overhead as a fix currently, we really just need Chrome to add a patch at some point. Will create an issue with some notes on the `e.preventDefault` workaround and the tracker for ref.

PR #263, opened 2021-07-19
224- fix dirty-tracker indicator

### LINKED ISSUE #224: TrackDirtyMixin: Shows the dirty tick on initial load
**Describe the bug**
If the input has an initial value set in the DOM it shows the tracker indicator.

**To Reproduce**
1. Edit https://github.com/infor-design/enterprise-wc/blob/main/app/ids-input/example.html#L20 and add value="test"
2. Go to http://localhost:4300/ids-input
3. See error (dirty indicator should not show)

**Expected behavior**
If the value is set in the DOM initially should not show the dirty indicator

**Additional context**
Should fix an issue in color picker of the same.

vivz753:

- I applied `.toLowerCase()` to all hex color codes--this is why the dirty-tracker indicator is not working correctly for `ids-color-picker` component, because it was comparing a lower-cased value to an all-caps value (it is initially entered as all caps in the attribute, but gets lower-cased in the process of being set). 
- I was not able to reproduce the issue saying that initial values also triggered the dirty-tracker indicator on the `ids-input` component. I believe the issue was just due to inconsistent casing for the `ids-color picker`, so this should solve the ticket.

- Closes #224

- `npm run start:watch`

- change the color hex code by typing it in or clicking on the color swatch
- the dirty-tracking indicator (yellow triangular flag in top left) should only appear when values differ from the initial value from when page first loaded

Initial value here is `#b94e4e`
Screen Shot 2021-07-19 at 4 17 13 PM
Screen Shot 2021-07-19 at 4 17 32 PM

- change line 20 in `app/ids-input/example.html` to `<ids-input value="test" type="text" label="Dirty Tracking" dirty-tracker="true" placeholder="Dirty Tracking"></ids-input>`
- the dirty-tracking indicator already works as it should (dirty-tracking indicator only appears when value differs from initial value "test")

Screen Shot 2021-07-19 at 4 18 31 PM
Screen Shot 2021-07-19 at 4 18 37 PM

### DIFF
--- a/app/ids-input/example.html
+++ b/app/ids-input/example.html
@@ -17,7 +17,7 @@
     <ids-input type="text" label="Email Address" placeholder="Company@address.com" validate="email required"></ids-input>
     <ids-input type="text" label="Disabled" disabled="true" value="Field not editable"></ids-input>
     <ids-input type="text" label="Readonly" readonly="true" value="02006"></ids-input>
-    <ids-input type="text" label="Dirty Tracking" dirty-tracker="true" placeholder="Dirty Tracking"></ids-input>
+    <ids-input type="text" label="Dirty Tracking" dirty-tracker="true" placeholder="Dirty Tracking" value="02006"></ids-input>
     <ids-input type="text" label="Clearable" placeholder="Clearable text Field" clearable="true"></ids-input>
     <ids-input type="text" label="Autoselect" value="Text select on focus" autoselect="true"></ids-input>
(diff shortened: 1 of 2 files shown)


tmcconechy (approved):
This seems to fix the issue. @danielortiz1982 can you review?
I pushed an example to http://localhost:4300/ids-input to test 
So will have to approve the images changes when it builds

EdwardCoyle (approved):
Looks good 👍

PR #270, opened 2021-07-22
245 - Added disabled font color without themes

### LINKED ISSUE #245: IdsBreadCrumb: Relies on themes and slotted for color
**Describe the bug**
When the theme is not set the color is not working for disabled.

**To Reproduce**
1. Go to http://localhost:4300/ids-breadcrumb
2. Notice the color is ok for disabled
3. Go to http://localhost:4300/ids-breadcrumb/side-by-side
4. Notice the color is not ok for disabled

**Expected behavior**
Since the theme is not set on this page it should set the color for lighht them by default and not use slotted for the styles

Jah25:
Updated the styles for bread crumbs to add a styling case when there is no mode. I used ::slotted styles to avoid making changes to the underlying hyperlink styles which are being intentionally overridden with color=unset.

Added slotting styles in ids-bread crumb to deal with the case where no theme is selected. this prevents disabled fields from being the same color as normal bread crumbs. I made this change on the ids-bread-crumb to avoid putting breadcrumb specific changes in ids-hyperlink especially since the base themeless styling for the hyperlinks is being overridden with color="unset".

Closes #245

-build and run the application
-navigate to http://localhost:4300/ids-breadcrumb
-open another tab and navigate to http://localhost:4300/ids-breadcrumb/side-by-side
-the disabled option should be the same color on both pages, see below

Screen Shot 2021-07-22 at 3 36 42 PM


Jah25:
wasn't sure if I should update the change log here in the same way we do it for enterprise

tmcconechy:
@Jah25 regarding the image - i think that is a timing issue i approved it on another PR and it should be ok now that i pulled in main

tmcconechy:
@Jah25 I approved the changes i think there are a few things going on but not your fault

- header was merged in (header changes out of sync)
- message changes might be an intermittant error to watch
- link changes made small change to the counts which uses links

Will keep an eye on these but everything looks fine to me

tmcconechy (dismissed):

tmcconechy (approved):

EdwardCoyle (approved):
@Jah25 looks good to me 👍

PR #271, opened 2021-07-23
ids-element: memo style objects for performance improvement per-render

rob2d:

Currently on IDSElement, we are recreating styles on every render via `this.appendStyles`. There are only about 53 different namespaces for components registered and they are not using the same resources even though they are the same components that share the same styles so could spare to cut it down a bit with a simple memoization obj.

This is not the most perfect/thorough optimization since constructed stylesheets currently include a lot of content per-component and we go through setAttribute reflecting as well which takes a bulk of first render and there's other things after doing perf tests tonight, but very low hanging fruit on this one.

Found that on the demo page with changes, 
- 95.7% of renders could avoid creating a stylesheet/adopted stylesheet instance were avoided
- average for each component of just the demo page styles 1x went from ~0.573ms to ~0.033ms or 19-20x for 1st initial render/page-load of the components with no interactions (scale would gradually increase in favor of optimization as well). 

These numbers aren't crazy in scope (saves about ~0.5s on our demo page render on my PC in that test) but gives room for things to feel smoother per-render on average since there's only 16.6ms to work with for 60fps and time to first paint is one of the most important metrics on SEO.

(Note: haven't ran tests locally just pushing so PR can be up here for tomorrow AM in case coverage is not there now in ids-element)

- Related to #140 which is not confined to this issue and this was known for a while, just finally getting to this tonight since was browsing ids-element earlier and remembered to do that.

(1) On our demo app at http://localhost:4300, if you pull branch `component-stylesheets-optimization-perf-a` it saves cache hits/misses in `window.cacheHits`/`cacheMisses` and if logging in console after app loads (before there is interaction; as interactions will just increase cache hits thereafter here): 

```
window.cacheHits => 1204
window.cacheMisses => 54 ​
% of cache hits => ~95.7%
```

(2) for the performance check, check out the spreadsheet at:
https://docs.google.com/spreadsheets/d/1zbB23xci4Eh-qIjhLhR_FrD41-rl_mIsWIobHZyJaNs/edit?usp=sharing

Feel free to try to reproduce in the spreadsheet by first copying and then:

- pulling `component-stylesheets-pre-optimization-log` and copying console, stripping with find/replace in text editor and pasting into column `A` of `main`. Then notice the value on `B:1`.

- pulling `component-stylesheets-optimization-perf-b` and copying `console.time` logs; stripping with find/replace in text editor and pasting into column `A` of `optimized-branch-${your-browser}`. 
- Notice the value on `B:1` is much lower (this represents the average of all values in A which are per-component `appendStyles` runtime in ms).
- Component styles should render fine and tests and coverage should pass

### DIFF
--- a/src/ids-base/ids-element.js
+++ b/src/ids-base/ids-element.js
@@ -11,13 +11,6 @@ import renderLoop from '../ids-render-loop/ids-render-loop-global';
 import IdsRenderLoopItem from '../ids-render-loop/ids-render-loop-item';
 import { stringUtils } from './ids-string-utils';
 
-const areAdoptedStyleSheetsSupported = (
-  window.ShadowRoot
-  && (window.ShadyCSS === undefined || window.ShadyCSS.nativeShadow)
-  && 'adoptedStyleSheets' in Document.prototype
-  && 'replace' in CSSStyleSheet.prototype
-);
-
 /**
  * simple dictionary used to memoize attribute names
  * to their corresponding property names.
@@ -52,9 +45,14 @@ const memodComponentStyles = {};
  *
  * @param {string} namespace namespace of component e.g. ids-text
  * @param {string} cssStyles the CSS Styles for the component
+ * @param {boolean} areAdoptedStyleSheetsSupported whether or not adopted stylesheets are supported
  * @returns {HTMLElement} style tag object
  */
-function getMemodComponentStyle(namespace, cssStyles) {
+function getMemodComponentStyle(
+  namespace,
+  cssStyles,
+  areAdoptedStyleSheetsSupported
+) {
   // if cache doesn't exist, create the stylesheet depending on
   // whether adopted stylesheets are supported or not by
   // the browser
@@ -223,9 +221,7 @@ class IdsElement extends HTMLElement {
       this.attachShadow({ mode: 'open' });
     }
 
-    if (this.shadowRoot.adoptedStyleSheets) {
-      this.appendStyles();
-    }
+    this.appendStyles();
 
     template.innerHTML = this.template();
     this.shadowRoot?.appendChild(template.content.cloneNode(true));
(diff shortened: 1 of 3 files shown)


rob2d:
Seems there were a few things I didn't account for in some newer components e.g. datagrid relies on :host re-styling so can't have a shared instance of CSSStylesheet currently to alter CSS vars (unfortunately), and our Jest tests restrict what we can do on base without losing coverage and also has some quirks on CSSStyleSheet behavior. 

There's a few approaches to that but may need to close this PR to rethink that for time sake for now.

tmcconechy (dismissed):
ok this might be fine for now but on that performance issue i had planned to try links instead which i found seem to be a lot faster. but this is unconfirm/not completely tested.

So this seems good for now since it helps and doesnt break anything

tmcconechy:
oh except for a few tests https://github.com/infor-design/enterprise-c/pull/271/checks?check_run_id=3139602758#step:9:909

rob2d:
@tmcconechy yeah had mentioned in comment above there were quirks... managed to fix/work around a few of them so far -- it seems actually mostly because of the way Jest polyfills (or doesn't) the adoptedStylesheets and the way it's CSSStyleSheet works. So we worked around it in certain ways but then to do proper checks in our app code for that causes it to fail and read the DOM as if stylesheets are not supported since JSDOM is pretty brittle. 

Was looking into some solutions yesterday to get past the nitty gritty stuff like this without losing coverage in Jest and maybe we can have a `process.env.JEST_WORKER_ID` check in ids-element to emulate shadow DOM as worst case scenario for some of these super base level things?

For the data grid I started looking into a solution last night also to try to put the CSS var relied on somewhere besides in `:host`. These sort of edge cases would still probably be an issue with `<link>` as well. 

For this branch mainly was just seeing how much faster recycling the styles could be, and can still use basically the same approach as link tags. The link tags also should be memo'd since running on render so even though there may be some browser optimization it is probably more expensive for the object/string serializing than the browser to make the styles.

Going to return to this one but just to update on it here.

rob2d:
Closing for now just due to other priorities on the sprint or smaller side tasks outside of this at the moment to look at since it's not a critical issue, but it should be helpful for us to memo our style objects after seeing the practical perf gain (either way whether we're using `link` or `style` attrib, it should be the same). 

Maybe in the meantime on new component PRs in retrospect for reviews it can be good for us to keep an eye out for style via JS on "host" with any vars specifically since these cases throw a small monkey wrench to untangle (not a bad practice in itself, just based on if we want to be able to share resources between similar components with the same theme context).

rob2d:
just to add a thought on this one for whenever myself/anyone else looks at this branch again: just realized it might be possible to add an attribute or some kind of meta to denote the style object as the "adopted stylesheet" internal node to fix the issues I was having. Then vs replace the shadowRoot always keep existing styles there and target the one to replace if possible (ref on whether data attribs are ok @ https://stackoverflow.com/a/48841488/1291659 )

PR #278, opened 2021-07-28
254 - Add Color Variant Mixin

### LINKED ISSUE #254: IdsButton: Ability to set color scheme to white/inverse
**Describe the bug**
To support components that have colored/personalized backgrounds. We need a setting added to ids-button and ids-text to support https://main-enterprise.demo.design.infor.com/components/header/example-index.html (white colors)

**Expected behavior**

- [x] New setting called `inverse` (as in inverse color)
- [x] Style is as per https://main-enterprise.demo.design.infor.com/components/header/example-index.html  with hover, focus, active state changes , this css would be added to ids-button and ids-text
- [x] Test this in ids-header (will need to set this setting in the component so the developer doesnt have to worry about it)

**Additional context**
#236

EdwardCoyle:
This PR adds a mixin that allows for a switch to alternate color variants on some components, providing styles that may combine better with alternate background colors.  For example, the standard Light theme's text color is Slate 100 (almost black), but when placed on an IdsHeader or some other dark background, this color should flip to Slate 10 (almost white).

This PR also changes the way text colors are provided to `ids-text` and `ids-icon` nodes.  The "real" text color has been reset to default (the `currentColor`) in the ids-text component, to match the behavior of the ids-icon component.  The "actual" color of text and icons is now passed down from wrapping/host components.  The base text color is provided by `ids-container` (similar to background color), and can be overridden by child components of the container (IdsHeader, IdsPopupMenu, and IdsButton also control text colors to some degree).  After some testing, I haven't seen any issues with this method clashing with shadow root encapsulation, but we should thoroughly check all components to make sure.

Included in this PR:
- Added code/types/tests/docs/examples for the Color Variant Mixin
- Moved base text colors out of the `ids-text` component.
- Established base theme colors and "alternate" colors on `ids-container`
- Fleshed out all theme styles and added "alternate" styles for all types of default, primary, and secondary IdsButtons.
- Fixed some button visual bugs
- Added animation to background elements (header/container) to match the change on buttons.

Closes #254 

Pull, build, run, then check the following:

_Check basic "alternate" color variant styles:_
Look at these pages (which are inverse by default), and make sure the styling makes sense.  Try all modes/versions:
- http://localhost:4300/ids-text/variant-alternate
- http://localhost:4300/ids-icon/variant-alternate
- http://localhost:4300/ids-button/variant-alternate

_Check "alternate" color variant implementations:_
Try all button states on IdsHeader and ensure they are usable/make sense:
- http://localhost:4300/ids-header (all buttons here are "alternate" by default)
- http://localhost:4300/ids-header/button-types (use the disabled checkbox and check disabled states here as well)
- http://localhost:4300/ids-header/standalone-css (should now have correct icon/text colors)


EdwardCoyle:
@danielortiz1982 can you double-check the changes to the Header and make sure everything's working as expected?

EdwardCoyle:
Marking WIP for now because some of the text color changes are affecting lots of Percy tests.  A lot of the changes are legitimate but some aren't, just need to sort through.

rob2d:
Overall a great idea/addition that we definitely need this functionality with but for the convention there's something I just wanted to bring up for consideration if possible: 

it seems like "invert" as a term may be a little ambiguous here (e.g. digital graphics or color wheels with blue => yellow) and could be more genericized.  

Reasoning: often in dark/light/high contrast theme for "material" themes, the operation that happens may be different on some components than just inverting it. For example sometimes there is a canvas that becomes opaque or suddenly appears due to high contrast. 

So not sure of some way to phrase "color-variant"? (or insert any more generally-applicable term for re-sub-theming a component) and this is just mainly a subjective suggestion. And maybe instead of a switch having a sub-variant of the components just to be adaptable without having to add edge cases later or custom attributes for styles?

I may be talking completely out of context here also; if this is something really consistent in the app since the project has existed or seems it won't be an issue then please just disregard this. Also checked the code and the way it is happening e.g. somewhat like application on theme looks like a solid approach. The suggestion is just with concerns of more general flexibility in case any changes on design side (which we may not have control on) suddenly change for any reason and requires monkey patching.

EdwardCoyle:
Thanks for the feedback!  I actually like the idea of adding the possibility for more variants.  Also, I hadn't fully considered this before:

> the operation that happens may be different on some components than just inverting it. For example sometimes there is a canvas that becomes transparent or suddenly appears due to high contrast, or there may be a subtle color variant in one area that is common in the app but not elsewhere 

We had become accustomed to calling it "inverse" in 4.x because it was generally opposite coloration (white text on black/dark slate background, for example), but there were definitely cases that weren't totally "opposite".  This is already present in some of these 5.x color schemes in this PR (like either Dark Mode not having a fully black background).

If we go the multi-variant route, we'd just have to think about how we eventually categorize all the current 4.x patterns that we'd previously call "inverse".  We can do that as we add them to 5.x.  It's a good opportunity to define those a bit more.

EdwardCoyle:
Re: The name of the enum property -- Maybe some form of `variant` (`color-variant` is a possibility), or we have referred to them as `alternate` in some places in 4.x.  @tmcconechy any preference?

tmcconechy:
I guess i like color-variant that seems pretty clear? like color-variant="white"? But also ok with inverse or alternate. I dont care very strongly

EdwardCoyle:
We could go with `color-variant="header"`, or maybe `aside` for the styles present in this PR, thinking that they'll mostly be applied to things like Headers or sidebar content like App Menus.  Any objections?

tmcconechy:
I dont follow "aside" as a name. For color-variant i think we would use the white variant on places other than the header tho.

rob2d:
@EdwardCoyle I don't have any other an idea on the name but between the two I would also vote on "alternate" and especially if it is one that occurs outside of the header or most frequent in the app like @tmcconechy mentioned.

EdwardCoyle:
@tmcconechy @rob2d I went through and changed the mixin to `IdsColorVariantMixin`, renamed the attribute to `color-variant`, and changed to enum behavior.  Added some tests around this as well.  I think it's good to go now.

EdwardCoyle:
NOTE: I went through all the flagged Percy tests, and I fixed the ones that weren't right.  The ones that are left are all slightly off due to the container's new text color applied to the "New Light" version/mode, which is actually matching what we have in 4.x now (before most things seemed to be using black text).

app/ids-header/color.html
```diff
@@ -2,20 +2,20 @@
 <ids-header class="ids-header-example" color="#bb5500">
   <ids-toolbar>
     <ids-toolbar-section>
-      <ids-button icon="menu" role="button">
+      <ids-button icon="menu" role="button" color-variant="alternate">
```
tmcconechy, line 5: Do we want to add this within the header component or do you think its better to let the user do it?

tmcconechy (commented):
Looks pretty great to me. I just had one question / discussion point i added. Also im wondering if some of the text changes were accidental as i see the text color changed in a few places. I approved the ones that seemed ok and left the rest that im wondering about.

https://percy.io/00aaa841/enterprise-wc/builds/11855143?utm_campaign=00aaa841&utm_content=enterprise-wc&utm_source=github_status_public

There is a couple in there where contrast seems to regress?

app/ids-header/color.html
EdwardCoyle: I think we ought to leave it "manual" for the time being and see if it becomes an issue.... unless you think we can guarantee that every header someone implements will always have the alternate styling?

EdwardCoyle:
@tmcconechy In a lot of cases on the light theme, the font color was previously black because it was inheriting from the browser styles since there was no color set in any of the components. On `IdsContainer` there's now a base font color of Slate 100 on the light theme, which is what's in 4.x.

I did a quick check of the pages in question and the only ones I can see having any significant color contrast issues (that aren't disabled items) are on http://localhost:4300/ids-notification-banner/.  aXe is specifically flagging the text nodes inside the Card (not the notifications, but the actual list item nodes)... which doesn't make sense to me because its Slate 100 text over a white background...

Screen Shot 2021-08-02 at 10 20 48 AM

tmcconechy (approved):

danielortiz1982 (approved):

danielortiz1982:
@EdwardCoyle Everything looks good to me. I do not see anything of a concern here. Great work on this! 👍

tmcconechy:
@rob2d can you take a last look as @EdwardCoyle addresses comments based on your suggestions

deep7102 (approved):
🍌 Looks great!

PR #291, opened 2021-08-04
webpack: share runtime for normalized modules & update the source-map-loader

### LINKED ISSUE #290: Chore: Improve Bundling to reduce consistency
**Describe the bug**
Something (may) be wrong with our bundling but not sure. 

**To Reproduce**
1. Go to ids-theme-mixin and add
```
import { attributes } from '../ids-base/ids-attributes';

console.log('In IdsThemeMixin');

/**
 * A mixin that adds theming functionality to components
 * @param {any} superclass Accepts a superclass and creates a new subclass from it
 * @returns {any} The extended object
 */
const IdsThemeMixin = (superclass) => class extends superclass {
  constructor() {
    super();
  }
```
2. Go to http://localhost:4300/ids-button
3. Check console

**Requirements**
- We do have the requirement that each component can be imported by itself with what it needs to run standalone/side by side
- We also need each mixin to have its own shared instance i.e. the events mixin has a map and event 
- should not give CSP errors (some bundlers did use an eval which was flagged)

**Expected behavior**
We do have the requirement that each component can be imported by itself. But is there someway to reduce the redundancy?

**Additional context**
We could use import instead and import from the npm folder like the locale scripts do https://webpack.js.org/api/module-methods/#import-1

Or better perhaps changing to ESM will work https://github.com/infor-design/enterprise-wc/blob/main/webpack.config.js#L55-L57

rob2d:

(1) changes a parameter to share the runtime across app entry-points as the nature of our library is one global set of instances (#290). 

Fixes a few hidden quirks about using `instanceof` or `Ids{Component}.prototype` consistency in mixins/callbacks and saves a tiny bit of RAM/time executing JS on-load.

(2) adds `source-map-loader` to debug some types of problems a bit quicker.

Fixes #290 

- pull this branch
- run `npm i`
- close browser or run in incognito/hard refresh

- open Chrome Dev Tools => Source Tab => Ctrl + P  and search an ids component
image
- should see the compiled source map vs the webpack dist
- breakpoints work (but one caveat: since the code is source-mapped on first load, going another level now does not work oddly. But the source-maps seem to be correct/loaded by default now)

- visit http://localhost:4000/ids-pager
- open Chrome Dev Tools => Sources Tab => Ctrl + P and search for `ids-theme-mixin`
- put a breakpoint by clicking line number 8 and refresh page
- theme mixin module should run the line declaring 1x instead of repeat for each/every component in the app
(does call super class so may need to add parens as that's how I was testing to be sure; alternate method to debug is just old fashioned console.log above the `IdsThemeMixin` declaration)
image

N/A

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16679,6 +16679,28 @@
       "integrity": "sha512-/3GptzWzu0+0MBQFrDKzw/DvvMTUORvgY6k6jd/VS6iCR4RDTKWH6v6WPwQoUO8667uQEf9Oe38DxAYWY5F/Ug==",
       "dev": true
     },
+    "source-map-loader": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/source-map-loader/-/source-map-loader-3.0.0.tgz",
+      "integrity": "sha512-GKGWqWvYr04M7tn8dryIWvb0s8YM41z82iQv01yBtIylgxax0CwvSy6gc2Y02iuXwEfGWRlMicH0nvms9UZphw==",
+      "dev": true,
+      "requires": {
+        "abab": "^2.0.5",
+        "iconv-lite": "^0.6.2",
+        "source-map-js": "^0.6.2"
+      },
+      "dependencies": {
+        "iconv-lite": {
+          "version": "0.6.3",
+          "resolved": "https://registry.npmjs.org/iconv-lite/-/iconv-lite-0.6.3.tgz",
+          "integrity": "sha512-4fCk79wshMdzMp2rH06qWrJE4iolqLhCUH+OiuIgU++RB0+94NlDL81atO7GX55uUKueo0txHNtvEyI6D7WdMw==",
+          "dev": true,
+          "requires": {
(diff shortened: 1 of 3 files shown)


danielortiz1982 (approved):
@rob2d This is a great solution for this issue, good find adding the **source-map-loader** plugin and changing the runtimeChunk to a single app. I think there could be some small improvements on the code splitting strategy but for now I think this will work fine.

tmcconechy (changes requested):
1) Great find on the single part ❤️ 
2) But have some questions about the source-loader addition.

- i read up on it but i dont get what it does?
- we were using terser for the source maps already...so how does it ensure that works properly?
- worse is i can no longer debug the way im used to (or at all?) see attached video

https://drive.google.com/file/d/1zgsg9ZHqSWN-h_OeOymZPHBHxZqrUekn/view

tmcconechy (approved):
After clarifying what this added with @rob2d i misunderstood part of this. So seems good 👍🏻

EdwardCoyle (approved):
Really nice to be able to debug the compiled page files, thanks for this @rob2d!

PR #297, opened 2021-08-05
129 - Add ids toast

### LINKED ISSUE #129: IdsToast: Add new web component
**Describe the bug**

Bring over the current toast component.  See if we can simplify and improve the code at all while improving as we go. Current toast component https://master-enterprise.demo.design.infor.com/components/toast

```
<ids-toast-container draggable="true" position="top-left">
 <ids-toast title="This is a toast">
    Toast message blah blah
 </ids-toast>
</ids-toast-container>
```

- Supports multiple toasts
- Supports drag (mobile can be for later)
- Supports positions
- Supports links
- Supports accessibility 

**Additional context**
- 100% test coverage
- DOCS, CHANGE log ect. 
- themes 

deep7102:
Added IDS-Toast component

Closes #129

- Navigate to:
- http://localhost:4300/ids-toast
- http://localhost:4300/ids-toast/sandbox
- http://localhost:4300/ids-toast/side-by-side
- Use each button to see related toast
- See the docs and tests


src/ids-toast/README.MD
```diff
@@ -0,0 +1,414 @@
+# Ids Toast Component
+
+## Description
+
+The IDS Toast component provides information, warning small and temporary messages. The Toast will close itself after a timeout (default 6s) or clicked on close button. It is also possible to have more than one toast onscreen at a time.
```
tmcconechy, line 5:
```suggestion
Toasts are used to display confirmations of success, failure, or other statuses of system processes related to a user’s workflow. If a user submits a form, a toast will confirm the successful or unsuccessful completion of that submission. Toasts are dismissible via the “close” icon or the escape key. If not dismissed, the toast will hide after a configurable amount of time. The Toast will close itself after a timeout (default 6s) or if the close button is clicked. It is also possible to have more than one toast onscreen at a time.
```

src/ids-toast/README.MD
```diff
@@ -0,0 +1,414 @@
...
+```
+
+```javascript
+const btnToast = document.querySelector('#btn-toast');
+
+btnToast?.addEventListener('click', () => {
+  const toastId = 'demo-toast';
+  let toast = document.querySelector(`#${toastId}`);
+  if (!toast) {
+    toast = document.createElement('ids-toast');
```
tmcconechy, line 36: Not 100% sure what we can do but wonder if there is a way to have less setup for a toast message. Did you compare with message? Are they similar?

src/ids-toast/README.MD
```diff
@@ -0,0 +1,414 @@
...
+- `audible` {boolean} Set as invisible on the screen, but still read out loud by screen readers
+- `destroyOnComplete` {boolean} Set to destroy after complete all the toast messages
+- `draggable` {boolean} Set user to allows drag/drop the toast container
+- `position` {string} Set position of the toast container in specific place
+- `progressBar` {boolean} Set toast to have a visible progress bar
+- `savePosition` {boolean} Set toast container to save position to local storage, apply only when draggable set to true
+- `timeout` {number} Set the amount of time, the toast should be present on-screen
+- `uniqueId` {number} Set uniqueId to save to local storage, so same saved position can be use for whole app, apply only when draggable set to true
+
+## Settings and Attributes (Toast Message)
```
tmcconechy, line 366: We should add sections for the events and public methods too

src/ids-toast/ids-toast.js
```diff
@@ -0,0 +1,615 @@
...
+} from '../ids-mixins';
+
+import { IdsToastShared as shared } from './ids-toast-shared';
+import { IdsToastMessage } from './ids-toast-message';
+import styles from './ids-toast.scss';
+
+import { IdsDraggable } from '../ids-draggable';
+
+// Attributes observed by the Toast
+const TOAST_ATTRIBUTES = [
```
tmcconechy, line 26: We decided its better to put these down below.

src/ids-toast/ids-toast.js
```diff
@@ -0,0 +1,615 @@
...
+    });
+    this.#removeFromDom();
+  }
+
+  /**
+   * Get message element by given message id.
+   * @param {string} messageId A message id to use.
+   * @returns {HTMLElement|undefined} The message element
+   */
+  messageElem(messageId) {
```
tmcconechy, line 170: I think some of these are private? So should use #128

tmcconechy:
This looks really great.

- small comments
- how do i use the save position? I cant verify its saved anything
- the X looks really big can we try the smaller one?

Screen Shot 2021-08-06 at 11 52 47 AM

Also working on the visual image, the first thing didnt work.

tmcconechy (commented):
See comments

src/ids-toast/README.MD
deep7102: This `toastId` and `messageId` are optional, if they use it will be easy to target the element do things like watch events or etc..

src/ids-toast/ids-toast.js
deep7102: this `messageElem(messageId)` need as public to get the message element reference to provide by message id

deep7102:
Hi @tmcconechy updated now,
all the points you saw should be working now: http://localhost:4300/ids-toast/sandbox

tmcconechy (approved):
Good job on this one @deep7102 looks good to me

src/ids-base/ids-string-utils.js
```diff
@@ -79,6 +79,24 @@ export function buildClassAttrib(...classes) {
   return !classAttrib ? '' : ` class="${classAttrib}"`;
 }
 
+/**
+ * Check whether the given object is an object or not.
+ * @param {any} obj The object to check.
+ * @returns {boolean} true if given object is an object.
+ */
```
rob2d, line 86: just a note on this but it seems: maybe this should be in a standalone utility? or at least another class of them (since we input any and it's just checking for flag)

src/ids-base/ids-string-utils.js
```diff
@@ -79,6 +79,24 @@ export function buildClassAttrib(...classes) {
...
+export function isObject(obj) {
+  return obj instanceof Object && !(obj instanceof Number) && !(obj instanceof Array);
+}
+
+/**
+ * Check the given object is an object and NOT empty.
+ * @param {any} obj The object to check.
+ * @returns {boolean} true if given object is an object and NOT empty.
+ */
```
rob2d, line 95: same on this as isObject

src/ids-base/ids-xss-utils.js
```diff
@@ -77,17 +77,15 @@ export function stripTags(html, allowed) {
 
   const allowList = ((`${allowed || ''}`)
     .toLowerCase()
-    .match(/<[a-z][a-z0-9]*>/g) || [])
-    .join(''); // making sure the allowed arg is a string containing only tags in lowercase (<a><b><c>)
+    .match(/<[a-z][a-z0-9|ids\-a-z]*>/g) || [])
```
rob2d, line 80: just checking but: is the "|" an intentional character here? 

In regex `[]` would eval to a group so just making sure.

Seems also though maybe we should also allow uppercase e.g. `a-zA-Z`?

src/ids-toast/ids-toast-message.js
```diff
@@ -0,0 +1,312 @@
...
+} from '../ids-base';
+
+// Import Mixins
+import {
+  IdsEventsMixin,
+  IdsKeyboardMixin,
+  IdsThemeMixin
+} from '../ids-mixins';
+
+import { IdsToastShared as shared } from './ids-toast-shared';
```
rob2d, line 16: nit: 

(1) it seems like if it's a shared namespace/global, it should keep the `IdsToastShared` convention within the component? just muddles things a bit if it's not an instance of something here but used throughout.

(2) there isn't a noun on this one which makes it feel a little off... maybe IdsToastConstants?

src/ids-toast/ids-toast-message.js
```diff
@@ -0,0 +1,312 @@
...
+   */
+  set progressBar(value) {
+    if (shared.isBool(value)) {
+      this.setAttribute(attributes.PROGRESS_BAR, value.toString());
+    } else {
+      this.removeAttribute(attributes.PROGRESS_BAR);
+    }
+  }
+
+  get progressBar() { return shared.getBoolVal(this, attributes.PROGRESS_BAR); }
```
rob2d, line 278: nit: seems can use common (attrib)stringUtils? (since has == true, doesn't have === "false"|"true")

src/ids-toast/ids-toast.d.ts
```diff
@@ -0,0 +1,128 @@
...
+}
+
+interface IdsToastRmContainerEventDetail extends Event {
+  detail: {
+    elem: IdsToast,
+    uniqueId: string
+  }
+}
+
+type IdsToastShowTypes = {
```
rob2d, line 45: nit: IdsToastShownTypes (?)

src/ids-toast/ids-toast-message.js
```diff
@@ -0,0 +1,312 @@
...
+      this.container?.classList.add(shared.TOAST_MESSAGE_CLASSES.start);
+    }
+
+    /* istanbul ignore next */
+    if (!this.progressBar && progressBar) {
+      progressBar.parentNode?.removeChild(progressBar);
+      progressBar = undefined;
+    }
+
+    const duration = this.audible ? 100 : this.timeout;
```
rob2d, line 129: nit: seems if we have a namespace dedicated to IdsToast's constants, this can be there maybe? (as centralized place for ref)

src/ids-toast/ids-toast-message.js
```diff
@@ -0,0 +1,312 @@
...
+    if (value) {
+      this.setAttribute(attributes.TIMEOUT, value.toString());
+    } else {
+      this.removeAttribute(attributes.TIMEOUT);
+    }
+  }
+
+  get timeout() {
+    const timeout = parseInt(this.getAttribute(attributes.TIMEOUT), 10);
+    return !isNaN(timeout) ? timeout : shared.DEFAULTS.timeout; // eslint-disable-line
```
rob2d, line 294: (ok seeing there is a default so maybe comment above setting to 100 was just a missed case)

.eslintrc.js
```diff
@@ -95,6 +95,7 @@ module.exports = {
     // Allow console.info
     'no-console': ['error', { allow: ['error', 'info'] }],
     'template-curly-spacing': ['off'],
+    'no-underscore-dangle': ['error', { allow: ['_client'] }],
```
rob2d, line 98: this was one I noticed for a few things... I think maybe we should allow it generally but just not make it a "common practice"?

Reason: there's other cases in ES7 naturally where they also make sense that I'm not sure can be allowed in this rule where we'd need currently 2 rules ignored there, for example:
```js
const [_match, group1, group2] = regexMatch; 
```

Or:

```js
const { _property1, _property2, ...otherProperties } = obj;

for(const property of otherProperties) {
  doAnythingWithoutNeedOfSwitchOrIf();
}
```

(in these cases then it's clear _property1 and _property2 aren't used because of the _; vs just awkward hanging vars)

rob2d (commented):
@deep7102 I just had some some minor nit:s I put comments on but looking good!

src/ids-toast/ids-toast-message.js
rob2d: actually nvmd: I see there is an audible duration vs default duration

.eslintrc.js
tmcconechy: I think we should not allow `no-underscore-dangle` and avoid the dangle unless there is a reason (for example i added _client because thats what the test API has but i can see it's not being used anymore) but we can keep this in in case we have to add any underscore-dangle for any APIS we have no control over. otherwise shuold not be any hanging dangles. I would make your example

```
const [match, group1, group2] = regexMatch; 
```

tmcconechy: I added this for

```
await page._client.send('Animation.setPlaybackRate', { playbackRate: 3 });
```

src/ids-base/ids-xss-utils.js
deep7102: @rob2d 
yes, need these `|` and `[]` here and allow uppercase, think we should add when we need them.

src/ids-toast/ids-toast-message.js
deep7102: @rob2d 
great approach, but here can not use it need null=true, true=true and false=false

tmcconechy (approved):

src/ids-toast/ids-toast-message.js
rob2d: oh interesting, thanks on clarifying 👍

src/ids-base/ids-xss-utils.js
rob2d: makes sense 👍

deep7102:
@tmcconechy @rob2d ready for review now.

tmcconechy (dismissed):

rob2d:
looks great, thanks Deepak!

rob2d (approved):

PR #300, opened 2021-08-06
276 add summary field

### LINKED ISSUE #276: IdsSummaryField:  Add new web component

Add an ids-summary-form like the summary form section on https://main-enterprise.demo.design.infor.com/components/form/example-forms.html

The new name should be` IdsSummaryForm`. Will use a slot for extra flexibility.

**Phase 1 Features (MVP)**

- Themes / RTL
- Margin / Padding (like inputs and like before)
- Set the colors and font size unless overridden in the slot to the old values
- build the same example as 4.x (just the summary field part)

**Possible Internal Markup**
```js
  <ids-summary-field>
    <ids-text slot="label" color="red">Label</ids-text>
    <ids-text slot="data" color="red">Label</ids-text>
  </ids-summary-field>
```

Also could be...this but having the slots allows text overrides.
```
<ids-summary-field data="label" value="value"></ids-summary-field>
```

**Suggested Dependencies**
- ids-text

**Settings**
N/A

**Events**
N/A

**Other**

vivz753:

- Adds the Summary Form Web Component 
- Based on 4.x Summary Fields here https://main-enterprise.demo.design.infor.com/components/form/example-forms.html

- Closes #276 

- npm run start:watch
- http://localhost:4300/ids-summary-field

Screen Shot 2021-08-06 at 2 01 58 PM


src/ids-summary-form/ids-summary-form.js
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * Inner template contents
+   * @returns {string} The template
+   */
+  template() {
+    return `
+      <div class="ids-summary-form">
+        <ids-text class="label">${this.label ?? ''}</ids-text>
+        <ids-text class="data" font-weight="${this.fontWeight ?? 'bold'}">${this.data ?? ''}</ids-text>
```
vivz753, line 55: Null coalescing operator calls the get() one less than doing `this.label ? this.label : ''`

Read more here: https://medium.com/quick-programming/3-javascript-features-you-might-not-know-de8ef797b6e4

vivz753:
I don't think I have the ability to approve Percy tests, which I'm assuming is done with the green thumb icon in the top right (it's disabled for me)

Screen Shot 2021-08-06 at 2 26 54 PM

app/ids-progress-chart/example.html
```diff
@@ -6,7 +6,7 @@
   <ids-layout-grid-cell>
     <ids-progress-chart color="#606066" label-progress="90%" progress="90" label="dark"></ids-progress-chart>
     <ids-progress-chart label="error" color="error" label-progress="33%" progress="33">
-      <ids-icon slot="icon" icon="error" size="small"></ids-icon>
+      <ids-icon slot="icon" icon="error" size="medium"></ids-icon>
```
vivz753, line 9: Forgot to change this in the example for `ids-progress-chart` PR @tmcconechy

src/ids-mixins/README.md
```diff
@@ -80,18 +80,18 @@ This mixin add functionality for validation to the component. This includes a ad
...
 ```
 
-1. Make sure if you use connectedCallback that you have a `super.connectedCallback()` in the method
-1. Add types for MODE and VERSION to the `d.ts` file for the new attributes.
-1. Add the theme mixin name to the @mixes tag for future docs.
-1. Add the color changes for each theme scss file. For example:
+3. Make sure if you use connectedCallback that you have a `super.connectedCallback()` in the method
+4. Add types for MODE and VERSION to the `d.ts` file for the new attributes.
+5. Add the theme mixin name to the @mixes tag for future docs.
+6. Add the color changes for each theme scss file. For example:
```
vivz753, line 94: @EdwardCoyle noticed this when we were discussing the color variant, please confirm

app/ids-summary-form/example.html
```diff
@@ -0,0 +1,122 @@
...
+</ids-layout-grid>
+
+<ids-layout-grid>
+  <ids-summary-form label="Shipping to" data="4209 Industrial Avenue<br/>Los Angeles, California 90001 USA">
+  </ids-summary-form>
+  <ids-summary-form label="Shipping Method" data="Freight">
+  </ids-summary-form>
+  <ids-summary-form label="Estimated Delivery" data="June 21, 2015 (4 days)">
+  </ids-summary-form>
+  <ids-summary-form label="Synopsis of Game of Thrones pilot episode" data="
```
tmcconechy, line 12: Might be a little much text going on for the example. While an interesting plot i think some of the topics could be a little sensitive to some sensitive people. We generally stick to something like "office ipsum" or "corporate ipsum" and sometimes "hipster ipsum".

So maybe use one of these and make it a bit shorter just enough to show the line wraps

src/ids-mixins/README.md
tmcconechy: If that works because of the spaces thats good? but normally you can do this:

1. item one
1. item two

It has an advantage in that you dont need to renumber them all if you add one

src/ids-progress-chart/ids-progress-chart.js
```diff
@@ -63,11 +63,11 @@ class IdsProgressChart extends mix(IdsElement).with(IdsEventsMixin, IdsThemeMixi
     return `
     <div class="ids-progress-chart" part="chart">
       <div class="labels">
-        <ids-text class="label-main">${this.label ? this.label : ''}</ids-text>
+        <ids-text class="label-main">${this.label ?? ''}</ids-text>
         <slot name="icon"></slot>
-        <ids-text class="label-progress">${this.progressLabel ? this.progressLabel : ''}</ids-text>
```
tmcconechy, line 68: Just to make sure these changes on ids-progress-chart are intentional?

src/ids-summary-form/ids-summary-form.js
```diff
@@ -0,0 +1,125 @@
...
+  customElement,
+  scss,
+  mix,
+  attributes
+} from '../ids-base';
+
+// Import Mixins
+import {
+  IdsEventsMixin,
+  IdsThemeMixin
```
tmcconechy, line 12: Is theme mixin actually doing anything? or is it just getting the properties (i think ids-text handles the colors fo this so it might not be needed)

tmcconechy (changes requested):
Looks really good just a few small things

src/ids-progress-chart/ids-progress-chart.js
vivz753: Yeah this is just to replace the ternary with null coalescing operator

src/ids-mixins/README.md
vivz753: Ohh, I see. I didn't know that! In that case I'll just leave it

src/ids-summary-form/ids-summary-form.js
vivz753: Did you mean to say IdsEventsMixin? Because I'm not using any events... However, if I remove IdsThemeMixin, the component does not render at all

tmcconechy: No i did mean it might not need IdsThemeMixin as the ids-text elements do the themes right. What does it do when you remove? (Can ping me and chat)

But your right i assume IdsEventsMixin isnt needed either since there are no events here. I missed that, good catch

app/ids-summary-form/example.html
vivz753: Oops, sorry! I didn't even read the second half of it. Changed.

tmcconechy:
Two more issues i found during testing

1) i dont think we need the "not bold" setting. I dont think this is something that will be used and its maybe confusing to have the option
2) i made a mistake on the original ticket this component is call SummaryField not SummaryForm

tmcconechy (commented):
Looks fantastic. Only one more thing. I think the bottom section where we have a non bold text is a bit confusing we dont really need that case. So i would just cut the example off here:

Removing the bottom bit with random data and the non bold text

Screen Shot 2021-08-10 at 10 02 20 AM

vivz753:
There is no SummaryField optioin in `doc/CHECKLIST.md`, should I add it there?

app/ids-summary-field/example.html
```diff
@@ -0,0 +1,17 @@
...
+</ids-layout-grid>
+
+<ids-layout-grid>
+  <ids-summary-field label="Shipping to" data="4209 Industrial Avenue<br/>Los Angeles, California 90001 USA">
+  </ids-summary-field>
+  <ids-summary-field label="Shipping Method" data="Freight">
+  </ids-summary-field>
+  <ids-summary-field label="Estimated Delivery" data="June 21, 2015 (4 days)">
+  </ids-summary-field>
+  <ids-summary-field label="This is some random label" data="
```
tmcconechy, line 12:
```suggestion
  <ids-summary-field label="Comments" data="
```

tmcconechy (commented):
Sure 👍🏻 and maybe lets label this one better 👍🏻

tmcconechy (dismissed):

tmcconechy (approved):

jmacaluso711 (approved):
Looks good 👍

Issue #304, opened 2021-08-11
IdsSlider: Add new web component


tmcconechy:
Add the slide component MVP https://main-enterprise.demo.design.infor.com/components/slider to the web component repo. 

For MVP we can just have these (some can be descoped so put in importance order)

- single https://main-enterprise.demo.design.infor.com/components/slider/example-index.html
- step (and colors) https://main-enterprise.demo.design.infor.com/components/slider/example-colors-and-ticks.html
- range https://main-enterprise.demo.design.infor.com/components/slider/example-range.html
- vertical (if easy) https://main-enterprise.demo.design.infor.com/components/slider/example-vertical.html

The new name should be IdsSlider (ids-slider). If it helps can try and use the `<input type="range">` element https://www.w3schools.com/howto/howto_js_rangeslider.asp

**Phase 1 Features (MVP)**
- Type Option (single (default), range, step, vertical)
- Colors Option (any palette color like `azure-80` or any status color)
- in-progress option (show in pink in progress color)
- Tooltip
- Themes / RTL
- Decimals 

**Future Features (Not Now)**
- Short
- Vertical (if not easy)

**Possible Internal Markup**
```js
<ids-slider min="0" max="10" value="7" steps></ids-slider>
```

**Suggested Dependencies**
- ids-color (maybe)
- ids-text

**Settings**
- color
- value
- min
- max
- type
- step labels
- tooltip

**Design**
https://main-enterprise.demo.design.infor.com/components/stepchart/example-index.html

**Events**
- change

**Other**
Pass the following checklist ordered in priority (some may be left for later) https://github.com/infor-design/enterprise-wc/blob/main/doc/CHECKLIST.md

Issue #143, opened 2021-04-27
IdsSplitter: Add new web component 


tmcconechy:
**Describe the bug**

Bring over the current splitter component.  See if we can simplify and improve the code at all while improving as we go. Current splitter component https://master-enterprise.demo.design.infor.com/components/splitter

```
Markup to be added
```

**Functionality to bring over**
- Supports both x and y splitter
- Supports nesting in both cases important https://github.com/infor-design/enterprise/issues/2697
- May need a parent container in the pattern
- Keep current designs
- Resizes inner containers to 100% those items are slots

**Additional context**
- 100% test coverage
- DOCS, CHANGE log ect. 
- https://github.com/infor-design/enterprise-wc/blob/master/doc/CHECKLIST.MD
- themes 

See completed components like card, tag, badge, hyperlink as reference

rob2d:
Just to update on the scope of this component after discussing with @tmcconechy and @deep7102 today: a lot of the core functionality seems should be abstractable/delegated to a specific component (`<ids-draggable>`), which would also be needed for some base functionalities in other components such as `ids-toast`. 

`ids-splitter` would then just be a wrapper/container pattern with width-setting/height-setting around that, which lowers the complexity of it quite a bit after that.

Created a ticket for this (#253 ) and working on it now as pre-req for the sake of scope creep/review-ability and getting it in the codebase for usage.

rob2d:
After some iteration now that the first stage on draggable is done, I think this approach on the interface may work:

### Basic example
```html
<ids-splitter axis="x">
  <ids-splitter-pane min-size="30%" max-size="50%">
    some content in a child div
  </ids-splitter-pane>
  <ids-splitter-pane>
    some content in a child div
  </ids-splitter-pane>
</ids-splitter>
```

### Nested content

```html
<ids-splitter axis="x">
  <ids-splitter-pane min-size="30%" max-size="50%" size="25%">
    H1
  </ids-splitter-pane>
  <ids-splitter-pane>
    <ids-splitter axis="y">
      <ids-splitter-pane min-size="20%" size="20%">
        H2/V1 (20% of height for content)
      </ids-splitter-pane>
      <ids-splitter-pane>
        H2/V2
      </ids-splitter-pane>
    </ids-splitter>
  </ids-splitter-pane>
</ids-splitter>
```

[...]

PR #307, opened 2021-08-18
ids-draggable: add relative bounds, fixes/iteration

rob2d:

On `ids-draggable`:
- there were some issues with `ids-draggable` where I hadn't had the `event.detail` populated properly for `ids-drag` event. This is needed/blocker for #304 to have the slider grab the `transformX/Y` when the drag event is occuring to do calculations on the slider effectively (otherwise need to parse in styles which is a bit inefficient). Also added detail to the `ids-dragstart` in case there's some initial offset needed.
 
- added `mouseX`/`mouseY` for `x`/`clientX` and `y`/`clientY` of the mouse drag event to the drag events.

- Pushed some functionality which limits the `max/min-transform-x/y`, which will make the containment logic easier to modify for other purposes later, but also relied on for #143. Pushing up early just so this PR can be iterative since it's related and in case other components might need it.
- added `relativeBounds` attrib/property for the IdsSplitter which goes in effect when `parent-containment` is set. This doesn't have documentation yet but pushed since the logic changes involved for supporting it were closely related to the new min/max transform attribs. Was needed on #143 branch specifically and am seeing if there might be another interface change.

- Just some console logging on the demo page so people can see what they can listen for (hopefully can come up with something more intuitive later; just added for now since it isn't well documented yet).

Related to #143  #304 but does not close either directly.

- pull the branch to local env, and open `localhost:4000/ids-draggable`
- dragging the box labeled "w/limits" should confine the user to a square thats (-20, -80) => (+80, +80) around the initial draggable position.
- logs for drag events should show correct numbers
- tests/coverage should pass
eventdetail


rob2d:
Am seeing errors for percy but: seems to be something timing related we're doing since the only changes on this branch are only directly related to `ids-draggable` 💭

app/ids-draggable/example.js
```diff
@@ -1 +1,23 @@
...
+
+const draggables = [...document.querySelectorAll('ids-draggable')];
+
+// add event listeners to console log the draggables;
+// temporary way to document these
+
+draggables.forEach((d) => {
+  d.addEventListener('ids-dragstart', (e) => {
+    /* eslint-disable-next-line no-console */
+    console.log('ids-dragstart', d, e.detail);
```
tmcconechy, line 11: Can use console.info without an eslint-disable-next-line needd

app/ids-draggable/example.html
```diff
@@ -16,6 +16,24 @@
         <ids-text>axis='x'</ids-text>
       </div>
     </ids-draggable>
+    <ids-draggable
+      max-transform-x="80"
```
tmcconechy, line 20: Only wondering if this is the best name as it isnt "transforming" anything its like a limit right? so maybe max-x min-x or maybe x-limit, y-limit.

Maybe this is fine , just discussing?

tmcconechy (commented):
Just two small comments. Also wondering looking at this if there should be some easy way to make it so you cant drag off the page (the other examples).

But i can see this being useful.

app/ids-draggable/example.html
rob2d: oh so this one is a little confusing and I initially named it max_x/y, etc until I realized: 

The x/y are actually relative to it's "real" (existing) position on the DOM... but the div itself isn't moving here (it's only the translate which it's limiting). 

So the bounds of the real object position in DOM aren't changing -- but something can definitely have a real coordinate `offsetTop`/`Left` which also has to be considered/becomes offset... so this is just being translated for the drag which helps it to be flexible without overlapping over things, but if you inspect while translating you'll see there's a real position on the DOM still.

Something else I considered was `max-translate-x`/`min-translate-x` but currently named it transform because of the style property that was being changed.

tmcconechy (approved):

tmcconechy:
sorry @rob2d can you work through the conflicts when you have a chance due to that big restructure PR. Sorry to do that to ya

rob2d:
@tmcconechy no prob; working on it (better that it happened sooner than later in the long run).

tmcconechy (dismissed):

deep7102 (approved):
🍌 Looks good!
Thanks @rob2d

Issue #301, opened 2021-08-11
Webpack optimization for new demo app


danielortiz1982:
**Is your feature request related to a problem or use case? Please describe.**

This optimization will try to address the concerns listed below.
1. Bundle size
2. Speed
3. Duplicate code included in bundles
4. Better code splitting strategy 
5. Lazy loading 
6. Better manage shared dependencies 

**Resources**
- [x] 1. [Code Splitting](https://webpack.js.org/guides/code-splitting/)
- [x] 2. [Build Performance](https://webpack.js.org/guides/build-performance/)
- [x] 3. [Dependency Management](https://webpack.js.org/guides/dependency-management/)
- [ ] 4. [Tree Shaking](https://webpack.js.org/guides/tree-shaking/)
- [ ] 5. [Lazy Loading](https://webpack.js.org/guides/lazy-loading/)
- [x] 6. Loading the cultures correctly

**Describe the solution you'd like**
TBD...

**Describe alternatives you've considered**
TBD...

**Additional context**
After this is done we should look at:

a) that this can load in our examples (NG, React Ect)
b) if we can fix the loading of cultures which is currently broken:

```
zone.js:1063 Unhandled Promise rejection: Loading chunk src_components_ids-locale_cultures_en-messages_js failed.
(error: http://localhost:4200/src_components_ids-locale_cultures_en-messages_js.js) ; Zone: <root> ; Task: Promise.then ; Value: ChunkLoadError: Loading chunk src_components_ids-locale_cultures_en-messages_js failed.
(error: http://localhost:4200/src_components_ids-locale_cultures_en-messages_js.js)
```

tmcconechy:
I also want to look at ESM https://webpack.js.org/guides/ecma-script-modules/ if its a solution for any of this.
Will sync up tomorrow.

This is probably going to be a joint PR while i do https://github.com/infor-design/enterprise-wc/issues/288

PR #310, opened 2021-08-23
288 - Project restructure

### LINKED ISSUE #288: Chore: Change Project Structure

We want the following second level structure in src

src
 - components
 - mixins
     - ids-mixin1 (in same folder)
     - ids-mixin2
utils
  - all utils from ids-base
core
  - rest of ids-base minus the mixins

Checklist

- [x] Do renaming of folders
- [x] rename app to examples or demos
- [x] example yaml stays
- [x] separate lib code from app code during compilation (someone pulling in component via script tags shouldn't get demoapp code) move index js code from sample app out
- [x] import from indexes
- [x] slim down the npm package (remove dev deps and examples)
- [x] create a static site for deploy for simpler pipeline

Do this when less branches/prs in progress

tmcconechy:

This PR solves two major things:
a) changes to agreed structure using git mv to preserve history (might have missed on or two but got the most of it)
```
demos (was app)
src
  - components
  - mixins
  - utils
  - core
```
b) components,mixins,core  now has an index and uss the index pattern so you can shorten your imports (i made all the corrections) for example `import { IdsEventsMixin, IdsThemeMixin } from '../../mixins';`
c) Makes depolying as a static site possible
d) slims down the dependencies in the dist
e) separates the demo app web pack config from the dist web pack config
f) uses new icons (so may see some small visual diffs)

**Important Notes**

1) the sub pages now require .html at the end for example http://localhost:8000/ids-tooltip/test-sandbox.html
2) will continue on https://github.com/infor-design/enterprise-wc/issues/301 after this is merged

Closes #288 

**General**
- pull
- npm prune && npm i
- npm run start and look around the demo app for any issues
- all tests pass

**Build Dist**
- `npm run build:dist` -> check output in dist
- `npm run build` -> check output in app

**Static Site**
- `npm run build`
- `cd demo-dist`
- `python -m SimpleHTTPServer`
- go to http://localhost:8000/ -> should be the same was with `npm run start`
- http://localhost:8000/ids-tooltip/test-sandbox.html -> should be the same was with `npm run start`
- http://localhost:8000/ids-block-grid/ -> images still serve
- http://localhost:8000/ids-data-grid/ -> data still serves
- http://localhost:8000/ids-upload-advanced/ -> will now only error on static deploy but this is fine (it works local and we dont want all kinds of uploads on our static deploys)


danielortiz1982 (approved):
I think these changes are great! I really like how clean the project looks, also it brings over some of the enterprise structure which will work well for consistency coming from the enterprise repo. The index import and object destructuring pattern are a good addition. For some reason the Python server is returning a 404 however the Node server run perfectly fine, not a deal breaker in my book. I think these changes are good to merge... I tested the branch with the steps listed below:

1. git fetch --all
2. git checkout project-restructure
3. rm -rf node_modules
4. rm -rf dist
5. rm -rf coverage 
6. npm ci
7. npm run build
8. npm start
9. npm run test
10. python -m SimpleHTTPServer
Screen Shot 2021-08-23 at 4 58 55 PM
Screen Shot 2021-08-23 at 5 08 04 PM
Screen Shot 2021-08-23 at 5 13 10 PM
Screen Shot 2021-08-23 at 5 13 16 PM

deep7102:
Looks nice!
all working fine and python server just need to adjust the path, looks it missing root `demo-dist`
http://localhost:8000/ids-tooltip/test-sandbox.html
http://localhost:8000/demo-dist/ids-tooltip/test-sandbox.html

deep7102 (approved):

rob2d:
Looks like you guys have this PR covered but just had one thought on the server added: I was wondering if Python was needed as a dep or maybe it would be worth it to add a simple lightweight one (?). 

If it is due to config or package size, there is stuff like `koa` or `hapi` which is basically analogous as what you would use express.js but more minimal and no middleware is included with Koa. 

Mainly, less deps help keep CI/other things lean as we don't have to introduce core deps for builds which would need to persist across OS (specifically Windows + Python 2/3 can be a bit of a pain to troubleshoot). This would especially be the case if we ever wanted to containerize for some reason, etc as you can pick an image with less deps or just be more flexible.

Example koa app serving content (this is a full static app once `koa` and `koa-static` plugin are added; at least to use locally for dev the way we are):

`server.js`
```js
var serve = require('koa-static');
var koa = require('koa');
var app = koa();

app.use(serve('./static-dir'));
app.use(serve('./another-static-dir'));

app.listen(8000);
```

If we wanted to iterate on it we could strip out the need of `.html` if we wanted to expand in it using a regular Router and that would be also a few lines since it could be done recursively.

Not totally necessary since having a static server is a good improvement but just sharing an idea (and I'm not 100% aware on the context of it either e.g. whether there were other benefits of using Python in this instance that I can learn about).

tmcconechy:
@danielortiz1982 - As Deepak pointed out i missed a step to `cd demo-dist` before starting the server as that should be the root directory.

@rob2d I like koa as well but the python http server was just a simple one prove the static site would work. Any server you want to test it with should work and thats the point. But for what we are including for Dev its still the node/express webpack dev server since that works with live reload so we are still using that and dont need to include anything else.

PR #319, opened 2021-08-30
279 - tabs header variant, content/context, attribute provider iteration

### LINKED ISSUE #279: IdsHeaderTabs:  Add new web component (or tab setting)

Add a ids header examples. Try and use the placement in the header and inverse. Or add a setting. Finally new component if necessary.

Also any tabs cleanup and todos you want to resolve https://github.com/infor-design/enterprise-wc/blob/main/src/ids-tabs/TODO.md with aim to get header tabs working

**Phase 1 Features (MVP)**

- Simple default example https://main-enterprise.demo.design.infor.com/components/tabs-header/example-index.html
- Example with additional text area ? Extended Header? https://main-enterprise.demo.design.infor.com/components/personalize/example-form.html

**Possible Internal Markup**
```js
  <ids-header>
    ..
    <ids-tabs>...</ids-tabs>
  </ids-header>
```

**Suggested Dependencies**
- ids-inverse-color-mixin (indirectly?)
- ids-tabs
- ids-header

**Other**

rob2d:

(1) iteration on `ids-tabs`:
- adds `ids-tabs-context` & `ids-tabs-content` components:
  - `ids-tabs-context` passes `active` flag to the `ids-tabs-content` that has a matching `value` attribute in it's DOM.
  - listens to `ids-tab/s` callback to update its value/pass that.
- examples: add tab content, standalone CSS and some things missing on first go around.
- change `ids-tabs`/`ids-tab` to use callbacks so it would be easier to refactor later.
- added `color-variant="alternate"` for header tabs. When ids-tabs gets placed in a header this would be set automatically (but is overridable).

(2) iterate on `ids-attribute-provider-mixin`: 
  - added an optional `valueTransformer` param to set `ids-tab-content.active` based on values when updated.
  - passes `depth` option (changed to BFS vs DFS to do that)
  - iterated on the interface to just accept object entries so it's easier for mapping up from children on changes (iterative/WIP).
  - accepts`attributeProviderDefs` as an argument, since will need to create lookup maps that are shared between classes once it allows for two-way binding with child => parent (related to above).

(3) some small adjustments on `expectEnumAttributeBehavior` util since some coverage wasn't being reached.

⚠️ Ready for review on the component changes but need to add some notes on documentation so marked WIP ⚠️
 
Closes #279 

- pull branch and reset dev server
- visit: 
  - http://localhost:4300/ids-tabs 
  - http://localhost:4300/ids-tabs/header-tabs.html
  - http://localhost:4300/ids-tabs/standalone-css.html
- make sure functionality is alright there (standalone is just a skeleton)


demos/ids-tabs/standalone-css.js
```diff
@@ -0,0 +1,70 @@
...
+ * or indicative of best practices for your
+ * application
+ */
+class TabController {
+  constructor(idsTabsElement) {
+    this.tabContainers = [...idsTabsElement.querySelectorAll('.ids-tab-container')];
+    this.tabContainers?.[0]?.focus();
+
+    this.tabContainers?.forEach?.((t, i) => {
+      t.addEventListener('click', () => { this.selectedIndex = i; });
```
danielortiz1982, line 16: This looks good, however I would recommend a small change to check for a click target and apply one event handled for just the ids-tab-container you want to work with. That will avoid creating a fn for DOM elements that are not currently being clicked on. Since this is the standalone js I think it's fine.

demos/ids-tabs/standalone-css.js
```diff
@@ -0,0 +1,70 @@
...
+    underlineDiv.classList.add('selection-underline');
+    this.tabContainers[this.#selectedIndex].children[0].appendChild(underlineDiv);
+  }
+
+  get selectedIndex() {
+    return this.#selectedIndex;
+  }
+}
+
+window.addEventListener('load', () => {
```
danielortiz1982, line 67: Same comment as above.

src/components/ids-tabs/ids-tabs.js
```diff
@@ -69,16 +74,19 @@ export default class IdsTabs extends mix(IdsElement).with(
...
-
     // set initial selection state
     this.#updateSelectionState();
+
+    /* istanbul ignore next */
+    if (!this.hasAttribute(attributes.COLOR_VARIANT)) {
+      this.#checkAndSetColorVariant();
+    }
+
+    this.onEvent('tabselect', this, (e) => /* istanbul ignore next */ {
```
danielortiz1982, line 85: Looks good, however... Same comment as above, add this to it's own #handleEvents fn.

src/components/ids-tabs/ids-tabs.js
```diff
@@ -176,14 +179,15 @@ export default class IdsTabs extends mix(IdsElement).with(
...
-   * used to detach event listeners properly
+   * Used to detach event listeners properly
    * @type {Set<string>}
    * @private
    */
   #tabValueSet = new Set();
 
   /** observes changes in tabs */
   #tabObserver = new MutationObserver((mutations) => {
+    /* istanbul ignore next */
```
danielortiz1982, line 190: This seems like a good solution, I was wondering can this be done in a setter/getter instead???

src/components/ids-tabs/ids-tabs.js
```diff
@@ -282,6 +314,7 @@ export default class IdsTabs extends mix(IdsElement).with(
 
     // clear tab values tracked
 
+    /* istanbul ignore next */
     for (const tabValue of this.#tabValueSet) {
       this.offEvent(`click.${tabValue}`);
```
danielortiz1982, line 319: I am not 100% sure what this is doing, however if your working with an event I think you can do this with targets and reduce the amount of iterations being done. Again, I am not sure what your goal is here so it's just a recommendation.

src/mixins/ids-attribute-provider-mixin/ids-attribute-provider-mixin.js
```diff
@@ -11,102 +11,165 @@ const identityFn = (v) => v;
...
+      scannedEl = this,
+      depth = 1
+    ) {
+      const domNodes = [
+        ...(scannedEl.children || []),
+        ...(scannedEl.shadowRoot?.children || [])
+      ];
 
-          if (!component || !(el instanceof component)) { continue; }
+      for (const element of domNodes) {
```
danielortiz1982, line 104: I am still a little unclear what this is doing, however I think three nested for loops is a bit excessive. Maybe we can fine a time to talk this solution over a little more.

demos/ids-summary-field/demo.scss
```diff
@@ -0,0 +1,12 @@
+.tab-container {
```
tmcconechy, line 1: This do anything?

demos/ids-tabs/index.yaml
```diff
@@ -15,3 +15,5 @@
     description: Showing a tab element side by side working with 4.x
   - link: standalone-css.html
     description: Showing limited tabs that works with css only
+  - link: header-tabs.html
+    description: Tabs within header and content in body updating.
```
tmcconechy, line 19:
```suggestion
    description: Header Tabs and content in body updating.
```

Just thinking should refer to it as header tabs as thats whats known

src/components/ids-tabs/README.md
```diff
@@ -32,7 +32,29 @@ A vertical tabs component.
 </ids-tabs>
 ```
 
-TODO: phase i examples
+Using a tab context auto-updates the active `ids-tab-content`.
```
tmcconechy, line 35:
```suggestion
Using a tab context to show the active tab content in `ids-tab-content`.
```

More clear?

src/components/ids-tabs/README.md
```diff
@@ -55,7 +77,7 @@ TODO
 
 ## States and Variations (With Code Examples)
 
-TODO
+Ids Header
```
tmcconechy, line 80: Should we document the header tabs as well as the "updating content"?

src/components/ids-tabs/ids-tab-content.js
```diff
@@ -0,0 +1,92 @@
...
+    /* istanbul ignore next */
+    this.setAttribute(attributes.VALUE, value);
+  }
+
+  /* istanbul ignore next */
+  get value() {
+    return this.getAttribute(attributes.VALUE);
+  }
+
+  set active(value) {
```
tmcconechy, line 77: Some comment on this one?

src/components/ids-tabs/ids-tab.js
```diff
@@ -30,7 +30,7 @@ const { stringToBool, buildClassAttrib } = IdsStringUtils;
  */
 @customElement('ids-tab')
 @scss(styles)
-class IdsTab extends mix(IdsElement).with(IdsEventsMixin) {
+export default class IdsTab extends mix(IdsElement).with(IdsEventsMixin) {
```
tmcconechy, line 33: Is  this right? Maybe should keep it at the bottom like every other component?

src/components/ids-tabs/ids-tabs-context.js
```diff
@@ -0,0 +1,93 @@
...
+    targetAttribute: attributes.ACTIVE,
+    valueTransformer: ({ value, element }) => element.getAttribute(attributes.VALUE) === value,
+    cliffTags: ['IDS-TABS', 'IDS-TAB-CONTENT']
+  }],
+  attributesListenedFor: [],
+  maxDepth: 8
+};
+
+/**
+ * IDS TabsContext Component
```
tmcconechy, line 37:
```suggestion
 * IDS Tabs Context Component
```

src/components/ids-tabs/ids-tabs-context.js
```diff
@@ -0,0 +1,93 @@
...
+ * @type {IdsTabsContext}
+ * @inherits IdsElement
+ * @mixes IdsEventsMixin
+ * @mixes IdsThemeMixin
+ * @mixes IdsKeyboardMixin
+ * @part container - the container of all tabs
```
tmcconechy, line 43: Dont think this is being used?

src/ids-mixins/README.md
```diff
@@ -0,0 +1,268 @@
+## Ids Mixins
+
+Mixins are simply functions with shared functionality that can be injected into a component. For example the IdsEventOmitter. They get around the issue that in JS that you cannot inherit from more than one object. Also they prevent the Base Element from getting bloated with functionality that not every component uses. Ids is using a simple object as a mixin that in "injected" into the component in the constructor and then used according to its documentation. If the mixin has UI elements it should probably be a web component instead.
+
+## Ids Keyboard Mixin
```
tmcconechy, line 5: Something wrong with the merge here because now each mixin has its own README.md in the folder. 

This root readme should just be about "what are mixins"

src/components/ids-tabs/ids-tabs.js
rob2d: The mutation observer observes "childList" which looks for mutation in the DOM -- basically if a user changes the structure of the IdsTabs which they can do through appendChild/innerHTML/etc. That can't be reproduced with a getter/setter

tmcconechy (changes requested):
A few minor doc suggestions and nits. 

- The only bigger one is the mixin readme got some extra stuff back in it
- Also for placeholder text we decided to use https://hipsum.co/ so it just placeholder and nothing to read. For the tabs could we use `type="p"` so there is more padding ect. See https://github.com/infor-design/enterprise/issues/5325 (we might want to update ids-text to show proper p type padding and line height)

In terms of functionality i see a couple things:
- click on the theme switcher then tab into the tabs on http://localhost:4300/ids-tabs/header-tabs.html its not showing the focus on the selected tabs
- use arrow left and hit space -> should select the tab but only enter is working see https://main-enterprise.demo.design.infor.com/components/tabs-header/example-index.html

src/components/ids-tabs/ids-tabs.js
rob2d: ln 314:   // clear tab values tracked

rob2d: So when the component mounts/unmounts, the listeners are tracked (comment is above the line where you commented). The tab events are namespaced with their value e.g. `click.value-a`. 

They need to be unsubscribed when the tabs change e.g. user edits innerHTML or appends a new tab, or the tab values change. 

Otherwise there will be memory leaks. It doesn't have to do with when the events themselves are called;  only happens on initial bind e.g. if child list is changed or component is re-rendered/connected.

src/components/ids-tabs/ids-tabs-context.js
```diff
@@ -0,0 +1,93 @@
...
+  template() {
+    return '<slot></slot>';
+  }
+
+  connectedCallback() {
+    super.connectedCallback?.();
+    this.provideAttributes();
+
+    /* istanbul ignore next */
+    this.onEvent('tabselect', this, (e) => {
```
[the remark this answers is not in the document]
rob2d: So on the context this is purposefully a really simple component (just there to provide values); there is no other place these are binded on the context, and offEvent cleans itself. So seems adding a method there is a little overkill since it only happens once on lifecycle and events are auto-unbinded.

There's another issue which is the naming convention `this.#handleEvents` has some issues (e.g. it gives the connotation the event handler is handling the event on the component, but actually it's binding events). That's sort of an unrelated issue but the convention is a little confusing since "event handler" interfaces give the impression that calling them is having them handled", while with that we're kinda saying "handle" the event... but its pretty ambiguous e.g. if the event is handled its firing.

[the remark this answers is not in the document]
tmcconechy: Depends how you look at it. It is binding the events but then it adds a handler for it too. So basically both half make sense. We just always used handleXXX

demos/ids-tabs/standalone-css.js
rob2d: Since we're binding the index, it was done that way to just grab the index that was iterated on during the click. Otherwise need to use O(N) to scan vs simple setter on-click. 

(if it was not a specific circumstance though it would be a good idea 👍)

src/mixins/ids-attribute-provider-mixin/ids-attribute-provider-mixin.js
rob2d: 80/20 rule. This is just walking the DOM to find children elements; the result is the work saved on components that use it.

src/components/ids-tabs/ids-tab.js
rob2d: It's valid syntax and seemed a little more compact, but I'll change it to match the other components also.

tmcconechy: Yeah keep it the same as much as possible for example as i was looking for it at the bottom when doing other things as i recall. This would trip me up.

src/components/ids-tabs/README.md
rob2d: yeah this one was just WIP but I got a little sidetracked troubleshooting an uncaught bug on a setter in the button yesterday. Am just going to sort the minor things and then try to add a clear example before removing the WIP label here thanks

demos/ids-summary-field/demo.scss
rob2d: good catch this was outdated after some changes, thanks.

src/components/ids-tabs/ids-tabs-context.js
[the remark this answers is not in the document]
rob2d: I understand the different perspective, I was more saying it in the context of programming terminology on the web and maybe even resources in libraries we use. Some use cases of `handle` elsewhere, which make this a bit fuzzy:

- in React and several other libs `handle` is used for the action the method is taking on a callback.
- in NodeJS' `EventEmitter` (which is kind of the core to most the stuff going on in Node since it's event based), they use `handle` everywhere to describe when a callback is executed (vs binded)
image
- on sites like W3 schools or MDN they use the verb "handle" to mean a callback that is being executed by a pub/sub
image

There is actually even a `handleEvent` method in native JS objects here too -- this gets fired when any event is called, vs to add the event: can see an example here: https://dev.to/rikschennink/the-fantastically-magical-handleevent-function-1bp4

The term "event handler" also means the method itself just to clarify, as in the method itself that is "handle-event"ing is a "handler", not the over all component. Just as a reference: it's not something I'm inventing here, its just sort of the term for this:
https://en.wikipedia.org/wiki/Event_(computing)#Event_handler

I am ok with if we want to keep doing what we're doing at this point (though I don't think this sub needs a specific method specifically), but I guess just an fyi that it can create some fuzzy thinking to use the same word to mean different things with events in overall ecosystem of JS.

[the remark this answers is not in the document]
tmcconechy: I dont mind a different name. But for consistency in finding the event handler code i think its good to have it in specific component method. So if your just avoiding doing that because you dont like `handle` lets use something else.

src/mixins/ids-attribute-provider-mixin/ids-attribute-provider-mixin.js
```diff
@@ -11,102 +11,165 @@ const identityFn = (v) => v;
  */
 const traversibleHTMLTags = new Set(['DIV', 'SPAN']);
 
+/**
+ * @typedef AttributeProvidedDef
+ * @property {string} attribute the type of attribute
+ * @property {string} targetAttribute the attribute being targeted
+ * on the child which is assigned
+ * @property {Function} valueXformer transforms the value assigned
+ * @property {Array} cliffTags tags where we know we won't find our
```
tmcconechy, line 20: stopAtTag? or some other name?

src/mixins/ids-attribute-provider-mixin/ids-attribute-provider-mixin.js
```diff
@@ -11,102 +11,165 @@ const identityFn = (v) => v;
  */
 const traversibleHTMLTags = new Set(['DIV', 'SPAN']);
 
+/**
+ * @typedef AttributeProvidedDef
+ * @property {string} attribute the type of attribute
+ * @property {string} targetAttribute the attribute being targeted
+ * on the child which is assigned
+ * @property {Function} valueXformer transforms the value assigned
```
tmcconechy, line 19: valueTransform

src/mixins/ids-attribute-provider-mixin/ids-attribute-provider-mixin.js
rob2d: @tmcconechy would `sentinelTags` make sense? (just from perspective of it terminating the search at those points e.g. https://en.wikipedia.org/wiki/Sentinel_value /  https://en.wikipedia.org/wiki/Sentinel_node)

tmcconechy: Maybe something less "fancy". I've never heard of that? Like `exitTags`,  `breakTags`( like break in a loop) or `tripTags` (sentinel value also referred to as a flag value, trip value, rogue value, signal value, or dummy data) ?

deep7102: `excludeTags`, `ignoreTags`

tmcconechy: I like `exitTags` better.

OK, Maybe i heard of it way back or something and the concept did resonate when i read the link so i learned a new word again. But the main reason for suggesting something simpler is i think it's more likely people (developers/users) will be able to tell with the setting does being something they don't have to google to understand.

I'm only commenting on Apis and things when the names aren't as clear when there are other options or seem confusing or difficult to follow. If we get locked in to some named API that causes confusing it's hard to change it. I know this is internal but that's my general thought pattern when these come up.

rob2d: @tmcconechy thanks on the explanation/reasoning for that. Will do.

tmcconechy:
@rob2d to clarify is this ready to review again (still see WIP tag?)

tmcconechy:
@rob2d ready to approve when the checks pass :)

tmcconechy (approved):

deep7102 (approved):

PR #340, opened 2021-09-14
143 - ids splitter (WIP)

rob2d:

- adds the `ids-splitter` and `ids-splitter` component.
- some minor logging was missing on `ids-draggable` events.

- N/A, branch is WIP


rob2d:
Earlier today I had to wrap up a few other things, so hadn't quite gotten this component to where I wanted to get it before leaving. 

There was a stack overflow when components mount, and this is because when `ids-splitter` mounts, it thrashes the setter with the check and values. I think there may need to be a 3rd state outside of internal style but didn't have time today to come up with a solution that wasn't invasive enough to implement in a few hours. 

Mainly wanted to get to the point of showing the idea of resizing and positioning the ids-draggables automatically, nesting, and responding to size changes.
----------------
Some notes on `ids-splitter`:
- `#dimension` tracks the axis as a CSS parsible value (e.g. `width`/`height`).
- size will track both width/height to reference `#dimension` so that code doesn't become complex for both axes
- ResizeObserver is just there to track the size which is used to size the panes
- the `paneDraggableMap` maps each pane (as reference to `IdsElement` being key) to metadata about the draggables that are auto-created. There is possibly a `before` and `after` draggable (ref to the draggables that surround that-specific-pane).
- #repositionDraggable sets the draggables to be along the splitter. By auto updating using this method when a drag ends, we get a simple way to get draggable in a predictable position.
- `#updatePaneSizes` is where the math happens for updating the splitter panes and it parses the minSizePx, maxSizePx and sizePx. This data is taken when a pane updates by listening to `splitter-pane-size-attrib-change`.
- There was an issue with circularly updating which needed another state var in `ids-splitter-pane` where firing splitter-pane-size-attrib-change` on init did not work as intended properly, but the purpose of this mechanic was to DRY things. May need a workaround because WC register the *child* pane before parent when running connectedCallback. Alternatively, could loop through children and add childList entry on MutationObserver.

Some notes on the mechanics of `ids-splitter-pane` sizing (how it exists now and the intention)
- The `sizesHash` is used to diff `min-size`, `max-size`, and `size` from prev/next values to prevent retriggering updates (there is currently a logic loop on this that has to be sorted).
- setting a size property (including min/max types) causes the `{ number, unit }` to be set in a private var matching the name of the var e.g. #size/#minSize/#maxSize. This would just let us calculate the `%` as `px` for size values. 
- when the size property updates, its resized by the `ids-splitter` since ids-splitter ideally will do the math in one core place based on it's width/height (which depends on it's axis).
---------------
Hope that is helpful and apologize to leave this one hanging particularly because it's pretty involved.

Thanks everyone!

tmcconechy:
Thanks @rob2d - @deep7102 is picking this up and will do a new PR. Leaving the branch but closing the PR for now

Issue #330, opened 2021-09-08
IdsContextualActionPanel: New Web Component


tmcconechy:
Add a new `IdsContextualActionPanel` (ids-contextual-action-panel) component MVP to the WebComponent repo based on https://main-enterprise.demo.design.infor.com/components/contextualactionpanel

Think it should work like the modal possibly inheriting from it. Just change the animation and the toolbar and styling. Should be possible to set the content with the shadowDOM/slots not using a buttons setting like before

**Phase 1 Features (MVP)**
- Animation
- Toolbar
- Contents
- Events fire
- Modal Breakpoints/fullsize
- use new design https://github.com/infor-design/enterprise/pull/5688

**Phase 2 Features (MVP)**
- Most Anything else 

**Possible Internal Markup**
```html
<ids-contextual-action-panel id="my-cap" aria-labelledby="my-cap-title">
  <ids-text slot="title" font-size="24" type="h2" id="my-cap-title">Active IDS CAP</ids-text>
  <ids-toolbar slot="toolbar">
  </ids-toolbar>
  <div slot="contents">
     <!-- contents -->
  </div>
</ids-contextual-action-panel>
```

**Suggested Dependencies**
- ids-modal

**Design**
https://main-enterprise.demo.design.infor.com/components/line/example-index.html

**Events**
- beforeopen, open, afteropen
- beforeclose, close, afterclose

Close are vetoable
https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-tag/ids-tag.js#L225-L233

**Other**
Pass the following checklist ordered in priority (some may be left for later) https://github.com/infor-design/enterprise-wc/blob/main/doc/CHECKLIST.md

PR #349, opened 2021-09-20
295 - Ids Popup/Modal/Message Improvements

### LINKED ISSUE #295: IdsMessage: Overlay/Position Issues
**Describe the bug**
The message/modal component has an issue when working with the theme changer

**To Reproduce**
1. Go to http://localhost:4300/ids-message
2. Click on ...
3. See the modal move and the overlay is misplaced and in the first place the menu shouldnt be openable
4. because of the percy image for the message is "flaky"

**Expected behavior**
1. The popup theme switcher should not be clickable when the overlay is up (but modals with menu buttons should work)
1. The overlay should cover the page

We could also now use the header on the page at the top if that helps

**Screenshots**
Screen Shot 2021-08-05 at 9 33 27 AM

EdwardCoyle:
This PR makes a handful of changes to the IdsPopup component and some components that depend on the Popup.  These changes include:
- Separates the concept of placement from visibility change in the Popup, for finer-grain control of the lifecycle of Popup-based components.
- Adds `position-style="viewport"` to the Popup specifically for Modal-style centering within the viewport with CSS transforms (replaces `setModalPosition` and JS-based calculation for centering the Modal and fixes one part of #295)
- Changes the way overlays are positioned so they no longer get pushed around by other DOM elements (fixes the other part of #295)
- Replaces `this.state` properties with class-level private properties (per recent dev discussions).
- Make many methods in IdsPopup async compatible (show/hide/place/refreshVisibility).
- Implement new Popup lifecycle into Modal/About/Message/Tooltip components.
- Removes the IdsResizeMixin (per recent dev discussions, it's not needed as ResizeObserver is more performant when used directly)
- Adds the loading of the Resize Observer mock globally to `jest.config.js` to accommodate direct use of ResizeObserver in source code.
- Adds missing/new docs and Typescript defs

Closes #295 

Pull, build, run, then:

_Test the original issue #295:_
- Open http://localhost:4300/ids-message/visible.html
- When the Message is displayed, the Overlay should be in the correct place and cover the entire viewport.  The Theme Switcher button should not be usable.
- Try refreshing the page several times.  When displayed, the Message should not appear in a different place each time.
- Resize the browser viewport.  The Message should remain centered in the viewport.  The overlay should continue to sit over top of the rest of the page.

_Test other changes:_
All tests should pass.  Also, smoke test the following components to ensure everything is working as expected (specifically Popups opening/closing/being in the correct place):
- http://localhost:4300/ids-popup
- http://localhost:4300/ids-popup/test-sandbox.html
- http://localhost:4300/ids-about
- http://localhost:4300/ids-message
- http://localhost:4300/ids-tooltip
- http://localhost:4300/ids-tooltip/test-sandbox.html


EdwardCoyle:
@andriivarhanov Thanks for the feedback!  I addressed the problem with the Escape key not closing the Modal when no inner elements were focused.  I've left the sizing problem alone for now -- I still need to do further iteration on the Modal to get the `fullsize` setting added to the WebComponents, which I need to complete #330, so I'll address Modal boundaries in that issue.

andriivarhanov (approved):
@EdwardCoyle looks great, Escape key works now, thanks

src/components/ids-tooltip/ids-tooltip.js
```diff
@@ -117,33 +118,33 @@ class IdsTooltip extends mix(IdsElement).with(
...
+        this.#hide();
       });
       this.onEvent('click.tooltip', targetElem, () => {
-        this.visible = false;
+        this.#hide();
       });
 
       // Long Press
       this.onEvent('longpress.tooltip', targetElem, () => {
-        this.visible = true;
```
tmcconechy, line 131: Im a bit confused about this change? I liked using the setter to set the visibility over a method? Does this remove that ability?

tmcconechy (commented):
This might be out of scope but i noticed that the about appears disabled in dark mode now that the images are working https://images.percy.io/3cd2951044ff6c806080709f9712bc21617969d9ddcb0672e0f08ff1c5d7e285

src/components/ids-tooltip/ids-tooltip.js
EdwardCoyle: I had changed this originally to try to "normalize" one change I made to the Popup, where the `visible` setter is no longer tied to actually displaying/hiding the Popup (`hide()/show()` must be explicitly called to change its visibility because of some lifecycle issues).  I realized this isn't necessary to do in the Tooltip so I've reverted it back.  Using the setter is all you need for Tooltip

EdwardCoyle:
@tmcconechy I separated Modal styles from the Menu in this PR but hadn't written theme styles against the `modal` type.  Just added those, good catch.

tmcconechy:
Only one thing i see an extra border on the about next to the button now.
Screen Shot 2021-09-23 at 10 17 16 AM

Coverage dropped but i lowered the threshold to 90% for now. Might want to fix since it was 100 in tooltip

andriivarhanov:
@EdwardCoyle it looks like I didn't add theme mode colors for ids-about content body area, my bad, can I include it to your PR? https://github.com/infor-design/enterprise-wc/commit/e4a41d64e781499d65e7f73b510d3339fa11ac31 

image

tmcconechy (approved):

deep7102 (approved):
Looks good 🍌

PR #370, opened 2021-09-30
305 icon notification

Jah25:

Adding functionality for notification badges to the ids-icon function. allows the user to add the notification circle to one of the four corners in five color options

closes [#305](https://github.com/infor-design/enterprise-wc/issues/305)

- run project locally
- navigate to http://localhost:4300/ids-icon/
- example should be present for new position and color options


demos/ids-icon/example.html
```diff
@@ -1,3 +1,5 @@
+{{> notification-badge.html}}
```
tmcconechy, line 1: Lets take this out and just let it be on its own sub page.

demos/ids-icon/example.js
```diff
@@ -1,5 +1,6 @@
 // Imports Used in the Examples
 import pathData from 'ids-identity/dist/theme-new/icons/standard/path-data.json';
+import './notification-badge';
```
tmcconechy, line 3: Ditto would move to notification-badge.hs

demos/ids-icon/notification-badge.html
```diff
@@ -0,0 +1,59 @@
+<ids-layout-grid>
+  <ids-text font-size="12" type="h1">Ids Icon with notification badge</ids-text>
+</ids-layout-grid>
+<ids-layout-grid>
+  <ids-layout-grid-cell>
+   <ids-component>
```
tmcconechy, line 6: This a mistake?

demos/ids-icon/notification-badge.html
```diff
@@ -0,0 +1,59 @@
+<ids-layout-grid>
+  <ids-text font-size="12" type="h1">Ids Icon with notification badge</ids-text>
+</ids-layout-grid>
+<ids-layout-grid>
```
tmcconechy, line 4: Check other grid examples i think you need `<ids-layout-grid auto="true">` and also need `ids-layout-grid-cell`

demos/ids-icon/notification-badge.scss
```diff
@@ -0,0 +1,8 @@
+.notification-container{
```
tmcconechy, line 1: Instead of doing this could we use the ids-layout grid properly?

src/components/ids-icon/ids-icon.scss
```diff
@@ -1,10 +1,12 @@
 @import '../../core/ids-base';
+@import '../../mixins/sass/ids-colorpalette-variables.scss';
```
tmcconechy, line 2: I think this is also in ../../core/ids-base ?

src/components/ids-icon/ids-icon.scss
```diff
@@ -14,3 +16,57 @@
 .flipped {
   transform: scaleX(-1);
 }
+
+.notification-badge {
+  border-radius: 4px;
+  height: 6px;
+  width: 6px;
+  background-color: red;
```
tmcconechy, line 24: Should use the color var.. Is it `--ids-color-status-caution` or `--ids-color-status-danger`? The read one

tmcconechy (changes requested):
Just a few things.. one thing to check and i don't see why in the PR is that http://localhost:4300/ids-icon seems to be all one column but i fixed that (could be on an open PR tho)

demos/ids-icon/notification-badge.html
Jah25: yeah i think this might have happened in the merge

demos/ids-icon/example.html
Jah25: yeah I had intentionally imported it into the main example because I had seen some other components consolidate the examples but I'll pull it out and let it stand alone

src/components/ids-icon/ids-icon.scss
Jah25: it's --ids-color-status-danger sorry I thought I had gotten that one.

src/components/ids-icon/ids-icon.scss
Jah25: I didn't think so but I'll try it and see what happens

tmcconechy: Yeah those are just the sass versions of the variables. You might use them if you need to loop the colors in sass https://github.com/infor-design/enterprise-wc/blob/4fcf691099d53b59acc3baa83459a3b553858732/src/mixins/sass/ids-checkbox-mixin.scss#L183

But the core/ids-base has the css variables already:
Screen Shot 2021-10-04 at 9 54 39 AM

src/components/ids-icon/README.md
```diff
@@ -31,11 +31,32 @@ Icons come in 4 sizes depending where it is used.
...
+<ids-icon icon="notes" badge-position="bottom-left"></ids-icon>
+<ids-icon icon="notes" badge-position="bottom-right"></ids-icon>
+```
+These badges can also be displayed in 5 possible colors
+
+```html
+
+<ids-icon icon="notes" badge-position="top-left" badge-color="base"></ids-icon>
+<ids-icon icon="notes" badge-position="top-left" badge-color="caution"></ids-icon>
+<ids-icon icon="notes" badge-position="top-left"></ids-icon badge-color="danger">
```
andriivarhanov, line 48: attribute in closing tag, please move to the left part

andriivarhanov (commented):
looks good, a couple of notes:
- is there a way to remove notification badge after it was initialized? Something like
```js
const icon = document.querySelector('ids-icon')
icon.badgeColor = null
icon.badgePosition = null
```
should the functionality exist?

- I think notification badge example page http://localhost:4300/ids-icon/notification-badge.html should have different icons like https://main-enterprise.demo.design.infor.com/components/notification-badge/example-index.html does

src/components/ids-icon/README.md
tmcconechy: I agree it probably would be good to set the settings to null to remove the badges in case the notification needs to be turned off. Good suggestion 👍🏻

Jah25:
Just to confirm, we are no longer supporting the purple option correct?

tmcconechy:
yes i can confirm purple is out of the design system now

PR #411, opened 2021-11-04
345 - IdsTabs Improvements

### LINKED ISSUE #345: Ids Tabs Improvements
Need to review the current IDS Tabs component and review for accuracy, complexity

- [ ] Review current IDS Tabs solution
- [ ] Simplify the implementation if we can
- [ ] Rewrite types/tests/docs to match any changes
- [ ] Rework examples

EdwardCoyle:
This PR refactors parts of the IdsTabs component to simplify its design:
- Got rid of stored internal references to child tabs, reducing the need to remove/reapply event handlers
- Reworked keyboard navigation to not depend on said references
- Reworked events not to use indexes for referencing tabs but `e.target` (again to reduce looping for detection)
- Fixed setter logic in many places to prevent double-setting from occuring, removed `#prevSelected` since its unneeded after this.
- Fixed bugs related to setting junk values referencing non-existent tab content (setters should ignore these)
- Broke out logic for Orientation into its own `IdsOrientationMixin` (could be used elsewhere)
- Updated tests/types/docs to match

No visual states or interactions should change with this PR, and it should behave pretty much the same as before, aside from the bug fixes.

Closes #345

- Pull/Build/Run
- Open http://localhost:4300/ids-tabs
- By default the "Contracts" tab should be selected
- Clicking the tabs should change the tab clicked to "selected", and should deselect the previously selected tab.
- After clicking a tab, use the left/right arrow keys to navigate.  It should be possible to cycle focus in both directions, and focus should loop between first/last tabs.
- Pressing "Enter" should select a keyboard-focused tab.
- Pressing "Home"/"End" keys should set focus to the first/last tabs respectively.
- Open a dev tools console.  Inspect the `<ids-tabs>` element and change its value attribute to match one of the `<ids-tab-content>` pane's value attributes.  The selected tab should be updated in the DOM.
- Open http://localhost:4300/ids-tabs/vertical.html.  Try the same keyboard tests, but use the up/down arrows.
- Check counts on http://localhost:4300/ids-tabs/counts.html
- Check the color variant and nested tabs on http://localhost:4300/ids-tabs/header-tabs.html

### DIFF
--- a/demos/ids-tabs/example.html
+++ b/demos/ids-tabs/example.html
@@ -7,7 +7,7 @@
     <ids-tab value="contracts">Contracts</ids-tab>
     <ids-tab value="opportunities">Opportunities</ids-tab>
     <ids-tab value="attachments">Attachments</ids-tab>
-    <ids-tab value="notes">Notes</ids-tab>
+    <ids-tab value="notes" selected>Notes</ids-tab>
   </ids-tabs>
   <div class="ids-tabs-content">
     <ids-tab-content value="contracts">
--- a/demos/ids-tabs/example.html
+++ b/demos/ids-tabs/example.html
@@ -7,7 +7,7 @@
     <ids-tab value="contracts">Contracts</ids-tab>
     <ids-tab value="opportunities">Opportunities</ids-tab>
     <ids-tab value="attachments">Attachments</ids-tab>
-    <ids-tab value="notes" selected>Notes</ids-tab>
+    <ids-tab value="notes">Notes</ids-tab>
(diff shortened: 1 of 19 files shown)


EdwardCoyle:
Waiting for #412 to fix that test problem

danielortiz1982 (approved):
@EdwardCoyle This is much cleaner, good use of event delegation here... One thing, when this is ready please remove any remaining console log statements. Great job!

tmcconechy (approved):

PR #418, opened 2021-11-10
Chore: Remove IdsAttributeProvider mixin from IdsAccordion

EdwardCoyle:
In an effort to simplify component hierarchy we wanted to remove the IdsAttributeProviderMixin from IdsAccordion.  The solution in this PR maintains the same functionality (which was just managing locale/language at the Accordion level instead of having each child component detect it), but does so by setting the property on children directly, instead of with the attribute provider.

Related to #301

- Smoke test http://localhost:4300/ids-accordion and http://localhost:4300/ids-app-menu
- Ensure all tests pass


EdwardCoyle:
Closing for now since the Locale implementation may change, we can re-open if we keep it

tmcconechy:
Going to try and revive this as i need it to solve accordion bugs. I didn't need to change locale enough to impact this..

tmcconechy (approved):

tmcconechy:
Dont need this after all. I encorporated it with the locale changes on the web pack branch in the end.

PR #421, opened 2021-11-11
419 - Add Autofit to Some Components

vivz753:
**Explain the details for making this change. What existing problem does the pull request solve?**

TODO: change base branch to main 
* merge in PR for #414 first

**Related github/jira issue (required)**:

**Steps necessary to review your pull request (required)**:

**Included in this Pull Request**:
- [ ] An e2e or functional test for the bug or feature.
- [ ] A note to the change log.

PR #429, opened 2021-11-16
419 autofit components

### LINKED ISSUE #419: Chore: Add auto height and fit options to several components
**Describe the Ask**

For several key components the heights can either be a default, fit to whats inside it or expand down to the bottom of the page. The rules are as follows:

Virtual Scroll -> Dynamic by default and can pick up fixed height
Card -> 360px fixed by default - but set to auto-fit or auto-height
List View -> auto-fit by default and nothing else right now (so really card or page of the parent)
Datagrid -> Number of rows or auto-fit (future option)

**Expected behavior**
Add auto-fit and auto-height to the above components:

Virtual Scroll -> Just dynamic to what its in
Card -> Fixed default with both auto-height and auto-fit. Also for the example remove the code that is hiding the header and take the header out of the example to make it work the same way.. So the header can be optional.
List View  -> Auto-fit only
Datagrid -> Fixed default with both auto-fit added

**Other Rejected Ideas**
Use a different set of options
```
height="auto" -> auto-fit
height="100vh" -> fit down? but from screen
height="100%" ->  fit down? but from parent
height="400px" -> fixed??
```

Other rejected name ideas.
```
autogrow = auto-height
full-height
content-fit
size-to-parent
fit-parent
fit-content
size-parent
size-content?
```

vivz753:

**ids-list-view**
- Adjusts to parent container by default

**ids-data-grid**
- Added `auto-fit` attribute, which grows the grid to the parent-container height
- Added dedicated example `auto-fit.html` to display how `auto-fit` is used

**ids-card**
- Renamed `auto-size.html` to `auto-height.html`
- Remove the CSS that hides the header when `auto-height` is enabled (user can just remove header if they don't want it)

 - Closes #419 

http://localhost:4300/ids-list-view
http://localhost:4300/ids-data-grid/auto-fit.html
http://localhost:4300/ids-card/auto-height.html


demos/ids-card/auto-height.html
```diff
@@ -3,7 +3,21 @@
...
+        <ids-text font-size="20" type="h2">Card Contents</ids-text>
+      </div>
+    </ids-card>
+  </ids-layout-grid-cell>
+  <ids-layout-grid-cell>
+    <ids-card auto-height>
+      <div slot="card-header">
+        <ids-text font-size="20" type="h2">Card Title Two</ids-text>
+      </div>
       <div slot="card-content">
```
vivz753, line 21: Added another example to showcase `auto-height` card with or without a header

demos/ids-data-grid/auto-fit.html
```diff
@@ -0,0 +1,12 @@
...
+</ids-layout-grid>
+<ids-layout-grid auto="true">
+  <ids-layout-grid-cell>
+    <!-- Wrapping ids-data-grid inside of a parent container with a fixed height while using the auto-fit attribute will allow the grid to grow to the size of said fixed height -->
+    <div class="parent-container">
+      <ids-data-grid auto-fit id="data-grid-1" label="Books" row-height="medium">
+      </ids-data-grid>
+    </div>
+  </ids-layout-grid-cell>
+</ids-layout-grid>
```
vivz753, line 12: Added `auto-fit` example for ids-data-grid

demos/ids-list-view/example.html
```diff
@@ -2,14 +2,28 @@
...
+    <div class="parent-container border">
+        <ids-list-view virtual-scroll="false">
+          <template>
+            <ids-text font-size="16" type="h2">${productName}</ids-text>
+            <ids-text font-size="12" type="span">Count: ${units}</ids-text>
+            <ids-text font-size="12" type="span">Price: $ ${unitPrice}</ids-text>
+          </template>
+        </ids-list-view>
+    </div>
+  </ids-layout-grid-cell>
```
vivz753, line 17: Added an example of ids-list-view not being wrapped in anything

src/components/ids-card/ids-card.js
```diff
@@ -50,14 +50,16 @@ class IdsCard extends mix(IdsElement).with(IdsEventsMixin, IdsThemeMixin) {
...
+        </div>
+        <div class="ids-card-content ${this.overflow === 'hidden' ? 'overflow-hidden' : ''}" part="content">
+          <slot name="card-content"></slot>
+        </div>
       </div>
-      <div class="ids-card-content${this.overflow === 'hidden' ? ' overflow-hidden' : ''}" part="content">
-        <slot name="card-content"></slot>
-      </div>
-    </div>`;
+    `;
```
vivz753, line 62: Changed for readability

src/components/ids-card/ids-card.js
```diff
@@ -72,7 +74,7 @@ class IdsCard extends mix(IdsElement).with(IdsEventsMixin, IdsThemeMixin) {
     this.removeAttribute(attributes.AUTO_FIT);
   }
 
-  get autoFit() { return this.getAttribute(attributes.AUTO_FIT); }
+  get autoFit() { return IdsStringUtils.stringToBool(this.getAttribute(attributes.AUTO_FIT)); }
```
vivz753, line 77: Standardizing getters for boolean attributes

src/components/ids-card/ids-card.scss
```diff
@@ -45,10 +47,6 @@
     min-height: auto;
     height: auto;
 
-    .ids-card-header {
-      display: none;
-    }
-
```
vivz753, line 51: Remove CSS that hides header when `auto-height` is toggled

demos/ids-list-view/index.scss
```diff
@@ -0,0 +1,7 @@
+.parent-container {
+  height: 800px;
+}
+
+.border {
+  border: 1px solid gray;
```
tmcconechy, line 6: Maybe just use #B7B7BA here.. var(--ids-color-palette-slate-30)

tmcconechy (commented):
Really good - just one super minor comment. Wondering if we should even put a border around the listview for when its not a card (if that rule is possible?). So we can change the border in a theme.

Or does the auto-height option work so you can just wrap it in a card instead of the grey border in the example?

vivz753:
> Really good - just one super minor comment. Wondering if we should even put a border around the listview for when its not a card (if that rule is possible?). So we can change the border in a theme.
> 
> 
> 
> Or does the auto-height option work so you can just wrap it in a card instead of the grey border in the example?

The border is something I added in the example css, it does not have any border by default. Can be easily removed. Please confirm if this is what you want me to do, thx.

tmcconechy (commented):
@vivz753 can you also test the pages with auto-height="true" on them:

http://localhost:4300/ids-swipe-action
https://main.wc.design.infor.com/ids-swipe-action

Not 100% sure why on that example but the card is no longer showing. i guess was the last PR but maybe you know quick?

tmcconechy:
> The border is something I added in the example css, it does not have any border by default. Can be easily removed. Please confirm if this is what you want me to do, thx.

Yes, i got that i was wondering if either we make it the color in our palette tho. So sure, just do that for now`#B7B7BA`

src/components/ids-card/ids-card.js
tmcconechy: Just as a note we are changing these all to stringToBool due to web pack changes on that eventual PR.. and changes to the Utils (but this works for now)

vivz753: Should I not bother changing them on future PR's then is what you're saying?

tmcconechy: yes because it has a new pattern coming so they will be consistent (new mixin / util imported in different)

vivz753:
ids-swipe-action appears the same on my branch and the example, as well as the main branch aside the tiny grey line on the left

main
Screen Shot 2021-11-18 at 1 03 36 PM


419-autofit-components
Screen Shot 2021-11-18 at 1 01 20 PM

https://main.wc.design.infor.com/ids-swipe-action
Screen Shot 2021-11-18 at 1 02 27 PM

NOTE: ids-notification-banner test was timing out and failing as a result--changing the data set to a smaller one, `bikes.json` solved the issue, thanks to help from @jmacaluso711

tmcconechy:
@vivz753 I think the swipe example stopped working after https://github.com/infor-design/enterprise-wc/pull/421 - any idea if there is something we can update to make the card show? (i didnt look very deep as i thought you might know)

tmcconechy (approved):

tmcconechy:
FYI I figured out why swipe action doesnt look like a card anymore. It was simply missing the import in the example. Have this on my branch fixed...

👍🏻

jmacaluso711 (approved):

Issue #151, opened 2021-04-01
IdsTimePicker: Add new web component 


ghost:
Add the timepicker component https://main-enterprise.demo.design.infor.com/components/timepicker/list

**Phase 1 Features (MVP)**
- Configure to show hours, minutes, seconds (and period depending on locale info) -> default is everything except seconds
- Ability to set locale (using local mixin) -> this will format the field according
- Since dropdown isnt done can use an input for the fields
- Keyboard: Down arrow to open then tab in the popup and esc to close
- Input states (validation, required, enabled, disabled, dirty ect)
- Theme

**Future Features (Not Now)**
NA - Smaller component so can probably do it all

**Possible Internal Markup**
```js
 <ids-time-picker id="time-picker-1" disabled="false" label="Regular Time Picker" seconds="true">
```

**Suggested Dependencies**
- ids-trigger-field
- ids-locale-mixin
- ids-popup
- ids-input / ids-dropdown

**Settings**
- seconds
- locale
- language
- id/ automation id
- states settings

**Design**
- https://www.figma.com/file/ri2Knf3KchdfdzRAeds0Ab/IDS-Mobility-v4.6?node-id=1%3A5740

**Events**
- change/input

**Other**
Pass the following checklist ordered in priority (some may be left for later) https://github.com/infor-design/enterprise-wc/blob/main/doc/CHECKLIST.md

tmcconechy:
dropdown is merged so all the trigger changes are in. so this is ready to fix now 👍🏻

PR #432, opened 2021-11-18
151 - Added IdsTimePicker component

clayinfor:

Adds the IdsTimePicker web component: 

```html
<ids-time-picker
  label="12-Hour Time Picker w/ seconds and period (am/pm)"
  placeholder="Enter your start time"
  format="hh:mm:ss a"
  value="10:30:00 AM"
  language="en"
  locale="en-US"
></ids-time-picker>

<ids-time-picker autoselect></ids-time-picker>
<ids-time-picker disabled></ids-time-picker>
<ids-time-picker readonly></ids-time-picker>
```

Closes https://github.com/infor-design/enterprise-wc/issues/151
See [Legacy Time Picker](https://main-enterprise.demo.design.infor.com/components/timepicker/list)
See [Figma Designs](https://www.figma.com/file/ri2Knf3KchdfdzRAeds0Ab/IDS-Mobility-v4.6?node-id=1%3A5740)

1. Checkout this branch: `151-time-picker-component`
2. Run: `nvm use`
3. Run: `npm install`
4. Run: `npm run test time-picker-func`
4. Run: `npm run start:watch`
5. Go to and test:  http://localhost:4300/ids-time-picker/
6. See [Legacy Time Picker](https://main-enterprise.demo.design.infor.com/components/timepicker/list)
7. See [Figma Designs](https://www.figma.com/file/ri2Knf3KchdfdzRAeds0Ab/IDS-Mobility-v4.6?node-id=1%3A5740)


tmcconechy (changes requested):
This looks overall really excellent just a couple things im thinking about ....

1. On the dialog the space between the fields looks uneven and there is a bigger gap on the right side. 
Screen Shot 2021-11-18 at 9 49 53 AM

Should look more like this https://main-enterprise.demo.design.infor.com/components/timepicker/example-index.html

2. Is the keyboard behavior in scope for this issue? As if i try down arrow to open it doesnt focus the dialog and i cant use the keyboard to select. Can also try keyboard to see the behavior on https://main-enterprise.demo.design.infor.com/components/timepicker/example-index.html

3. For organization could we take the `locale` and `language` off so we can set it on the ids-container and see that it updates the format by default? I.E. should take the current locale by default in addition to "hard coding" one but i cant seem to test that on the examples. Could show them all using the set locale and one with a hard coded locale (with a label).

The test failure may be temporary. i synced with main as well with the button so we will see if that fixes anything....

tmcconechy (commented):
I was also wondering about the `timepicker` vs `time-picker` but i can see there is a `color-picker` so thats a good call.

tmcconechy:
Checked through again and see its not all done but I found one more when at a smaller breakpoint the dropdowns would look nicer centered:

Screen Shot 2021-11-19 at 10 00 42 AM

src/components/ids-time-picker/ids-time-picker.js
```diff
@@ -0,0 +1,400 @@
...
+  /**
+   * @see IdsElement.getAttribute()
+   * @override
+   * @param {string} name the attribute's name
+   * @returns {string} the attribute's value
+   */
+  getAttribute(name) {
+    const value = super.getAttribute(name);
+    return value === 'false' ? false : value;
+  }
```
vivz753, line 78: Nice and clean. I like it. Didn't even need to use the stringToBool()

src/components/ids-time-picker/ids-time-picker.js
```diff
@@ -0,0 +1,400 @@
...
+    }
+  }
+
+  get value() { return this.getAttribute('value') || ''; }
+
+  get autoselect() { return this.hasAttribute(attributes.AUTOSELECT); }
+
+  set disabled(value) {
+    value = value === '' ? true : value;
+    const disabled = value ?? (value !== 'false' && value !== false);
```
vivz753, line 123: Could probably use the stringToBool() function here, if I'm reading your logic correctly.. You're just checking for the presence of the attribute, and returning true so long as it's not `'false'` or `false`, correct?

src/components/ids-time-picker/ids-time-picker.js
```diff
@@ -0,0 +1,400 @@
...
+
+    this.setAttribute(attributes.DISABLED, disabled);
```
vivz753, line 125: Do values have to be casted to strings before setting the attributes? 

Nvm, they don't have to be, [source](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute#example)
```
value
A DOMString containing the value to assign to the attribute. Any non-string value specified is converted automatically into a string.
```

I ask this because I've seen `value.toString()` passed in for setAttribute in some components

src/components/ids-time-picker/ids-time-picker.js
vivz753: This probably wouldn't ever happen.. but if `undefined` is passed to this function, it would return `true`

src/components/ids-time-picker/ids-time-picker.js
```diff
@@ -0,0 +1,400 @@
...
+    this.elements.triggerField.disabled = disabled;
+    this.elements.triggerButton.disabled = disabled;
+    this.elements.input.disabled = disabled;
+  }
+
+  get disabled() { return this.getAttribute(attributes.DISABLED) ?? false; }
+
+  set readonly(value) {
+    value = value === '' ? true : value;
+    const readonly = value ?? (value !== 'false' && value !== false);
```
vivz753, line 135: Same for here. You sure you don't want to use the `stringToBool`? Maybe the function name should be changed to `castToBool` instead, since it technically checks bool types too. Then it wouldn't be misleading to use here

src/components/ids-trigger-field/ids-trigger-field.js
```diff
@@ -83,32 +88,41 @@ class IdsTriggerField extends IdsInput {
...
-          <ids-text label ${disabledAttribHtml}>${this.label}</ids-text>
-        </label>` : ''}
+        ${label}
         <div
           class="ids-trigger-field-content ${this.cssClass}"
           part="content"
-          ${this.readonly && ' readonly'}
-          ${this.disabled && ' disabled'}
+          ${attrs.readonly}
+          ${attrs.disabled}
```
vivz753, line 125: Nice refactoring

src/components/ids-trigger-field/ids-trigger-field.js
```diff
@@ -189,13 +203,11 @@ class IdsTriggerField extends IdsInput {
    * @param {boolean|string} value True of false depending if the trigger field is tabbable
    */
   set tabbable(value) {
-    const isTabbable = stringUtils.stringToBool(value);
-    /** @type {any} */
-    this.setAttribute(attributes.TABBABLE, value.toString());
+    this.setAttribute(attributes.TABBABLE, !!value);
```
vivz753, line 206: Not a fan of the stringToBool()? 

There's an issue here, though
`this.tabbable = 'false'` would set the attribute to `true`

src/components/ids-time-picker/ids-time-picker.js
```diff
@@ -0,0 +1,417 @@
...
+      attributes.AUTOUPDATE,
+      attributes.DISABLED,
+      attributes.LABEL,
+      attributes.READONLY,
+      attributes.PLACEHOLDER,
+      attributes.VALUE,
+    ];
+  }
+
+  get hasSeconds() { return this.format.toLowerCase().includes('ss'); }
```
jmacaluso711, line 101: Do we need jsdocs comments for these getters/setters?

src/components/ids-time-picker/ids-time-picker.js
```diff
@@ -0,0 +1,417 @@
...
+class IdsTimePicker extends mix(IdsElement).with(
+    IdsEventsMixin,
+    IdsKeyboardMixin,
+    IdsPopupOpenEventsMixin,
+    IdsThemeMixin
+  ) {
+  constructor() {
+    super();
+
+    this.elements = {
```
clayinfor, line 57: Adding `this.elements` was helpful because it provided a single-point of reference for adding mouse and keyboard event handlers, and for working with these elements in testing.  It also allowed me to target and affect elements within other components, such as the label inside of the `IdsTriggerField` component.  

Also bagging them within `this.elements`, instead of having a getter for each one reduced class-method pollution.

The only catch is, I needed to call `super()` to get `IdsElement` to render `template()` before `querySelector()` would actually return any of the elements.  One work around to this would be to `this.elements` without the `connectedCallback()` method, but, in opinion, the `constructor()` is the most-appropriate place to have this kind of class-variable defining/initialization.

src/components/ids-time-picker/ids-time-picker.d.ts
```diff
@@ -0,0 +1,91 @@
+// Ids is a JavaScript project, but we define TypeScript declarations so we can
+// confirm our code is type safe, and to support TypeScript users.
+
+import { IdsElement } from '../../core';
+
+export default class IdsTimePicker extends IdsElement {
+  /** Set the time format to indicate which dropdowns to render (Default: "hh:mm:ss a") */
+  format?: string | 'hh:mm:ss a';
```
clayinfor, line 8: Here, I'm making the attribute's default-value ('hh:mm:ss a') explicit.

src/components/ids-time-picker/ids-time-picker.js
```diff
@@ -0,0 +1,417 @@
...
+  /**
+   * @see IdsElement.getAttribute()
+   * @override
+   * @param {string} name the attribute's name
+   * @returns {string} the attribute's value
+   */
+  getAttribute(name) {
+    const value = super.getAttribute(name);
+    return value === 'false' ? false : value;
+  }
```
clayinfor, line 81: Overriding `getAttributes()` here gives this element a chance to dictate how it behaves rather than being dictated to from the top-down via inherited-behavior…

As Vivian mentioned, this could be a good place to centralize doing things with `stringToBool()` and other helper/utility functions.  I also considered overriding `setAttributes()`, and removing attributes such as `disabled` whenever it was set to false, but Daniel’s comment about Angular made me rethink this approach.

src/components/ids-time-picker/ids-time-picker.js
```diff
@@ -0,0 +1,417 @@
...
+    `;
+  }
+
+  /**
+   * Create the Template for the contents
+   * @returns {string} The template
+   */
+  template() {
+    const options = this.options;
+    const hours = this.dropdown({ id: 'hours', label: 'Hours', options: options.hours });
```
clayinfor, line 209: This is where I started to use to [tagged-functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates), but I decided to revert to a regular function, because I couldn't help but feel I was trying to shoehorn tagged functions into this code as pattern.  The main take away I wanted from tagged-functions was to have my IDE/code-editor help me to parse the HTML, but I wasn't getting that with the taggd-functions approach, so I used a regular, which was still helpful in building the final string template-literal for dropdowns.

src/components/ids-trigger-field/ids-trigger-field.js
```diff
@@ -83,32 +88,41 @@ class IdsTriggerField extends IdsInput {
    * @returns {string} The template
    */
   template() {
-    const disabledAttribHtml = this.hasAttribute(attributes.DISABLED)
-      ? ' disabled'
-      : '';
+    const attrs = {
```
clayinfor, line 91: This pattern could make it easier to add attributes to template literals. 

The class attribute could be handled as follows:
```javascript
const classes = [ attrs.disabled, attrs.readonly, attrs.required ].join('');
```

```javascript
const template = `<article class="${classes}">Article text</article>`;
```

src/components/ids-time-picker/ids-time-picker.scss
```diff
@@ -0,0 +1,151 @@
...
+    justify-content: center;
+    height: 100%;
+    padding: 0;
+  }
+}
+
+ids-dropdown#minutes,
+ids-dropdown#seconds,
+ids-dropdown#period {
+  @include ml-16();
```
clayinfor, line 44: replace this with `margin-inline-start` attribute

src/components/ids-time-picker/ids-time-picker.scss
```diff
@@ -0,0 +1,151 @@
...
+}
+
+ids-dropdown#minutes,
+ids-dropdown#seconds {
+  &::part(input) {
+    position: relative;
+
+    &::before {
+      color: var(--ids-color-font-muted, #606066);
+      content: ':';
```
clayinfor, line 54: for localization, we will need to have this in the template (i.e. `<span>:</span>`

src/components/ids-time-picker/ids-time-picker.js
EdwardCoyle: I like this for when there are many references to internal elements.  In a lot of places, we use readonly getters for these references, which changes the behavior a bit (on-demand reference as opposed to one that's cached).  This can be good if the ref isn't needed often, but if it is used often, maybe better to do it this way.

It's also true that when you have many of those readonly getters, readability decreases quite a bit.  If we agreed to standardize against a single object ref for storing those, I wouldn't be opposed.

src/components/ids-time-picker/ids-time-picker.js
EdwardCoyle: I like the idea of setting up attributes to automatically respond to their assigned types so we can avoid that bulky type checking everywhere.  Per our discussion on the call yesterday, I think we should consider targeting these more appropriately somehow (for example, not every attribute in this component is a boolean attribute, like 'label', 'placeholder', or 'value', and should respond to `'false'` in this manner).  Other than that, this looks good to me 👍

src/components/ids-time-picker/ids-time-picker.js
EdwardCoyle: I ran into a similar problem in the Toolbar component recently, where it made more sense to just offload to another method that returns a string.  Template literals are fine where appropriate but I don't think it's a problem to use regular methods.

My only comment might be to consider if the `.dropdown` method can be made private

src/components/ids-trigger-field/ids-trigger-field.js
EdwardCoyle: If dealing strictly with classes, this could definitely help.  Since `disabled` and `readonly` have to be handled by attributes here, I think this works well 👍

src/components/ids-time-picker/ids-time-picker.scss
EdwardCoyle: I raised infor-design/design-system#478 for eventually getting these in Tailwind-style mixin format

src/components/ids-trigger-field/ids-trigger-field.js
```diff
@@ -50,8 +56,7 @@ class IdsTriggerField extends IdsInput {
     this.#attachEventHandlers();
     super.connectedCallback();
 
-    const labelEl = this.container.querySelector('label');
-    this.onEvent('click.label', labelEl, () => {
+    this.onEvent('click.label', this.elements.label, () => {
```
EdwardCoyle, line 59: I'm wondering if this is necessary?  I can see it's focusing the IdsInput's label, but the IdsInput already handles that behavior as seen on http://localhost:4300/ids-input

src/components/ids-time-picker/ids-time-picker.js
```diff
@@ -0,0 +1,417 @@
...
+import IdsTriggerField from '../ids-trigger-field/ids-trigger-field';
+
+import styles from './ids-time-picker.scss';
+
+const range = (start, stop, step = 1) => (
+  start > stop ? [] : [start, ...range(start + Math.abs(step), stop, step)]
+);
+
+const TIME = {
+  TWELVE: range(1, 12),
```
EdwardCoyle, line 31: Is there a way we can generate these ranges using the `Date` object or some other API, so we don't have to assume manually?  I'm guessing you have to generate here because of the step feature.

Also worth noting that these will also be dependent on our Locale system at some point.  If the range can be driven by the format, we could potentially pull this information from the "culture" files that define each Locale's date format, day periods, whether to use 12 vs 24 hours, etc.  For example, [this info lives here for `en-US` locale](https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-locale/cultures/es-US.js#L42-L45).

I'd be OK with punting on this for now but we should raise another issue if we do, maybe for a Phase 2

src/components/ids-time-picker/ids-time-picker.js
```diff
@@ -0,0 +1,417 @@
...
+    ];
+  }
+
+  get hasSeconds() { return this.format.toLowerCase().includes('ss'); }
+
+  get hasPeriod() { return this.is12Hours && this.format.toLowerCase().includes(' a'); }
+
+  get is12Hours() { return this.format.includes('hh'); }
+
+  get is24Hours() { return this.format.includes('HH') || !this.hasPeriod; }
```
EdwardCoyle, line 107: I think in the 4.x components, some of these getters for detection of specific parts of the date format were public out of necessity (before we could hide them via `#privateMethods()`), but I'm not sure that's necessary for the Web Component.

src/components/ids-time-picker/ids-time-picker.js
```diff
@@ -0,0 +1,417 @@
...
+  #attachEventHandlers() {
+    const {
+      dropdowns,
+      input,
+      triggerButton,
+      setTimeButton,
+    } = this.elements;
+
+    if (this.autoupdate) {
+      this.onEvent('change', dropdowns.hours, (e) => { this.setTimeOnField({ hours: e.detail.value }); });
```
EdwardCoyle, line 374: For the sake of limiting the amount of event listeners tied to each element, would it be possible to have only one here that delegates down to all Dropdowns in the picker, and handles each accordingly?

For context, we've become aware recently that we're generating a huge amount of event listeners at the individual component level, and we are actively seeking out ways of cutting down the sheer number of them.

src/components/ids-time-picker/ids-time-picker.js
```diff
@@ -0,0 +1,417 @@
...
+      this.onEvent('change', dropdowns.minutes, (e) => { this.setTimeOnField({ minutes: e.detail.value }); });
+      this.onEvent('change', dropdowns.seconds, (e) => { this.setTimeOnField({ seconds: e.detail.value }); });
+      this.onEvent('change', dropdowns.period, (e) => { this.setTimeOnField({ period: e.detail.value }); });
+    }
+
+    // using on mouseup, because on click interferes with on Enter
+    this.onEvent('mouseup', setTimeButton, () => {
```
EdwardCoyle, line 381: Same deal with events here - maybe just have one `mouseup` handler and track which element is being dealt with via `e.target`

EdwardCoyle (changes requested):
@clayinfor This is looking good so far! I appreciate you looking into some of our current patterns and trying to improve our approaches 👍 

I responded to some of your initial comments above, and also added a few new ones below.   Some we probably want to address now but one or two could be pushed to a Phase 2 issue.

Also before we accept this PR, we'll need to merge/rebase against `main` FYI.

src/components/ids-time-picker/ids-time-picker.js
EdwardCoyle: Also as far as detecting the number of hours goes, you probably could get away with only detecting a single `H` or `h`.  Eventually we need to support leading zeroes in the time part output, which is signified by `HH` or `hh` respectively.

src/components/ids-time-picker/ids-time-picker.d.ts
EdwardCoyle: Having defaults is fine but keep in mind the default configuration for IDS Timepickers doesn't contain seconds (looks like `h:mm a`)

src/components/ids-time-picker/ids-time-picker.js
```diff
@@ -0,0 +1,417 @@
...
+  get format() { return this.getAttribute(attributes.FORMAT) ?? 'hh:mm:ss a'; }
+
+  set value(value) {
+    if (!this.disabled && !this.readonly) {
+      this.setAttribute('value', value);
+      this.elements.input.value = value;
+    }
+  }
+
+  get value() { return this.getAttribute('value') || ''; }
```
andriivarhanov, line 122: please use `attributes.VALUE` here

src/components/ids-time-picker/ids-time-picker.js
```diff
@@ -0,0 +1,417 @@
...
+
+  set autoselect(value) { this.setAttribute(attributes.AUTOSELECT, !!value); }
+
+  get autoselect() { return this.hasAttribute(attributes.AUTOSELECT); }
+
+  set autoupdate(value) { this.setAttribute(attributes.AUTOUPDATE, !!value); }
```
andriivarhanov, line 128: `autoupdate` can't be set/unset dynamically, `document.querySelector('ids-time-picker').autoupdate = false` - same behavior

src/components/ids-time-picker/ids-time-picker.js
```diff
@@ -0,0 +1,417 @@
...
+  set value(value) {
+    if (!this.disabled && !this.readonly) {
+      this.setAttribute('value', value);
+      this.elements.input.value = value;
+    }
+  }
+
+  get value() { return this.getAttribute('value') || ''; }
+
+  set autoselect(value) { this.setAttribute(attributes.AUTOSELECT, !!value); }
```
andriivarhanov, line 124: `autoselect` can't be set/unset dynamically, `document.querySelector('ids-time-picker').autoselect = false` - same behavior

src/components/ids-time-picker/README.MD
```diff
@@ -0,0 +1,155 @@
...
+- `input` allows you to further style the input element
+- `popup` allows you to further style the timepicker's popup window
+
+## States and Variations
+
+- `hasSeconds` {boolean} Does the timepicker format include seconds ("ss")?
+- `hasPeriod` {boolean} Does the timepicker format include the am/pm period (" a")?
+- `is12Hours` {boolean} Is the timepicker using a 12-Hour format ("hh")?
+- `is24Hours` {boolean} Is the timepicker using a 24-Hour format ("HH")?
+- `isOpen` {boolean} Is the timepicker's popup open?
```
vivz753, line 130:
```suggestion
- `isOpen` {boolean} Is the timepicker's popup open?

```

src/components/ids-time-picker/ids-time-picker.js
vivz753: I did a similar thing in `ids-slider` where I querySelected a bunch of elements and saved them into local variables, but I did it in connectedCallback() because some of my elements are conditionally rendered depending on certain attributes. Am wondering if there is any difference in performance when doing it in constructor() vs doing it in connectedCallback() 🤔 Technically connectedCallback() will get called every time and element is added to the DOM, so in my case it would make sense to have it there, since some elements are conditionally rendered depending on what type of slider it is (double slider needs two buttons, single slider needs one button)

tmcconechy:
@clayinfor any progress on resolving the remaining things on this one? Thank You

src/components/ids-time-picker/ids-time-picker.js
```diff
@@ -0,0 +1,417 @@
...
+        >
+          <ids-input
+            type="text"
+            placeholder="${this.placeholder}"
+            value="${this.value}"
+            disabled="${this.disabled}"
+          >
+          </ids-input>
+          <ids-trigger-button>
+            <ids-text audible="true">Timepicker trigger</ids-text>
```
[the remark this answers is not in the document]
clayinfor: Thanks for the tip!  I attempted to do this, but got a lot of errors thrown and a broken page. See screenshot:
image

src/components/ids-time-picker/ids-time-picker.js
clayinfor: I've decided to move this.elements out of the constructor and into a getter for the sake of making it `readonly`... the component would break if someone changed it's structure.

src/components/ids-time-picker/ids-time-picker.js
clayinfor: Thanks for the tip!  When I try to make `this.dropdowns()` private `this.#dropdowns()`, I get the following error:
image

src/components/ids-time-picker/ids-time-picker.js
clayinfor: Added to the timepicker's TODO.md file.

src/components/ids-time-picker/ids-time-picker.js
clayinfor: I attempted to make these private (i.e `this.is24Hours` into `this#is24Hours()`), but got the following error:
image

clayinfor: Added to the timepicker's TODO.md file.

src/components/ids-time-picker/ids-time-picker.js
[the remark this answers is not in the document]
clayinfor: Added to the timepicker's TODO.md file.

src/components/ids-time-picker/ids-time-picker.js
clayinfor: Added to the timepicker's TODO.md file.

src/components/ids-time-picker/ids-time-picker.js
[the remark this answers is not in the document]
clayinfor: Never mind, I think those errors were being thrown by Chrome's ScreenReader plugin which I was using to test this.

src/components/ids-time-picker/ids-time-picker.d.ts
clayinfor: Couldn't get `this.locale.calendar().timeFormat` to return a value for me, so I've tabled and added this suggestion to the TODO.md file.
image

src/components/ids-time-picker/ids-time-picker.js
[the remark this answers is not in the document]
andriivarhanov: @clayinfor thanks, it's there, but it doesn't translate for some reason 
image

src/components/ids-time-picker/ids-time-picker.js
andriivarhanov: @clayinfor could you please use `attributes.VALUE` in value setter as well

src/components/ids-time-picker/ids-time-picker.js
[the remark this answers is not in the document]
clayinfor: hey @andriivarhanov , I actually saw that, but thought that's the way the output was supposed to be.  I also had issues using `this.locale.calendar()` as a default for the `get format()` that @tmcconechy suggested.  Now I'm starting to wonder if I may be using the `IdsLocaleMixin` incorrectly.

src/components/ids-layout-grid/ids-layout-grid.scss
```diff
@@ -19,7 +19,7 @@ $grid-cols-sm: 4;
   --grid-row-end: 0;
   --grid-spacing-unit: 8px;
   --grid-fill-color: #efeff0;
-  --grid-min-col-width: 100px;
+  --grid-min-col-width: 50px;
```
tmcconechy, line 22: his could also be the reason http://localhost:4300/ids-dropdown broke? any reason for this? cc @jmacaluso711

src/components/ids-time-picker/TODO.md
```diff
@@ -0,0 +1,24 @@
+# TimePicker TODO's
+
+- [ ] Update README.md
```
tmcconechy, line 3: Are all these still valid? At least README.md seems updated

src/components/ids-time-picker/ids-time-picker.d.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export default class IdsTimePicker extends IdsElement {
+  /** Set the time format to indicate which dropdowns to render (Default: "hh:mm:ss a") */
+  format?: string | 'hh:mm a';
+
+  /** Set the input-field value (a timestring) */
+  value?: string;
+
+  /** Multiples of this value are displayed as options in the minutes dropdown. */
+  ['minute-interval']?: number;
```
tmcconechy, line 14: Shouldnt this be `minuteInternal`?

src/components/ids-time-picker/ids-time-picker.d.ts
```diff
@@ -0,0 +1,96 @@
...
+  closeTimePopup(): void;
+
+  /** Open the timepicker's popup window */
+  openTimePopup(): void;
+
+  /** Close the timepicker's popup window */
+  toggleTimePopup(): void;
+
+  /** Set the input-field's timestring value. */
+  setTimeOnField(timeunits: {
```
tmcconechy, line 90: Would it be cleaner to use a `value` setting that did this?

src/components/ids-trigger-field/ids-trigger-field.d.ts
```diff
@@ -31,4 +31,9 @@ export default class IdsTriggerField extends HTMLElement {
 
   /** Fires as the trigger button is clicked. */
   on(event: 'triggerclicked', listener: (detail: IdsTriggerFieldEventDetail) => void): this;
+
+  /** Get a list of element dependencies for this component */
+  readonly elements: {
```
tmcconechy, line 36: Would anyone need this using the component?

tmcconechy (changes requested):
Im not sure about the trigger field changes. What was necessary there. I can see it broke two things.

1. is badly broken http://localhost:4300/ids-dropdown compare too https://main.wc.design.infor.com/ids-dropdown/ (i live site of whats on main)
2. tests point to some errors https://github.com/infor-design/enterprise-wc/runs/4521487299?check_suite_focus=true#step:9:3481
3. the size of the dropdown broke
Screen Shot 2021-12-14 at 10 45 34 AM

src/components/ids-trigger-field/ids-trigger-field.d.ts
clayinfor: good question... probably not... I'll take it out.

src/components/ids-time-picker/TODO.md
clayinfor: I removed this Update README line from TODO.md

src/components/ids-time-picker/ids-time-picker.d.ts
clayinfor: `value` is already used to set the input-field's value.

src/components/ids-trigger-field/ids-trigger-field.scss
```diff
@@ -2,11 +2,12 @@
 @import '../ids-input/ids-input';
 
 .ids-trigger-field {
-  @include flex();
+  @include inline-flex();
```
jmacaluso711, line 5: @tmcconechy @clayinfor Looks like this is causing the layout issue in the dropdown. I think it can be reverted back to flex without breaking the time-picker

clayinfor: yup that fixed it, thanks @jmacaluso711 !

src/components/ids-time-picker/ids-time-picker.js
clayinfor: hey @EdwardCoyle , I was able to group the dropdown event handlers, but I left these `mouseup` events alone because the trigger button has an icon in the middle of it that interferes with it's event bubbling and causes inconsistent `mouseup` targeting behavior.

src/components/ids-time-picker/ids-time-picker.d.ts
tmcconechy: Thanks - i'll follow up on this on after merging my locale changes

src/components/ids-time-picker/ids-time-picker.js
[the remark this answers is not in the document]
andriivarhanov: @clayinfor  yeah, you are right, it may be something with `IdsLocaleMixin`. To see a correct behaviour of translations and `this.locale.calendar()` just remove `language="pt"` from the example
demosids-time-pickerexample html

including @tmcconechy in the conversation since he is working on locale changes now

src/components/ids-time-picker/ids-time-picker.d.ts
andriivarhanov: @clayinfor it looks like you need include `languagechange` `localechange` events and get timeformat once calendar is loaded. Example of the events is here
https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-icon/ids-icon.js#L76-L94

src/components/ids-time-picker/ids-time-picker.scss
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  ids-popup {
+    &::part(popup) {
+      margin: 0;
+      padding: 0;
+    }
+  }
+
+  #dropdowns {
```
jmacaluso711, line 47: nit pick here: but we might want to stay away from using IDs for styling? Otherwise this looks great to me

jmacaluso711 (dismissed):

src/components/ids-time-picker/ids-time-picker.d.ts
tmcconechy: Can you explain the difference im not sure i understand then what setTimeOnField vs value does?

tmcconechy:
@clayinfor still some test failures on trigger fields here. Not sure 100% whats going on with that but didnt spend much time on it. The rest looks good though. And one question i added.

src/components/ids-time-picker/ids-time-picker.d.ts
clayinfor: Sure, @tmcconechy :  `this.value` will set the input-field directly... whereas `this.setTimeOnField()` will calculate what the input-field's value should be by using the current values in the dropdown-selections.  

I actually had a different name for `this.setTimeOnField()` (something like `this.updateTimestring()`, but saw that's the name the legacy-component used, so I stuck with that convention.

clayinfor: To elaborate, if someone did `this.value = "Hello World"`, the input-field would display `Hello World`... `this.setTimeOnField()` would calculate and set the input-field's value back to what it should correctly be based on current dropdown-selections.  So `this.value` is just standard input-field behavior that gives users a way to override the input-field value.

tmcconechy: ok I see. On the old version we were a bit limited in that we couldnt override the value setting as we didnt have a custom component. So not attached to it.

One similar change was on the dropdown i decided to make value() actually lookup the dropdown value and set it https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-dropdown/ids-dropdown.js#L183 so its somewhat similar...

What if you used value here but in the setter but allow it two ways..

```
comp.value = time: {hours: 10, minutes: 20: seconds: 30: period: 'AM}
// or
comp.value = '10:10:20 AM';
```

i.e if the value is an object with `time` convert it and set it and if its just a string set that direct in?

clayinfor: Ok, that looks good 🤓   What should I then do with `this.setTimeOnField()`?

tmcconechy: get rid of it in favor of passing the time info in to the value setter... It can go away in that case, right? 👍🏻

clayinfor: Right, ok cool 👍

tmcconechy (approved):

tmcconechy (changes requested):
there is a couple things on the percy tests that caught
- disabled color on trigger disabled label
- dropdown bottom padding
- something on search?

src/components/ids-trigger-field/ids-trigger-field.js
clayinfor: I tried removing this block of code, but the input doesn't autofocus once the label is clicked... so this block of code appears to be necessary.

clayinfor:
> there is a couple things on the percy tests that caught
> 
> * disabled color on trigger disabled label
> * dropdown bottom padding
> * something on search?

Thanks @tmcconechy! I was able to fix these... the remaining percy test-fails just require an approval/thumbs-up.

tmcconechy (approved):

deep7102 (approved):
Looks good 🍌 
Thanks! @clayinfor

PR #435, opened 2021-11-18
415 virtualscroll arrowkeys

### LINKED ISSUE #415: VirtualScroll: Arrow Keys
**Describe the bug**

The arrow keys don't work well with virtual scroll.

**To Reproduce**
1. Go to http://localhost:4300/ids-list-view
2. Tab into the list
3. Down arrow continuously until it "scrolls" and notice the arrow key functionality breaks and loses focus

vivz753:

BUG: when navigating down the list with arrow keys, virtual-scroll re-renders and replaces the innerHTML, losing focus and any state of tabindex=0 for focused item

SOLUTION: refocus after each re-render, using the virtual-scroll::scrollToIndex() function and save the focused item index in a private variable so to keep track

- Closes #415 
- This fixes the issue where focus is lost when virtual-scroll re-renders (happens on scrolling)
- Also went ahead and fixed the issue where selection highlight on ids-list-builder is lost when virtual-scroll re-renders
- Reduced the buffer size of virtual-scroll to boost performance (no longer lags when scrolling, and renders faster)

http://localhost:4300/ids-list-view

https://user-images.githubusercontent.com/22609861/143933352-ac864378-d4bf-4d02-b368-c5c2d567d23f.mp4

http://localhost:4300/ids-list-builder

https://user-images.githubusercontent.com/22609861/143933291-9fe42986-9c12-4356-b28c-7f2f09499392.mp4


EdwardCoyle (commented):
Looks great @vivz753!  This probably needs a merge/rebase of `main` before the tests pass (will defer to you on that), but will approve once that's done 👍

demos/ids-virtual-scroll/index.scss
```diff
@@ -2,18 +2,44 @@ html {
...
 
 .ids-data-grid {
-  border: 0;
-}
 
-ul {
-  list-style: none;
-  margin: 0;
-  padding: 4px 20px;
+  [part="style-wrapper"] {
```
tmcconechy, line 15: Any particular reason to change this to [part]? As i think we should avoid styling off the parts as that was intended for end-developer customization and overrides, although not 100% sure on this yet, but its currently the only way to style things from the outside if needed for customization, while inside can just use the css classes ect. It also could impact the standalone css. Maybe fine in this one place but we could end up removing all these in the future as customization requirements are refined.

I missed the call about this but something i would have mentioned...

vivz753: I added it as a wrap around just to make the class selectors more secure by enforcing stricter hierarchy. I can change it to a class name or slot if you prefer that better. Or would you rather have it removed from the css sheet altogether?

I brought up in the last meeting that the workaround for styling within the shadow root of a child component is difficult from the parent and results in unelegant hierarchy. For example, in order for a parent component which uses virtual-scroll as a child component to style the list items inside of virtual-scroll, it has to shove a style-wrapper inside of it like so

```html
<div class="ids-list-view">
  <ids-virtual-scroll>
    <div part="style-wrapper" slot="contents">
     </div>
  </ids-virtual-scroll>
</div>
```

And virtual-scroll has to be coded in a way to querySelect that style-wrapper which isn't necessarily guaranteed to be there all the time. So the part="style-wrapper" was my attempt to make sense/unify that problem (it was previously slot="contents" but I felt style-wrapper was more informative and specific wording)

tmcconechy: Ok, Yeah maybe we should follow up on this conversation. But the idea I had around parts was for external customization. Similar to this example https://component.kitchen/elix/customizing#styling-element-parts

So basically because the css is "locked" in the shadow root its really the only tool we have to make the end user customize the css if ever needed for some tweaking / custom theming / customer specific styling ect.

So i think at least the lots names in the `@part section` should be well defined on the structure and free of styling that needs overriding in the future. But it might be ok to add a special `:part` that we dont let the end user customize?

I think the default should be a css class (or direct attribute /id selector) if the style is IN the component. It if its in another component it would be better to use `::slotted`.

vivz753: I see your point. I'll re-work it a bit and see what I can do.

tmcconechy (commented):
The only issues i see are:

1. I think we might want to reintroduce https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/list_role
2. Forgive me if i forgot early discussing but on http://localhost:4300/ids-list-view - could we put it in a card and have it only extend to the bottom of the page? (not lower) I forget why we didnt do this before?

vivz753:
In a previous PR for autofit changes, it was requested to have ids-list-view automatically fit to its parent container. The example for ids-list-view here has a parent container of 800px. Can change it to 70vh to fit the page responsively, since that's what your request sounds like. 
Also, I separated having a card wrapper around it into its own dedicated example http://localhost:4300/ids-list-view/card.html

I felt it made more sense to show the ids-list-view component pure and on its own as the default example.

vivz753:
Added role="list" to immediate parent of the items and role="listitem" to the divs that are focused when navigating the list

tmcconechy:
Looking Good!

In the previous version we did present it as a card https://main-enterprise.demo.design.infor.com/components/listview/example-index.html so thats why im thinking about it that way. We generally want to just go for what looks and presents it the best not necessarily "pure". Also would remove the need for a "fake" border css but thats minor.

I like the height change and addition of the aria roles 👍🏻 

1. But i see a new issue in that there is an extra tabs stop. Maybe we need tabindex="-1" on the root element? To reproduce click on the LISTVIEW text and tab then tab again and notice you have two tab stops
1. And maybe chuck it in a card like the older example but i wont insist on that 👍🏻

vivz753:
I felt it was important to show that the ids-list-view isn't dependent on ids-card to work/display decent UI. Not directly related, but in quite a few examples, I've seen names of classes (that are existing components) wrap other components in their example, i.e. it's confusing to see virtual-scroll wrapped in a div with a class named ids-list-view or ids-data-grid, because it suggests the possibility of using just CSS to mimic the appearance of components which also suggests that maybe those components don't have important dedicated functionality encapsulated within them/didn't need to be developed in the first place. If our end users are developers, I think it makes sense to show them the simplest example first, then if they want, they can look at the secondary example files which show how they can interact with other components, in this case how ids-list-view interacts with ids-card.

The "fake" border can be changed to something else or removed entirely if you like. If you want it to resemble the way card looks, I can just round the corners and add a box shadow.

I added an extra tabindex="0" to the the direct parent of the list items because my test was throwing an error saying that the container with scroll needs to be focusable... I just removed it and my tests passed, but let's see if the pipeline throws that error.

tmcconechy:
Ok maybe do it without the border then if you insist.  But i rather prefer the card look as its the most common design placement if you look at the designs...

See all these are in cards https://github.com/infor-design/enterprise/issues/5616

vivz753:
Those look like examples of different types of widgets with a combo of diff components (pagination with status indicators). Why not just have dedicated examples in the card folder like you suggested there? Having a dedicated area and filename for those specific widget examples makes sense i.e. cards/example-widget-status-indicators cards/example-widget-group-action-bar

tmcconechy:
yeah ok. we can think about those later (if we need them all). Im not sure if they should go in card/listview but definately not related to virtual scroll. Probably future listview examples.

tmcconechy (dismissed):

tmcconechy:
Can just ignore `scrollable-region-focusable` in the axeTest 👍🏻

tmcconechy (approved):

deep7102:
Looks good! only one comment -not sure it related to this ticket or separate issue, initially can not focus by keyboard
http://localhost:4300/ids-list-view/card.html
- Refresh page and keep pressing `Tab` key, there is no tab stop for it to focus the list

vivz753:
I think it was agreed to ignore that accessibility test for now, regarding Tim's comment above with `scrollable-region-focusable` test

deep7102 (approved):

Issue #376, opened 2021-10-04
Clearable button has 2 different colors


vivz753:
http://localhost:4300/ids-trigger-field
http://localhost:4300/ids-search-field

- focus styles

<img width="250" alt="image" src="https://user-images.githubusercontent.com/38439203/135632454-42c9f17f-346c-4c8d-a788-745303630fda.png">

tmcconechy:
The focus state should be a white border? Is this a question? Or is it about fixing the focus state color?

EdwardCoyle:
@vivz753 this might be solvable with an attribute `color-variant="alternate"` on the trigger buttons.  We may need some mechanism to automatically apply it when the IdsSearchField exists within an IdsHeader -- currently all buttons are expected to have the `color-variant` attribute applied manually when placed inside an IdsHeader.

tmcconechy:
assigned to Jeff but waiting on his email to come through so assigning to me

PR #436, opened 2021-11-19
376 - Automatically add alternate color variant to button components …

jeffwa428:
…in header

Before, we manually added `color-variant='alternate'` to the components in `ids-header`
Added new mixin for header so that color-variant can be added automatically for the buttons in `ids-header`

Closes https://github.com/infor-design/enterprise-wc/issues/376

Go to http://localhost:4300/ids-search-field
Check if menu, search-clearable button, search-input has alternate color variant

### DIFF
--- a/demos/ids-search-field/example.html
+++ b/demos/ids-search-field/example.html
@@ -13,25 +13,28 @@
     <ids-search-field disabled label="Disabled Search Bar" value="Anti-virus Software"></ids-search-field>
   </ids-layout-grid-cell>
 </ids-layout-grid>
+
 <ids-layout-grid cols="3" gap="xl">
   <ids-layout-grid-cell>
     <ids-search-field readonly label="Read-only Search Bar" value="Cellphone"></ids-search-field>
   </ids-layout-grid-cell>
   <ids-layout-grid-cell>
-      <form>
-        <ids-search-field label="Search Wrapped in a Form tag" value="Cellphone"></ids-search-field>
-      </form>
+    <form>
+      <ids-search-field label="Search Wrapped in a Form tag" value="Cellphone"></ids-search-field>
+    </form>
   </ids-layout-grid-cell>
   <ids-layout-grid-cell>
-        <ids-search-field color-variant="alternate" label="Alternate Color Search Bar" value="Cellphone"></ids-search-field>
+    <ids-search-field color-variant="alternate" label="Alternate Color Search Bar" value="Cellphone"></ids-search-field>
+  </ids-layout-grid-cell>
 </ids-layout-grid>
+
 <ids-layout-grid cols="1">
   <ids-layout-grid-cell>
     <ids-text font-size="12" type="h1">Inside a Header:</ids-text>
     <ids-header>
       <ids-toolbar>
         <ids-toolbar-section>
-          <ids-button icon="menu" role="button" color-variant="alternate">
+          <ids-button icon="menu" role="button">
             <span slot="text" class="audible">Application Menu Trigger</span>
           </ids-button>
         </ids-toolbar-section>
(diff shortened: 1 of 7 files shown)


jmacaluso711 (dismissed):
Looks good to me 👍

EdwardCoyle (changes requested):
@jeffwa428 Apologies but I think the intent of my [comments on #376](https://github.com/infor-design/enterprise-wc/issues/376#issuecomment-934704723) wasn't clear.

Doing the detection by adding a mixin to the buttons to detect their parent isn't ideal, I think.  This makes the button component implicitly tied to Headers when they don't have to be.

I think what might be better is to do a check for child buttons in [`IdsHeader`'s `connectedCallback()` method](https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-header/ids-header.js#L42), and apply the color variant there.  There may also need to be a process for running that check whenever Header contents are updated.

For context, this is how our App Menu does the color variant delegation: https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-app-menu/ids-app-menu.js#L81-L84

jeffwa428:
Okay, I got it. I will figure it out.

tmcconechy (changes requested):
Also @jeffwa428 im. not seeing that the issue is fixed if i put a value in the header search and tab to the x it its not the correct styles:

Screen Shot 2021-11-29 at 12 52 48 PM

Style should be: https://main-enterprise.demo.design.infor.com/components/header/example-non-collapsible-searchfield.html

src/components/ids-header/ids-header.js
```diff
@@ -62,6 +63,17 @@ class IdsHeader extends mix(IdsElement).with(
     </div>`;
   }
 
+  #refreshVariants() {
```
tmcconechy, line 66:
```suggestion
  /**
   * Refresh the color variants on all elements
   * @private 
   */
  #refreshVariants() {
```

tmcconechy (changes requested):
@jeffwa428 two things more

1. The way you change the color for focus looks good on the "INSIDE A HEADER:" example but it made the normal behavior (for example on "Products") disappear. The styling is different for the default VS the variant... So need to target the css variant
2. Also noticed if i tab to the X and hit enter it doesn't clear. But space does work. Should probably try to fix this as well if you can?

jeffwa428:
@tmcconechy Yes, I noticed that after I updated styles so I pushed new commit for that fix.

tmcconechy (approved):

tmcconechy (dismissed):
Assuming all tests pass now. Looks good to me 👍🏻

tmcconechy (changes requested):
@jeffwa428 Seems like this broke a few things. Have a look at the tests https://percy.io/00aaa841/enterprise-wc/builds/14328013/changed/809229527?browser=chrome&browser_ids=20&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280

1. Compare localhost http://localhost:4300/ids-trigger-field/test-sizes.html to https://main.wc.design.infor.com/ids-trigger-field/test-sizes.html it looks like the buttons are hidden out of view now
2. Compare http://localhost:4300/ids-header to https://main.wc.design.infor.com/ids-header the variant on this one stopped working?

There are some additional diffs (14 on the percy check)

src/components/ids-trigger-field/ids-trigger-button.scss
```diff
@@ -9,6 +9,7 @@
   @include p-0();
   @include rounded-round();
   @include text-slate-60();
+  @include bg-opacity-70();
```
tmcconechy, line 12: Looks like you changed the opacity for all trigger fields. https://github.com/infor-design/enterprise-wc/pull/436/files#diff-b7b144d543e7048fc295b1720c332571f159db20738ecb0be1f17b1a8b616517R12

I think all these css changes need to be only for the color variant case or else everything that uses trigger field is effected.

For example the button on https://main.wc.design.infor.com/ids-dropdown/ is now changed

tmcconechy (changes requested):
@jeffwa428 looking at the change we need to better isolate the change so it only effects the search example

tmcconechy (approved):

Issue #437, opened 2021-11-19
IdsSwipeAction: When loading the page the card is scrolled over


tmcconechy:
**Describe the bug**
The Swipe Action Component has an issue when its initially loaded thats sometimes the button is swiped over on load and should not be.

**To Reproduce**

1. Go to http://localhost:4300/ids-swipe-action
2. Reload the page until you see the buttons on the first swipe action card are scrolled over

Try to fix.. So that it looks like this:

![Screen Shot 2021-11-19 at 2 14 11 PM](https://user-images.githubusercontent.com/814283/142678673-0b952f4d-f1c3-4584-aef3-97ab8d1a8957.png)

**Additional context**
We tried initially to use this hack https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-swipe-action/ids-swipe-action.js#L65-L67 because the way this is it needs to be scrolled to the middle.

PR #441, opened 2021-11-23
437 - Hide button in swipe action card on load

jeffwa428:

Fix issue of showing swipe button on initial load

Closes https://github.com/infor-design/enterprise-wc/issues/437

- Go to http://localhost:4300/ids-swipe-action
- You should not see swipe button in first card whenever you reload the page


src/components/ids-swipe-action/ids-swipe-action.js
```diff
@@ -63,7 +63,9 @@ class IdsSwipeAction extends mix(IdsElement).with(IdsEventsMixin, IdsThemeMixin)
     }
 
     if (this.leftButton && this.swipeType === 'reveal') {
-      this.container.scrollLeft = 85;
+      setTimeout(() => {
```
EdwardCoyle, line 66: We're trying to standardize these in favor of `requestAnimationFrame()`:

```suggestion
      requestAnimationFrame(() => {
```

EdwardCoyle (approved):
This definitely fixes the state problem (no more swipe action on the right), but I wonder if we could improve this further by not animating the swipe state at all?  Currently when the page loads, the left swipe action is animating out.

EdwardCoyle (changes requested):
Sorry meant to request changes before

jeffwa428:
> Sorry meant to request changes before

Okay, I will make this change.

src/components/ids-swipe-action/ids-swipe-action.js
jeffwa428: @EdwardCoyle 
I tried with requestAnimationFrame but it was same. What can I do for now?

tmcconechy: It seems to work with it for me? The only thing i see is sometimes i get an extra line on the border:

Screen Shot 2021-12-01 at 12 54 08 PM

jeffwa428: You don't see any swipe left animation on page load?

tmcconechy: I see it move across the page (which isnt ideal) but i see the same thing with setTimeout or requestAnimationFrame

jeffwa428: Yes, it's same. The purpose was to remove that animation on the page load.

tmcconechy:
@jeffwa428 I still see it animating when the page loads? Wonder if there is a way to prevent that? I think worth exploring
Also still seeing the border on the left of 1px. Maybe that part can be fixed by:

```
.ids-swipe-container {
  ...
   // remove this margin-left: -1;
}
```

tmcconechy (changes requested):
1. See if we can hide the animation initially (somehow)
2. Fix the double border
3. Eds suggestion of requestAnimationFrame

tmcconechy (approved):
@jeffwa428 i fixed this up the "border" issue. turns out it was missing the card import so it looks different now. I still see the animation but i guess there is no great way to fix that so approving for now

src/components/ids-swipe-action/ids-swipe-action.js
```diff
@@ -63,7 +63,14 @@ class IdsSwipeAction extends mix(IdsElement).with(IdsEventsMixin, IdsThemeMixin)
     }
 
     if (this.leftButton && this.swipeType === 'reveal') {
-      this.container.scrollLeft = 85;
+      this.container.style.visibility = 'hidden';
+      setTimeout(() => {
+        this.container.scrollLeft = 85;
+
+        setTimeout(() => {
```
tmcconechy, line 70: The functionality is working well. But can we use the renderLoop pattern:

https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-toast/ids-toast-message.js#L131-L150


Is one example see also http://localhost:4300/ids-render-loop

tmcconechy (dismissed):

tmcconechy (approved):

tmcconechy (dismissed):

src/components/ids-swipe-action/ids-swipe-action.js
```diff
@@ -63,7 +64,24 @@ class IdsSwipeAction extends mix(IdsElement).with(IdsEventsMixin, IdsThemeMixin)
     }
 
     if (this.leftButton && this.swipeType === 'reveal') {
-      this.container.scrollLeft = 85;
+      this.container.style.visibility = 'hidden';
+
+      const self = this;
+      const timeout = renderLoop.register(new IdsRenderLoopItem({
```
EdwardCoyle, line 70: The use of the Renderloop can be good for the callback below this one, since it needs staggering by 200ms.  For the ones that should kick off during the next paint cycle, Renderloop usage is overkill.  This can be replaced with a call to `requestAnimationFrame` and save some cycles:

```suggestion
      requestAnimationFrame(() => {
        this.container.scrollLeft = 85;
      });
```

EdwardCoyle (commented):
One minor nitpick/improvement and we can call this one good. Also, not marking "request changes" since I'll be out for a few days.

tmcconechy (approved):

PR #445, opened 2021-12-02
Have stringToNumber() return NaN, simplify stringToBool()

vivz753:

- **Refactors util functions as discussed in a previous meeting**


src/utils/ids-string-utils/ids-string-utils.js
```diff
@@ -34,18 +34,16 @@ export function stringToBool(val) {
   if (typeof val === 'string' && val.toLowerCase() === 'false') {
     return false;
   }
-  return val !== null && (val === true || (typeof val === 'string' && val !== 'false'));
```
vivz753, line 37: Redundant, these cases already get checked by the if statement above. Also no need to check booleans, as this is meant to only take strings as parameters

src/utils/ids-string-utils/ids-string-utils.js
```diff
@@ -34,18 +34,16 @@ export function stringToBool(val) {
...
 /**
- * Converts an attribute string into a number
+ * Converts an attribute string into a number, or returns NaN
  * @param {string|number|any} val string value from the component attribute
  * @returns {number} The return boolean
  */
 export function stringToNumber(val) {
-  // eslint-disable-next-line no-unsafe-optional-chaining
-  const v = val?.toString() * 1; // Converting String to Number
-  return !isNaN(v) ? v : 0; // eslint-disable-line
```
vivz753, line 48: Unnecessary casting, and as discussed, makes more sense to return NaN rather than 0--which could be considered valid--for invalid values like `null` or `undefined`

tmcconechy (approved):
Looks good. Thanks for this. I just updated the branches...

andriivarhanov (approved):
Looks good, the only thing I was concerned is if someone was expecting `0` behavior from `stringToNumber` in existing components. I've checked all components where it's used and they seems to be working correctly with `NaN`

tmcconechy:
I had the same concern @andriivarhanov i Do think there may be some cases where we want zero not NaN but everything passes and im not 100% sure what they are at this time. So looks good for now

PR #450, opened 2021-12-06
379 ids process indicator responsive

### LINKED ISSUE #379: IdsProcessIndicator: add mobile/responsive styling
- Next iterative step of development for this component is to add responsive styling

- https://main-enterprise.demo.design.infor.com/components/processindicator/example-labels.html
- Just like the above. Test by adjusting the window size--the component will adjust accordingly

N/A

**Additional context**
N/A

vivz753:

- Refactored ids-process-indicator for more modular hierarchy for CSS classes and HTML elements
- Add responsive styling (when xs, hide the details and labels, and only display the active label)

- Closes #379 

- http://localhost:4300/ids-process-indicator
- Adjust the width of the window size to see responsive styling


src/components/ids-process-indicator/ids-process-indicator.js
```diff
@@ -26,8 +26,24 @@ class IdsProcessIndicator extends mix(IdsElement).with(IdsThemeMixin) {
     super();
   }
 
+  #activeStepLabel;
+
   connectedCallback() {
-    this.#calculateProgressLine();
```
vivz753, line 30: No longer needed, as progress lines are now handled individually by each process-step

demos/ids-process-indicator/example.html
```diff
@@ -7,57 +7,70 @@
     <ids-process-indicator>
       <ids-process-step label="Process Started" status="done">
         <ids-text>Unreleased</ids-text>
+        <div>
```
tmcconechy, line 10: Since this is the public api. Maybe we should make a new simple component for this just so that we have the custom element? `ids-process-details` maybe something like that? Or maybe a named slot? But seems like a component perhaps

tmcconechy (changes requested):
Looks good 👍🏻 

1. Maybe consider making that div a slot or component as mentioned
2. Can we make it break a breakpoint step earlier as when you get close to the breakpoint the alignment is a little off

Screen Shot 2021-12-07 at 10 11 18 AM

3. Seems like it doesnt go all the way over to the right (bit of a gap on the right)
Screen Shot 2021-12-07 at 10 12 55 AM

4. can we add a bit of test coverage for the new (and previous) code?
`npx jest --coverage -- process-ind`

Screen Shot 2021-12-07 at 10 15 07 AM

demos/ids-process-indicator/example.html
vivz753: Yeah we can do that, will open a new issue for it

tmcconechy: Ok or i would consider it part of this because your adding the `div` in here.

vivz753:
2. The breakpoint can be changed to condense when screen-size is below md. I had it like that initially but since the example didn't condense until below sm and ALSO had the cut off issue, I just mimicked that. 
3. It accounts space for details on the last step, let me try removing that extra width.
4. Will do!

demos/ids-process-indicator/example.html
vivz753: So the div was added around ids-icon and ids-hyperlink because icon and hyperlink needed to be inline/styled on the same row, otherwise they were separating onto different lines. I was going to add in slots initially but I think it needs to be spec'd out a bit more.

Is this format of data going to be guaranteed every time in this order/structure?
```
<ids-text slot="detail"></ids-text>
<div slot="persons"><ids-icon></ids-icon><ids-hyperlink></ids-hyperlink><div>
<div slot="timestamp"></div>
```

tmcconechy: i think maybe something like `details` instead as its not always going to be about people's people/status. They could technically put any information there.

FYI original specs if that helps https://github.com/infor-design/enterprise/issues/3918

I called it details before https://github.com/infor-design/enterprise/pull/4636/files?authenticity_token=cEdIcOZP4Cyq8KLZ%2FVq6JIXpmMRXr9Ab3x1yADYnBd9ZsX5m2l4crYPzk2HRDsKKclFJ5PEkfrNY8rhgQMNX7w%3D%3D&file-filters%5B%5D=.html&file-filters%5B%5D=.js&file-filters%5B%5D=.png&file-filters%5B%5D=.scss#diff-3346d3f1b32cff9ef0617302a1923925d8f877a9c31b9ef8d4c6598faadede53R13

So maybe simply:
```
<div slot="details">
  <ids-icon></ids-icon><ids-hyperlink></ids-hyperlink>
  <ids-icon></ids-icon><ids-hyperlink></ids-hyperlink>
  <ids-icon></ids-icon><ids-hyperlink></ids-hyperlink>
<div>
```

tmcconechy:
With the coverage i think just try to get it from yellow to green as a goal? Trying to keep it all green so maybe one more test 👍🏻  ?

Then i think the only thing left from me is adding a slot name to the <div slot="details"> wrapper

vivz753:
Slot for details added!

Can't figure out how to get this test to cover more branches.... it's stuck in yellow even though it runs through all the branches
Screen Shot 2021-12-08 at 4 17 27 PM

The yellow highlighted code isn't the cause for branches being 66%... 
Screen Shot 2021-12-08 at 4 17 37 PM

tmcconechy:
Try doing a rebuild.
```
npm run build
npx jest --coverage -- process-i
```
Then rerun the tests. I think sometimes because of the livereload the code gets a bit funny so that should recompile cleanly all the code.

vivz753:
Same result ^ 

But I was able to solve the problem, and increase branch coverage from 66% to 83% by adding another example into the demos. No matter what I did in the functional tests, it wasn't counted for those code paths. I had to add an example in the demos that didn't have labels to cover that ternary path for the xs-header label (which prints the index of the node if the active node has no defined label)

It works, but kind of dislike that I had to add an extra example in the demos index.html for ids-process-indicator. I tried having a dedicated example in its own html file, but that wasn't doing anything for the coverage percentage either. Lmk your thoughts

tmcconechy:
I think as we get more e2e tests going we will need examples for them, just about thinking of how to break it up in purposeful ways. I didn't see what you added yet but if we made it something different or more purposeful that would be good, or simply to split the example in two simpler ones i think thats totally valid. So if thats the case its probably good.

demos/ids-process-indicator/example.html
```diff
@@ -6,88 +6,204 @@
...
         </ids-hyperlink>
-        <ids-hyperlink href="#">
+        <ids-text slot="detail" class="date">Mar 2, 2021 01:42 AM</ids-text>
+      </ids-process-step>
+    </ids-process-indicator>
+  </ids-layout-grid-cell>
+</ids-layout-grid>
+
+<ids-layout-grid cols="1">
+  <ids-text font-size="12" type="h1">Process Indicator Without Labels</ids-text>
```
vivz753, line 107: New example here was added for test coverage

tmcconechy: Ok dont mind it but you could add a new page for the test `no-labels.html` at least that way you dont need to see it as the main example as its probably not something we want to encourage

vivz753:
I agree it would be better to break into more specific dedicated examples.

I don't really understand how or why the coverage results react to example.html only though and not my cases in the func test file and other html demo files.

tmcconechy:
Did you try this... https://github.com/infor-design/enterprise-wc/blob/main/test/ids-process-indicator/ids-process-indicator-e2e-test.js#L1

```
describe('Ids Process Indicator e2e Tests', () => {
  const exampleUrl = 'http://localhost:4444/ids-process-indicator';

  it('should not have errors', async () => {
    await page.goto(exampleUrl, { waitUntil: ['domcontentloaded', 'networkidle0'] });
    await expect(page.title()).resolves.toMatch('IDS Process Indicator Component');
  });

  it('should pass Axe accessibility tests', async () => {
    await page.setBypassCSP(true);
    await page.goto(exampleUrl, { waitUntil: ['networkidle2', 'load'] });
    await expect(page).toPassAxeTests({ disabledRules: ['color-contrast'] });
  });
});

describe('Ids Process Indicator No Label Tests', () => {
  const exampleUrl = 'http://localhost:4444/ids-process-indicator/no-labels.html';

  it('your test', async () => {
  });
});
```

i.e a new test loading the new page...

demos/ids-process-indicator/example.html
vivz753: That's what I saying in the above comment. I did initially, but it wasn't getting picked up by the test coverage.

vivz753:
What's the difference between e2e and functional tests? So far I've only really seen them used for axe testing and loading the DOM. I know recently there's been changes allowing UI tests involving clicking/dragging... but not sure how this would apply here or what I would write in the blanks for that test.

Tests aren't running properly on my personal mac atm... will try again later on the work mac.

test/ids-process-indicator/ids-process-indicator-e2e-test.js
```diff
@@ -1,9 +1,11 @@
 describe('Ids Process Indicator e2e Tests', () => {
   const exampleUrl = 'http://localhost:4444/ids-process-indicator';
+  const emptyLabelExampleUrl = `${exampleUrl}/empty-label.html`;
 
   it('should not have errors', async () => {
     await page.goto(exampleUrl, { waitUntil: ['domcontentloaded', 'networkidle0'] });
     await expect(page.title()).resolves.toMatch('IDS Process Indicator Component');
+    await page.goto(emptyLabelExampleUrl, { waitUntil: ['domcontentloaded', 'networkidle0'] });
```
vivz753, line 8: Adding dedicated example here increased test coverage. Lmk if this is acceptable @tmcconechy

tmcconechy: Yeah think this is acceptable enough for now. I have some fixes coming that may help fix coverage up more in the future.

tmcconechy (dismissed):
Approving but there is a minor lint error to fix 👍🏻

tmcconechy (approved):

andriivarhanov (approved):

PR #462, opened 2021-12-15
301 - Webpack optimization and Related Changes

tmcconechy:
**Explain the details for making this change. What existing problem does the pull request solve?**

This PR completes some initial work to optimize the releases. This includes but not limited to:

- Removes old mixin pattern in favor of one the reduces internal duplication and errors in web pack
- Changes the head section mainly of each component because of this
- Removes most index.js and changes all the imports to be direct (except for the demo app)
- Makes the webpack config dynamic and module with async chunks
- Went through each component and fixed a few minor things (comments/spelling/small errors/formatting ect)
- Updates all deps (web pack ect) to ensure testing the latest
- Removes attributeChangeCallback 
- Increase some test coverage
- Refactor Locale to only be attached on the container to solve some of these issues
- Related docs and small fixes

Results:
- Smaller build size
- More modular
- Able to proceed with lazy loading / module federation and other features (future issues)

**Related github/jira issue (required)**:
Fixes #301
Fixes #460
Fixes #427 

**Steps necessary to review your pull request (required)**:
Sorry about the diff. The github UI does'nt handle it well. For this it is mostly repetitive so you can possibly just scan over the diff. Also i did find working from the bottom to the top on the file changed works better, if you close and check them off working up. But you can also skip that for this one.

- Make sure are on Node 16.x (general comment)
- `rm -rf node_modules && npm i`
- `npm run build:stats` to see the bundle output visually
- `npm run start`
- check some random pages you know about
- look at the new component top area with imports for knowledge
- all checks on this issue pass
- see also working NG PR https://github.com/infor-design/enterprise-wc-examples/pull/7 using a release from this

**Included in this Pull Request**:
- [x] An e2e or functional test for the bug or feature.

Issue #442, opened 2021-11-29
IdsListBuilder: update data list when items are modified--virtual-scroll forgets all changes each re-render


vivz753:
**Is your feature request related to a problem or use case? Please describe.**

In list-builder with virtual-scroll="true", any modified/moved items, or items that were added/deleted are completely forgotten about once user scrolls --> virtual-scroll re-renders the original data list.

**Describe the solution you'd like**

List builder needs to call a function to update data each time there is a modification done with the list-builder toolbar--also needs to scroll to index of selected item if it's out of range and being modified with Enter key, or any of the buttons on list-builder toolbar. 

EDIT: as discussed with Tim, disabling the virtual scroll within listbuilder altogether will mitigate this. Another problem however happens with resizing the window. When window is resized and listbuilder becomes shorter, user scrolls, and when window is resized and listbuilder becomes taller, there is whitespace for that difference in height. An easy solution would be to just re-render() every time that ResizeObserver fires

12/17/21 EDIT: now that sortable has migrated to its own mixin, we can have it fire 'swapped' event whenever a swap successfully executes AKA the order of items in data has changed -> have the new order of items pushed to the data and fire a re-render(). Now that ids-list-view has an option of `sortable` it is also affected by this issue, and data needs to be updated each swap(). Solving it there would propagate to ids-list-builder, feeding 2 birds with 1 worm

Issue #368, opened 2021-09-30
SwapList: New Component


tmcconechy:
Add a new ids-swap-list component.

**Possible Internal Markup**
```js
<ids-swap-list>
   <div slot="available" title="Available">
   </div>
   <div slot="selected" title="Selected">
   </div>
   <div slot="additional" title="full access"> (optional)
   </div>
</ids-swap-list>
```

**Phase 1 MVP**
- https://main-enterprise.demo.design.infor.com/components/swaplist/example-index.html
 and nothing else just a single basic version on the main example with 2 or 3 sections
- may need to add sortable to the draggable examples
- disabled items

**Suggested Dependencies**
- ids-listview (maybe)
- ids-card 
- ids-sortable (depends on #399)
- ids-button

**Design**
https://main-enterprise.demo.design.infor.com/components/swaplist/example-index.html

**Other**
Pass the following checklist ordered in priority (some may be left for later) https://github.com/infor-design/enterprise-wc/blob/main/doc/CHECKLIST.md

PR #464, opened 2021-12-16
399 - Abstract functionality from ids-list-view/builder into its own ids-sortable-mixin

### LINKED ISSUE #399: IdsSortable: New Mixin

- [ ] Implement the ["Arrange"](https://github.com/infor-design/enterprise/blob/main/src/components/arrange/arrange.js) behavior from 4.x as `IdsSortable` (mixin?)
- [ ] Should extract drag/drop/sort code from Listbuilder (implemented in #392) into standalone functionality that can be applied to any component
- [ ] Need to do some R&D on how this works best in a WebComponent structure (Mixin vs Component vs other)

**Places that use it or will use it**
- [ ] Listview (listbuilder)

**Future Discussion*
- [ ] Tabs
- [ ] Datagrid
- [ ] Swaplist

vivz753:

Abstracts the sortable "drag and drop" functionality from `ids-list-view` and `ids-list-builder` into its own `ids-sortable-mixin`

The purpose of refactoring this is to allow modularity and usage of "swapping" `ids-draggable` elements on the DOM for other situations besides a list, i.e. `ids-tabs`

Added support for horizontal dragging (in consideration of tabs) in addition to vertical dragging (default).

- Closes #399 

 - https://localhost:4300/ids-list-view
 - https://localhost:4300/ids-list-builder


demos/ids-list-view/example.html
```diff
@@ -5,7 +5,7 @@
 <ids-layout-grid cols="2" gap="xl">
   <ids-layout-grid-cell col-span="1">
     <div class="parent-container">
-        <ids-list-view item-height="76" virtual-scroll="true">
+        <ids-list-view item-height="76" virtual-scroll="true" sortable>
```
tmcconechy, line 8: Just for seperation of concerns/examples. And also as this isnt the most common use case. Maybe we can we make a new demo/ids-list-view/sortable.html example?

src/components/ids-list-view/ids-list-view.js
```diff
@@ -62,13 +63,14 @@ class IdsListView extends mix(IdsElement).with(IdsEventsMixin, IdsKeyboardMixin,
...
 
-  getAllDraggable() {
-    return this.draggable ? this.container.querySelectorAll('ids-draggable') : null;
-  }
-
   attachEventListeners() {
-    this.attachClickListeners();
     this.attachKeyboardListeners();
+    if (this.sortable) {
+      this.attachDragEventListeners(); // for dragging list items
```
tmcconechy, line 69: Can we do this in the connectedCallback inside the mixin code? That way less setup...

src/components/ids-list-view/ids-list-view.js
```diff
@@ -232,7 +238,7 @@ class IdsListView extends mix(IdsElement).with(IdsEventsMixin, IdsKeyboardMixin,
    * @returns {string} The html for this item
    */
   itemTemplate(item) {
-    return stringUtils.injectTemplate(this.defaultTemplate, item);
+    return IdsStringUtils.injectTemplate(this.defaultTemplate, item);
```
tmcconechy, line 241: Just FYI the exports are changing on https://github.com/infor-design/enterprise-wc/pull/462 back to the previous way. So when that gets merged we might have to change this again.

src/mixins/ids-sortable-mixin/README.md
```diff
@@ -0,0 +1,21 @@
...
+3. User releases mouse button/ends drag
+    - The placeholder is removed from the DOM
+
+For more info on dragging and dropping elements in a list, and what the coding logic in this mixin is based on, refer to [this article](https://htmldom.dev/drag-and-drop-element-in-a-list/)
+
+The usage of this mixin would be:
+
+1. Import the mixin and add it to the `mix` list
+1. Manually add the `sortable` attribute to your example, or have your component set this property automatically.
+1. Call `attachDragEventListeners()` as needed. Calling it once in `connectedCallback()` usually suffices, **but if your list is dynamic** (items get added/deleted/modified) **or the list dynamically renders** (i.e. `ids-virtual-scroll`), you would need to make sure each new `ids-draggable` item gets the drag event listeners attached after they get added to the DOM.
```
tmcconechy, line 20: Again as above we could try to do this in the connectedCallback of the mixin to eliminate this setup step

src/mixins/ids-sortable-mixin/README.md
```diff
@@ -0,0 +1,21 @@
...
+1. **[Optional]** Override `OnDragStart()`, `OnDrag()`, `OnDragEnd()`, `createPlaceholderNode()` accordingly if you need additional function calls to modify or style elements
```
tmcconechy, line 21: One question: Is there an event that fires out for the component consumer so they can set/save the sorted items somehow?

tmcconechy (changes requested):
Looks really great! Thanks @vivz753 

1. See inline suggestions
1. See tests/lint errors
1. on the example http://localhost:4300/ids-list-view - can we make it a http://localhost:4300/ids-list-view/sortable.html and:
  a. Can we make it have a drag icon (this could be added but the mixin perhaps)
  b. Can we make the background the page background on drag, i cant take a screen shot but fix the fact thats its transparent

For TODO Future or Now if you have an idea?
- keyboard support
- fire events on `sorted`

jmacaluso711 (changes requested):
Seeing a bunch of lint errors:
Screen Shot 2021-12-16 at 12 08 39 PM

src/components/ids-list-view/ids-list-view.js
vivz753: That's definitely possible, however I think leaving the programmer the option to call attachDragEventListeners() where they want makes more sense and is the modular way to go. What if the component needed to perform some logic check before attachDragEventListeners()? (i.e. making sure list items get rendered before attaching drag event listeners? I'm certain this could be an issue with scenarios with dynamic rendering like with virtual-scroll). Calling it inside the mixin's connectedCallback() would mean that component's super.connectedCallback() would bunch this setup call with all the connectedCallback() functions of other mixins + `ids-element`'s connectedCallback(), with no control or clarity over what gets ran first.

src/components/ids-list-view/ids-list-view.js
vivz753: Sorry, can you explain to me again why stringUtils is better than using IdsStringUtils?

src/mixins/ids-sortable-mixin/README.md
vivz753: Are you saying you want a 'sorted' signal fired after drag end? What do you mean set/save? Like return an array of the items if they wanted to export the list into some excel sheet or something? Or are you referring to the issue discussed previously about how virtual-scroll currently doesn't save the state when items are moved around?

vivz753:
1. See my comment responses
2. Fixed lint errors
3. Separated example into sortable.html
  a. Are you saying you want it to have the dots on the left like how it is in ids-list-builder? If so, I would disagree and say that should not be included in the mixin. The mixin was solely for abstracting functionality of sorting, and adding specific CSS styles to it would be more coupling... If I add that CSS into the mixin, it will appear in tabs too, since this mixin was intended to be used for tabs.... and I don't think you want that... ?
  b. I don't really understand this, please elaborate what you mean by "make the background the page background on drag"?

I don't know what type of keyboard support would go well with this since it's "drag and drop". Can you give examples?

src/components/ids-list-view/ids-list-view.js
tmcconechy: OK sounds good to me. I think the way to think about it is..

a. do you always need this setup when sortable is enabled? if so it would go in the mixin to reduce duplication
b. do you need to call it under some logic that could change in the component. if so it goes in the component

I thought it was a. But if you say it could be b. Then this seems good for now.

src/components/ids-list-view/ids-list-view.js
tmcconechy: Because of the changes we made on #301 soon this will not compile. We changed all the utils so that they are individual exported methods. This is intended to fix some duplication in the build output and assist with better code splitting. Also it will make it more consistent as it was done two ways before and now it cant be. (after thats merged soon)

See:
https://github.com/infor-design/enterprise-wc/pull/462/files#diff-2b47b7839d06c727e58038556e4b8e233084c484c929dd4b2de376ef0b5b737fL106

https://github.com/infor-design/enterprise-wc/pull/462/files#diff-da2443f86311737d564ac5748f4778e299669a56bff09e677f957f68d9a34b15R3

src/mixins/ids-sortable-mixin/README.md
tmcconechy: I think for now we could just emit a sorted event after the drag end  / sort is complete. Im not sure how they will save it yet maybe they need to change how they put the data in ect and not really our problem. So maybe just omit an event so the developer knows a sort finished and pass what we have (the list elements and any indexes ect) in the data.

demos/ids-list-view/sortable.js
```diff
@@ -0,0 +1,22 @@
...
+import './index.scss';
+
+// Example for populating the List View
+const listView = document.querySelectorAll('ids-list-view');
+
+// Do an ajax request and apply the data to the list
+const xmlhttp = new XMLHttpRequest();
+const url = '/data/events.json';
+
+xmlhttp.onreadystatechange = function onreadystatechange() {
```
danielortiz1982, line 12: This is great! However I would recommend using the `fetch()` instead of the `XMLHttpRequest ()` it's a better cleaner solution for the given task at hand. You will not need to check for status code and you do not need the `onreadystatechange`. It returns a promise and we can let the JS engine do it's thing in the background.

Fetch APi DOCS
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

tmcconechy: Good idea - maybe we should standardize on `fetch` i use `XMLHttpRequest` in some examples too and should try to change to fetch someday

tmcconechy:
@vivz753 

>  Are you saying you want it to have the dots on the left like how it is in ids-list-builder? If so, I would disagree and say that > should not be included in the mixin. The mixin was solely for abstracting functionality of sorting, and adding specific CSS 
> styles to it would be more coupling... If I add that CSS into the mixin, it will appear in tabs too, since this mixin was intended to be used for tabs.... and I don't think you want that..

Ok you have a good point on tabs. I just suggested this because the drag dots icon is the indicator that you can drag it. So we put the icon on to show the behavior is there. Maybe you can just add it to the sortable example in the template so its specific to list view and in that css? 

> I don't really understand this, please elaborate what you mean by "make the background the page background on drag"?

I was able to stop in onDragend and get a screen shot:
Screen Shot 2021-12-17 at 11 19 13 AM

Im talking about the fact that the list is transparent so it looks wierd when you drag it. But the color can change in dark mode/ contrast ect so i think what you could do is use the same background colors as container https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-container/ids-container.scss on the list items or come up with something so its not transparent text on text when moving?

> I don't know what type of keyboard support would go well with this since it's "drag and drop". Can you give examples?
We can add this as a TODO? Nut its something like
- tab to the icon
- hit enter/space to activate drag mode
- move with arrow keys

Definitely out of scope for now but something we will need in the future to be accessible. https://www.barrierbreak.com/can-drag-and-drop-be-made-accessible-yes-it-can-be-in-different-ways/

src/mixins/ids-sortable-mixin/README.md
vivz753: Let's be more specific about the signal being fired here. I think a better event name to fire would be 'swapped', because technically user can drag an item around, but unless they go far up/down enough, nothing actually changes. And we don't want to fire events esp when nothing changed. My suggestion is:

- fire 'swapped' when a swap() is successfully executed

Currently, the data list doesn't get updated. Only the DOM reflects the order correctly. Will have to open a new issue to implement this.

demos/ids-list-view/sortable.js
vivz753: This is code that I just copied from other examples so maybe it'd be better to have a separate PR that changes all `XMLHttpRequest()` to `fetch()` for all examples in the demo folder. I'll make a separate issue and assign that to myself.

src/components/ids-list-view/ids-list-view.js
vivz753: Thx. It'll probably cause a merge conflict for you. Can change it back so there's less trouble for you

tmcconechy: Cool, that should work but be aware i did merge it in. So you might need to tweak it as you merge main on. let me know if you need help with the merge or have a question

src/mixins/ids-sortable-mixin/README.md
tmcconechy: yes good idea. i wasnt 100% sure what i wanted i just know it needs an event. I think swapped is fine. And it would be good if the data gets updated then you can send a ref to that in the event. If you would like to punt to a new issue. please do.

vivz753:
Got it. Added the dragging dots to ids-list-view.scss and background-color of list items now match container background.

Am hoping this gets approved soon. Jon and I were planning to pair program and start the ids-swap-list but need this mixin. Thanks to all reviewing this

SIDE NOTE: Regarding keyboard accessibility, wouldn't it be better to just make up/down buttons that call swap()? Similar to the toolbar in listbuilder. I think that makes more sense to put those functions in buttons if we're trying to "emulate" drag and drop through keyboard. Technically not drag and drop, just give the ability to swap through keyboard

src/mixins/ids-sortable-mixin/README.md
vivz753: Just added it to the existing issue #442

tmcconechy (commented):
I only see one issue now worth solving is one you can see in the percy images

https://percy.io/00aaa841/enterprise-wc/builds/14642716/changed/827855238?browser=chrome&browser_ids=20&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280

But if i change the theme in the page it looks good? Any ideas why? It looks like maybe there is a slight delay on the color change? Worst case we can try 
```
await page.waitForTimeout(200);
```

Or we could add the background as a class that only is added when dragging?

vivz753:
I'm not sure why it was doing that for the Percy tests. John & I tested locally too and it was fine how it looked. I know for sure it happened because of the background-color I added to the list items per your request. I was able to solve it though by moving bg-white from the default css styles into [mode='light']

tmcconechy (dismissed):

jmacaluso711 (commented):
The draggable list-view item is transparent on first load. It's not until you switch themes does the bg render as white. However, I'd be for approving this and maybe we addressing the bg color issue in #368 ?

Screen Shot 2021-12-20 at 10 41 21 AM

@vivz753 @tmcconechy

tmcconechy:
Up to you guys if you want to do it later thats fine with me?

src/components/ids-list-view/ids-list-view.scss
```diff
@@ -4,6 +4,11 @@
   height: 100%;
   overflow: auto;
   scroll-behavior: smooth;
+  position: relative;
+
+  ids-draggable {
+    width: 100%;
+  }
```
jmacaluso711, line 11: @vivz753 -- If you put the bg-color on `ids-draggable[is-dragging]` it should take care of the issue

jmacaluso711 (changes requested):

src/components/ids-list-view/ids-list-view.scss
vivz753: is `[is-dragging]` a css selector for an event fired? 

I just added `background-color: white;` in `ids-draggable` and it worked as well. Let's see if this passes the percy test

jmacaluso711: Looks like it's an attr that gets added to ids-draggable when the dragging event is firing

vivz753:
No disruptive changes on the percy screenshots! Ready for approval @tmcconechy @jmacaluso711

jmacaluso711 (approved):
🍌

tmcconechy (approved):

Issue #265, opened 2021-07-20
IdsCard: Add selection feature


tmcconechy:
**Describe the Task**
Should add the single and multi select feature of https://github.com/infor-design/enterprise/pull/5434#pullrequestreview-710677151 to cards in this repo as well.

See https://main-enterprise.demo.design.infor.com/components/cards/example-single-select.html and https://main-enterprise.demo.design.infor.com/components/cards/example-multi-select.html

**Checklist/Info**

- [x] Add setting selection: 'multiple' / false or 'single' with getter setter ect and attribs
- [x] Add the functionality to the template to show the checkbox/add class ect
- [x] Can use the checkbox mixin for the checkbox or add a checkbox component to the template
- [x] Trigger an event `triggerEvent('selectionchanged'...`
- [x] Unit tests

PR #470, opened 2021-12-21
265 - Add card selection

jeffwa428:
Adds a selection mixin in order to make item/card seletable, both single/multiple selection.
Here, added selection-mixin to `ids-card` so that card can be selectable.
`ids-card/selection.html` for example.

Closes https://github.com/infor-design/enterprise-wc/issues/265

- Go to http://localhost:4300/ids-card/single-selection.html and http://localhost:4300/ids-card/multiple-selection.html
- Try single/multiple selection for the cards


src/components/ids-card/ids-card.js
```diff
@@ -42,16 +48,66 @@ export default class IdsCard extends Base {
...
+    }
+
+    return this;
+  }
+
+  #changeSelection(e) {
+    this.container.querySelector('ids-checkbox').setAttribute(attributes.CHECKED, this.cardSelected !== 'true');
+    this.setAttribute(attributes.CARD_SELECTED, this.cardSelected !== 'true');
+
+    this.triggerEvent('selectionchanged', this, {
```
tmcconechy, line 101: Can we add a test for this? Also could it somehow be put in the mixin code?

tmcconechy (changes requested):
This is a good start.. Just some edits see inline. 

Overview:

- lets call the settings selection and selected instead
- add test for the event firing
- separate examples
- when selecting the card moves 2px because the border is added
- css changes broke the layout of cards on http://localhost:4300/
- on the multiselect one the title is in the wrong place

Screen Shot 2021-12-21 at 5 03 01 PM

src/components/ids-process-indicator/ids-process-step/ids-process-step.scss
```diff
@@ -1,8 +1,10 @@
 @import '../../../core/ids-base';
```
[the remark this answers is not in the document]
jeffwa428: I got bunch of issues from linting. So I fixed that. I am not sure how this was pushed to master.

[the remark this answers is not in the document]
tmcconechy: Possibly you created the branch before we made a bunch of changes on another PR thats also been merged. I'll check this but it probably wasn't your fault 😄

tmcconechy (changes requested):
I took a look and see a couple more improvements we could make

1. The border looks a bit different stylistically (is it 3px now with some grey in it?) I compared with https://main-enterprise.demo.design.infor.com/components/cards/example-multi-select.html

Screen Shot 2021-12-27 at 12 02 44 PM

1. The checkbox on multiple-selection doesn't seem centered vertically

1. We could try to make the example look like https://main-enterprise.demo.design.infor.com/components/cards/example-multi-select.html with ids-text for the elements - i think we could use https://main.wc.design.infor.com/ids-summary-field/ but add a vertical option to put them next to each other. Im ok with leaving this as is and this can be an idea for later tho.

1. If not the previous one we could just remove the card-header and have no contents for now and put that in for a TODO for later. As i dont think that the cards will have headers in these scenarios.

3. The c

src/components/ids-card/ids-card.js
```diff
@@ -42,16 +48,66 @@ export default class IdsCard extends Base {
...
     `;
   }
 
+  /**
+   * Establish Internal Event Handlers
+   * @private
+   * @returns {object} The object for chaining.
+   */
+  #handleEvents() {
+    this.onEvent('click', this, (e) => {
```
danielortiz1982, line 72: This is a nice solution, however I would recommend adding this to it's own function. We want to keep the handleEvents clean without too much business logic.

src/components/ids-card/ids-card.js
```diff
@@ -42,16 +48,66 @@ export default class IdsCard extends Base {
...
+      if (this.selection === 'single') {
+        const cardElements = document.querySelectorAll('ids-card[selection="single"]');
+        for (const elem of cardElements) {
```
danielortiz1982, line 75: Just a preference thing for me, if you spread the array you can do a forEach and shorten this statement to return the setAttribute something like, `[... cardElements].forEach( (elem) => elem.setAttribute(attributes.SELECTED, false))`. This is just a preference of mine I think it makes for a cleaner pattern.

danielortiz1982 (changes requested):
Remove business logic from handleEvents, Add logic for `this.selection === 'single')` into it's own function.

tmcconechy (approved):

jmacaluso711 (approved):

PR #479, opened 2021-12-28
442 update data list view

### LINKED ISSUE #442: IdsListBuilder: update data list when items are modified--virtual-scroll forgets all changes each re-render

vivz753:

- Fixes incorrect/duplicate selection bug when adding items on ids-list-builder
- Order of items in data property are correctly reflected when swapping items on the DOM for ids-list-view/builder
- Order of items in data property are correctly reflected when swapping/editing/adding/deleting items on ids-list-builder

- Closes #442 


demos/ids-list-view/virtual-scroll.html
```diff
@@ -5,9 +5,8 @@
 
 <ids-layout-grid cols="2" gap="xl">
   <ids-layout-grid-cell col-span="1">
-    <ids-text font-size="12" type="h1">With Border:</ids-text>
     <div class="parent-container border">
-        <ids-list-view>
+        <ids-list-view virtual-scroll="true" item-height="76">
```
vivz753, line 9: This wasn't enabled already for some reason

demos/ids-list-builder/example.html
```diff
@@ -4,7 +4,7 @@
 
 <ids-layout-grid cols="3" gap="md">
   <ids-layout-grid-cell>
-    <ids-list-builder height="310px" virtual-scroll="false">
+    <ids-list-builder height="310px">
```
vivz753, line 7: IdsListBuilder does not support virtual-scroll

src/components/ids-list-view/ids-list-view.js
```diff
@@ -259,26 +283,29 @@ export default class IdsListView extends Base {
...
-      `);
-
-      this.virtualScrollContainer.itemHeight = itemHeight; // calls renderItems()
-
-      this.virtualScrollContainer.itemTemplate = this.listItemTemplateFunc();
-      this.virtualScrollContainer.data = this.data; // calls renderItems()
-
-      // give the first list-item a tabbable index on first render
-      this.container.querySelector('div[part="list-item"]').setAttribute('tabindex', '0');
+      requestAnimationFrame(() => {
```
vivz753, line 286: Needed to grab the boundingRect height for the container to calculate the amount of items to render

src/components/ids-virtual-scroll/ids-virtual-scroll.js
```diff
@@ -65,32 +65,34 @@ export default class IdsVirtualScroll extends Base {
   renderItems(allowZero) {
     if (!this.data || (!allowZero && this.data.length === 0)) return;
 
-    const startIndex = this.startIndex;
-    const endIndex = this.startIndex + this.visibleItemCount();
+    requestAnimationFrame(() => {
```
vivz753, line 68: Added here as well to fix it

tmcconechy (commented):
How do we test this @vivz753 ? Maybe it would be a really good one to add a jest test for.

I was also thinking about if there was a way to make this reusable in the datasource object. But not sure about that. This is probably fine for now.

tmcconechy (approved):

tmcconechy (approved):

jmacaluso711 (approved):

Issue #332, opened 2021-09-08
IdsInput: New Password Features


tmcconechy:
**Describe the bug**
Add password related features to IdsInput from https://main-enterprise.demo.design.infor.com/components/input/example-password.html

**Additions**
- Reveal (hide/show button to toggle password setting) (should be able to show initially or not)
- Caps Lock indicator 
- Show password type on ids-input main page

**Possible Internal Markup**
```js
<ids-input type="password" reveal="true></ids-input>
<ids-input type="password" caps-lock="true></ids-input>
```

**Suggested Dependencies**
- ids-input

**Design**
https://main-enterprise.demo.design.infor.com/components/input/example-password.html
https://main-enterprise.demo.design.infor.com/components/signin/example-index.html -> hit caps lock and type in the password field

PR #482, opened 2021-12-29
332 update password functionality

Jah25:

Change updates password type input fields to include a capslock indicator and allow for password visibility to be toggleable

Closes [#332](https://github.com/infor-design/enterprise-wc/issues/332)

1) run the changes locally
2) go to http://localhost:4300/ids-input
3) at the bottom of the page toggle the password visibility
4) turn on capslock and type in the password field, an icon should appear


src/components/ids-input/ids-input.js
```diff
@@ -370,6 +448,39 @@ export default class IdsInput extends Base {
...
+        this.passwordVisibilityHandler();
+      });
+    } else {
+      this.offEvent('click.showhidepassword', showHidePasswordElem);
+      this.input.type = this.type;
+      showHidePasswordElem?.remove();
+    }
+  }
+
+  passwordVisibilityHandler() {
```
andriivarhanov, line 470: please make the function private

andriivarhanov (changes requested):
looks good, just a couple of items to fix:
  - comparing to the design https://main-enterprise.demo.design.infor.com/components/input/example-password.html?theme=new&mode=dark :
    - show/hide button should be uppercase
    - show/hide button should not be focusable or have focusable styles
    - please check light/dark/contrast theme show/hide button color
- please add new attributes/properties to `ids-input.d.ts` file
- when caps lock is enabled and you toggle revealable text option - icon/button changes order
image
- please add comments to any new getters/setters and functions in ids-input

src/components/ids-input/ids-input.js
```diff
@@ -75,12 +76,14 @@ export default class IdsInput extends Base {
...
       attributes.READONLY,
+      attributes.REVEALABLE_TEXT,
       attributes.TEXT_ALIGN,
       attributes.TEXT_ELLIPSIS,
       attributes.TRIGGERFIELD,
       attributes.TYPE,
       attributes.VALUE,
-      attributes.VERSION
+      attributes.VERSION,
+      attributes.PASSWORD_VISIBLE
```
tmcconechy, line 86: Can we keep this in alpha order?

src/components/ids-input/ids-input.js
```diff
@@ -170,6 +182,54 @@ export default class IdsInput extends Base {
     );
   }
 
+  get revealableText() {
```
tmcconechy, line 185: Pls add some JsDoc comments - will probably need this for docu generation

src/components/ids-input/ids-input.js
```diff
@@ -316,6 +376,24 @@ export default class IdsInput extends Base {
     }
   }
 
+  #capsLockEventSetUp(value) {
+    const indicatorDiv = this.shadowRoot.querySelector('.inline-indicators');
+    let capsLockIndicator = indicatorDiv.querySelector('#caps-lock-indicator');
+    if (value) {
+      this.onEvent('keyup.capslock', this.input, (event) => {
```
tmcconechy, line 383: This could get called twice i think. Might need..
```
this.offEvent('keyup.capslock', this.input);
this.onEvent('keyup.capslock', this.input, ...
```

src/components/ids-input/ids-input.js
```diff
@@ -370,6 +448,39 @@ export default class IdsInput extends Base {
...
 
+  #togglePasswordEventSetUp(value) {
+    const showHidePasswordElem = this.shadowRoot.querySelector(`.show-hide-password`);
+    const inputContainer = this.shadowRoot.querySelector('.inline-indicators');
+    if (value) {
+      if (!showHidePasswordElem && this.revealableText && this.type === TYPES.password) {
+        inputContainer.innerHTML += `<ids-button class="show-hide-password">${this.passwordVisible ? 'Hide' : 'Show'}</ids-button>`;
+        this.input.type = `${this.passwordVisible ? 'text' : this.type}`;
+      }
+      this.onEvent('click.showhidepassword', showHidePasswordElem, () => {
```
tmcconechy, line 459: Same here in case called more than once. Also check all methods/setters/getters do have a JSdoc even if private

src/components/ids-input/ids-input.scss
```diff
@@ -487,3 +487,20 @@ $input-compact-height: $input-field-height-xs;
...
+  right: 8px;
+  margin-top: -15px;
+
+  .show-hide-password {
+    &::part(button) {
+      padding: 0;
+    }
+  }
+
+  #caps-lock-indicator {
```
tmcconechy, line 503: Can we use class vs ID selectors for the css?

Jah25:
@andriivarhanov shouldn't the show/hide input be focusable since it's and input and changes the password field from a non screen readable to a screen readable format?

src/core/ids-attributes.js
```diff
@@ -197,6 +199,8 @@ export const attributes = {
   RESIZABLE: 'resizable',
   RESIZE_ON_DRAG_END: 'resize-on-drag-end',
   REQUIRED: 'required',
+  REVEAL: 'reveal',
```
tmcconechy, line 202: I like calling it revealable-text but I think maybe we didnt need both options here. But about reveal-password instead?

tmcconechy (changes requested):
Looks good. Just some questions / comments inline like reveal vs revealable-password?

Also in the example can we move it to under the IDS INPUT - SIZES section. I think that would fill out the page a bit more instead of making it scroll.

src/core/ids-attributes.js
Jah25: so reveal being there is a mistake on my part I refactored the name of the attribute from reveal to revealable-text to be more specific but forgot I had added reveal in the first place and didn't want to alter a pre-existing attribute. So I can just pull reveal entirely.

as for revealable-text, I was trying to make it more distinct from password-visible so there wasn't confusion between the too but I'm pretty open to whatever naming convention you think makes sense

tmcconechy: ok yeah looked like a dup. I guess i like reveal-password as you are revealing the password not the text? But im not that fixed on that point. I do like either better than just reveal?

andriivarhanov:
@Jah25 I think it should be focusable, I just noticed it's not in the original implementation and added it to the list. 
Let's ask @tmcconechy should show/hide button be focusable?

tmcconechy:
@Jah25 and @andriivarhanov I think maybe we should work it into the new setting `tabbable` i.e. if thats on you can tab to it if not you cant. Same with caps lock button. It wasn't this way before but that was done before we did some accessibility and testing that ultimately you should be able to tab to it for the best accessibility.

Jah25:
> 

When I was looking into updating this I was able to find the tabbable attribute in the project but I couldn't find anything that was using it. Is it an in progress feature?

tmcconechy:
@Jah25 I do see it being used in a few places (lookup, colorpicker, toolbar, search field).. search: `attributes.TABBABLE`

Jah25:
> @Jah25 I do see it being used in a few places (lookup, colorpicker, toolbar, search field).. search: `attributes.TABBABLE`

okay it looks like this was an issue with my VScode search locally, I did the search nothing showed up but with a restart I'm seeing results.

tmcconechy:
I've been seeing a bug with it too - like when i have the changelog or bigger text files open it doesnt work

Jah25:
@tmcconechy I added the tabbing functionality but it wound up being more of a separate story because it impacted things other than the show/hide functionality, should I add an item for tracking purposes or just move on?

tmcconechy:
@Jah25 you can just move on as at this stage a second issue for tracking wouldnt be used for anything if its done now.

tmcconechy (approved):
Looks all good to me now @Jah25 Thanks!
I just reorganized the input page to distribute the columns better. Will need visual test approvals.

andriivarhanov (approved):

tmcconechy (approved):

Issue #422, opened 2021-11-12
IdsDatagrid/IdsListview: Add Paging


tmcconechy:
**Describe the Enhancement**

Add ids-pager to the IdsDatagrid/IdsListview. This will probably be a mixin (IdsPagingMixin)

- The setting is `pagination` possible options are: unset/none (default), 'standalone', 'client-side', and server side

Standalone: Just add the pager and do not link it to the grid. In the example just show the pager events firing the end developer will do the rest...
ClientSide: Do the paging on the dataset thats attached. Probably this logic goes into `IdsDataSource`
ServerSide: Do the paging on the dataset thats attached and fire the pager events through the datagrid. Show this in the example working (can mock the data with a timeout instead of real server side logic since we dont have one except static JSON)

**Additional context**
See https://main-enterprise.demo.design.infor.com/components/datagrid/example-paging.html
and https://main-enterprise.demo.design.infor.com/components/datagrid/example-paging-client-side.html in the previous version

PR #483, opened 2021-12-31
422 - Added IdsPagerMixin to IdsDataGrid

clayinfor:

Adds the IdsPagerMixin (and is inherited by IdsDataGrid web component): 

```html
<ids-data-grid
  id="data-grid-1"
  row-selection="multiple"
  label="Books"
  row-height="md"
  pagination="client-side"
  page-number="1"
  page-size="50"
>
</ids-data-grid>
```

Closes https://github.com/infor-design/enterprise-wc/issues/422
See [Legacy Datagrid with Paging](https://main-enterprise.demo.design.infor.com/components/datagrid/example-paging-client-side.html)

1. Checkout this branch: `422-data-grid-add-paging`
2. Run: `nvm use`
3. Run: `npm install`
4. Run: `npm run test data-grid-func`
4. Run: `npm run start:watch`
5. Go to and test:  http://localhost:4300/ids-data-grid/pagination-client-side.html
6. See [Legacy Datagrid with Paging](https://main-enterprise.demo.design.infor.com/components/datagrid/example-paging-client-side.html)


demos/ids-data-grid/pagination-server-side.js
```diff
@@ -0,0 +1,93 @@
...
+    dataGrid.pageTotal = data.length;
+  };
+
+  dataGrid.columns = columns;
+  dataGrid.pageTotal = data.length;
+  paginateData(1, dataGrid.pageSize);
+
+  dataGrid.pager.addEventListener('pagenumberchange', async (e) => {
+    await new Promise((resolve) => {
+      setTimeout(() => {
```
danielortiz1982, line 84: This is a nice solution, however it's somewhat of an anti pattern having a setTime inside a Promise. You can remove the the promise and SetTime and use just the async and await pattern. The data will have to moved into a flat json file. Doing a fetch on that static json will return a promise with the data, this should be enough to simulate a server response.

tmcconechy (changes requested):
Looks really great. Just a few things i noticed:

**http://localhost:4300/ids-data-grid/pagination-client-side.html**
- the text is a bit cut off. before we had this dynamically size (in case its like 1000 pages)

Screen Shot 2022-01-03 at 11 44 49 AM

- the dropdown on the page size isnt using the style with the arrow pointer. i.e. instead of:
Screen Shot 2022-01-03 at 11 44 57 AM
should be:
Screen Shot 2022-01-03 at 11 45 03 AM
- Also when changing the page size 
- we will have to do some something with the auto fit option (so it sizes to the bottom of the page with the pager) but it appears this option is not working at the moment to size to bottom of the entire page container
http://localhost:4300/ids-data-grid/auto-fit.html


**http://localhost:4300/ids-data-grid/pagination-standalone.html**
- can we make this fire the events in the console so you can see the pager is responding?

**http://localhost:4300/ids-data-grid/pagination-server-side.html**
- could start out on 25 or 10 page size so it is showing the behavior without changing page size

demos/ids-data-grid/pagination-server-side.js
tmcconechy: Also removing that may make the tests a bit easier (percy/e2e doesnt like these delays)

src/components/ids-data-grid/ids-data-grid.d.ts
```diff
@@ -72,6 +72,9 @@ export default class IdsDataGrid extends HTMLElement {
   /** Sizes the datagrid to the parent container */
   autoFit: boolean;
 
+  /** Enables the pager  */
+  pagination: 'standalone' | 'client-side' | 'none';
```
tmcconechy, line 76: server-side too?

src/components/ids-pager/README.md
```diff
@@ -8,9 +8,10 @@ The IDS Pager Component is used to provide an interface for browsing paginated d
 - Controlling the pagination or keeping track of what is browsed in a data set.
 - Presenting digestible portions of large data sets.
 
-## Sub-Components
-**`<ids-pager-section>`**: divides the different sections on a pager panel. It will automatically align itself to the left, middle, or right side of the component depending on how many sections are provided.
+## Slots
+`start` and `end` slots allow you to position content to the left or the right side of the pager panel. If no slot is provided, the child-contents of the pager component will be automatically inserted into the middle slot.
```
tmcconechy, line 12: I think i prefer having a named component instead of just an unnamed div? Unless this is the only way? Or what was the purpose of this change?

src/mixins/ids-pager-mixin/TODO.md
```diff
@@ -0,0 +1,9 @@
+# TODO on Ids Pager Mixin
+
+## Features
+
+- [ ] Add support for IdsBlockGrid
```
tmcconechy, line 5:
```suggestion
- [ ] Add support for IdsBlockGrid
- [ ] Add support for IdsListView
```

tmcconechy (changes requested):
A couple more things im wondering about...

src/components/ids-pager/README.md
clayinfor: I didn't really want to delete named sub-component, but there was some other logic that was in place just to properly position this sub-component and I felt it was a bit over-engineered.  You can see in the code-changes that I deleted that extra code. 

Another thing is, I think having this extra `ids-page-section` sub-component adds extra cognitive-load (i.e. need-to-know) to the end-users of the `ids-pager` component but with little added-value.  I replaced `ids-page-section` with `div` in some places for expediency, but an end-user could use something more semantic like the `section` element with a slot-name... or whatever they think is best for their use-case.

clayinfor: Here is some of what I was able to delete once `ids-pager-section` was dropped:

https://github.com/infor-design/enterprise-wc/pull/483/files#diff-7f1a191e2c6e884cf5c47ca81bc5b1122b15abaf97256c22d7a0578ec8bad783L16

https://github.com/infor-design/enterprise-wc/pull/483/files#diff-934358cb4efcfba5d5c677ef80b35bd7426ff2618d91d3cdbee6ec84e4e99517L225

https://github.com/infor-design/enterprise-wc/pull/483/files#diff-934358cb4efcfba5d5c677ef80b35bd7426ff2618d91d3cdbee6ec84e4e99517L277

tmcconechy: Ok i see - yeah i guess removing the need for the MutationObserver for a slot makes sense. Just wanted to check the justification. I guess it was just the raw div throwing me off. Maybe it could be an HTML `<section>` element? Just for further semantics.

clayinfor:
Hey @tmcconechy, I think this is finally good to go... it just needs an approvals on the Percy tests.

One thing, I wasn't able to get the tooltip wedge to appear on the page-size popup.  I tried to resolve it, but couldn't ,so I've decided to revisit... I think it has to do with the ids-popup-menu appearing in ids-pager's slot, as opposed to it's shadow root... this is tripping up the `getClosestRootNode()` helper function.

tmcconechy:
@clayinfor the only thing i do see is looks like in the percy images that the tree doesnt load and also some borders are missing on http://localhost:4300/ids-data-grid/list-style.html

But when i try the live page it seems fine. Maybe push some something to rerun?

demos/ids-data-grid/pagination-client-side.html
```diff
@@ -0,0 +1,19 @@
+{{> ../layouts/head.html }}
+
+<ids-layout-grid auto="true">
+  <ids-text font-size="12" type="h1">Data Grid - client-side</ids-text>
```
tmcconechy, line 4:
```suggestion
  <ids-text font-size="12" type="h1">Data Grid - Paging Client Side</ids-text>
```

tmcconechy (approved):

danielortiz1982 (approved):
This PR looks great @clayinfor good job replacing XMLHttpRequest with the fetch api! 👍

Issue #471, opened 2021-12-21
IdsTreemap: Not rendering in Firefox


jmacaluso711:
**Describe the bug**
`ids-treemap` works as expected in Chrome and Safari, however is not rendering in Firefox

**To Reproduce**

1. In Firefox go to http://localhost:4300/ids-treemap
2. see issue

**Expected behavior**
`ids-treemap` should render correctly in all browsers. 

**Version**

- ids-enterprise-wc: 0.0.0-beta.17

**Screenshots**
![Screen Shot 2021-12-21 at 4 09 21 PM](https://user-images.githubusercontent.com/1084735/146998063-c579a390-d9bb-437d-97c1-34b4ca48cddb.png)

**Platform**
 - Infor Application/Team Name: Infor Design
 - Device: All
 - OS Version: iOs
 - Browser Name: Firefox 
 - Browser Version: 91.0.2

**Additional context**
n/a

PR #490, opened 2022-01-18
471 - Fix ids-treemap loading in firefox

jeffwa428:
ids-treemap works as expected in Chrome and Safari, however is not rendering in Firefox. It was an issue of ResizeObserver in firefox so moved the watcher to the parent element.

Closes https://github.com/infor-design/enterprise-wc/issues/471

- go to http://localhost:4300/ids-treemap (in FF/Safari/Chrome)
- resize the page -> should work

### DIFF
--- a/src/components/ids-treemap/ids-treemap.js
+++ b/src/components/ids-treemap/ids-treemap.js
@@ -433,10 +433,10 @@ export default class IdsTreeMap extends Base {
           }
         }
 
-        requestAnimationFrame(() => resizeObserver.observe(this.container));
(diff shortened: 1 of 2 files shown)


tmcconechy (commented):
@jeffwa428 im not actually seeing this work in any browser? Chrome/FF/Safari it all does not render. And im not seeing the resizeObserver fire.

But i think it may be true this needs to be somehow fixed in other places that do it this way. For example on http://localhost:4300/ids-wizard when i resize to mobile it shows ... on the test but in firefox it doesnt

Issue #141, opened 2021-05-04
Misc: Pull values from ui.config


tmcconechy:
**Describe the bug**
We have some places where config can be pulled from https://github.com/infor-design/design-system/blob/main/design-tokens/ui.config.js#L199 we can use this for colors. Fonts and more.

**To Reproduce**
1. Pull the ui.config in for https://github.com/infor-design/enterprise-wc/blob/main/src/ids-text/ids-text.js#L16 via  js import from node_modules
1. Remove the xs, sm, md and use numbers
1. Document it

**Questions**
Should this be a part of IdsElement or standalone import.

**Additional context**

tmcconechy:
Rethinking this because a) it adds quite a bit of unused code for little gain b) is easily synced up otherwise

lucacolumbu:
@tmcconechy let's chat about this in our meeting tomorrow.

PR #491, opened 2022-01-18
Misc: Add New Design Tokens and Imports

tmcconechy:

Made some changes to the mixins and mixin generator as follows:
- update a release with the new design repo (and other minor dependencies)
- added border radius mixins for every side https://github.com/infor-design/design-system/pull/480/files#diff-e68a04380dc97e741fcbbcefc04735c3c32bd8e335c79596da77e1ae0b5587acR269
- changed `ml mr mx pl pr px` to use  inline-start/end so they work better with RTL
- changed the color names for alerts and updated them for each themes
- added an importable fontSize.js to use in the components which is used in a few places

Fixes https://github.com/infor-design/design-system/issues/478
Fixes https://github.com/infor-design/enterprise-wc/issues/141)

- make sure all tests pass (by waiting for the checks
- do a quick smoke test of your favorite components
- review changes in this PR and on https://github.com/infor-design/design-system/issues/478 


jmacaluso711 (dismissed):

src/components/ids-badge/ids-badge.scss
```diff
@@ -42,18 +42,19 @@
...
-    background-color: var(--ids-color-status-danger);
-    border-color: var(--ids-color-status-danger);
+    @include bg-alert-error();
+    @include border-alert-error();
   }
 
-  &[color='alert'] {
-    background-color: var(--ids-color-status-caution);
-    border-color: var(--ids-color-status-caution);
+  &[color='alert'],
```
andriivarhanov, line 49: it looks like `alert` and `warning` are the same colors, but if you add `<ids-badge color="alert"...` to the example you will see they are different, scss styles vs inline css styles, bg-alert-warning() is not the same as var(--ids-color-status-warning)
image

andriivarhanov (commented):
looks good, just a comment on alert vs warning color in `ids-badge`

src/components/ids-badge/ids-badge.scss
tmcconechy: Yeah the idea was that we only have warning (no yellow alerts) due to bad color contrast accessibility. But i wasnt sure about `alert` vs `warning`. Maybe we should keep a diffference.

I'll take another look.

tmcconechy: @andriivarhanov i fixed this issue. I decided we shouldn't have `alert` as an alert color option as its like double alert. So only `warning` will work now. If you pass in the `alert` it will use the no color option.

There is an issue with that one `--ids-color-status-caution` so it shouldnt be used and i cant fix it as that would break the production code. So should be avoided in favor of the mixins for alert colors.

This PR will need image approvals but should be good now

andriivarhanov: @tmcconechy makes sense, thank you

andriivarhanov (dismissed):

andriivarhanov (approved):

PR #506, opened 2022-01-24
5768 (EP) -  Implement Actionable Button Card

ericangeles:

Implements actionable button card in the Card Component.

Closes https://github.com/infor-design/enterprise/issues/5768

- Pull this branch, and run the app
- Go to http://localhost:4300/ids-card
- It will act as a button with the same behavior
- It should be clickable and it should show ripple effect the same as the button component
- Tabbing through the actionable button card will show box-shadow effect the same as the button component
- Test in different modes and in classic theme with different modes too

~~- [ ] An e2e or functional test for the bug or feature.~~
~~- [ ] A note to the change log.~~


demos/ids-card/example.html
```diff
@@ -20,4 +20,11 @@
       </div>
     </ids-card>
   </ids-layout-grid-cell>
+  <ids-layout-grid-cell>
+    <ids-card actionable="true">
+      <div slot="actionable-text">
```
tmcconechy, line 25: Could we use the same named slot as the default? `<div slot="card-content">` this would maybe make it more consistent

src/components/ids-card/ids-card.js
```diff
@@ -37,16 +37,17 @@ export default class IdsCard extends Base {
       ...super.attributes,
       attributes.AUTO_FIT,
       attributes.AUTO_HEIGHT,
+      attributes.ACTIONABLE,
```
tmcconechy, line 40: Move up (alphabetical)

src/components/ids-card/ids-card.js
```diff
@@ -164,6 +193,21 @@ export default class IdsCard extends Base {
...
   get autoHeight() { return this.getAttribute(attributes.AUTO_HEIGHT); }
 
+  /**
+   * Set the card to be actionable button.
+   * @param {boolean | null} value The card can act as a button.
+   */
+  set actionable(value) {
+    const val = stringToBool(value);
+    if (stringToBool(value)) {
+      this.setAttribute(attributes.ACTIONABLE, val);
```
tmcconechy, line 203: This is probably not too important since i dont think they would change this on and off but we could for consistency make this adjust the template if you turn it on and off. from being actionable. I guess that would have to rerender the template. So can leave for now...

tmcconechy (commented):
A couple suggestions...Added. As per @danielortiz1982 issue i assume that would be something if we had a button too. So if you have a very long button and it did this we could make a new issue or you can have a quick look if you want.. As that does look off.

tmcconechy (approved):

ericangeles:
@tmcconechy we can raise an issue for the ripple effect one? And I'll try to take a look.

jmacaluso711 (approved):

Issue #428, opened 2021-11-16
Chore: Use Lit Like Attributes


tmcconechy:
**Describe the Task**
Use a lit like syntax to declare settings.

```
static get properties() {
  return {
    showDeviceSpecs: { type: Boolean, callback: () => {
      // do stuff when showDeviceSpecs changes
      this.#refreshProduct();
    }},
    data: { attribute: false },
    items: { type: Number, allowNaN: true },
  };
}
```

**What does it do?**

- Capture instance values for declared properties. Apply any property values that are set before the browser registers a custom element definition.
- Set up an observed (not reflected) attribute with the lowercased name of each property.
- Handle attribute conversion for properties declared as type String, Number, Boolean, Array, and Object.
- Use direct comparison (oldValue !== newValue) to test for property changes

PR #508, opened 2022-01-25
428 - Added IdsAttributesMixin (to autowire attribute getters+setters)

clayinfor:
# Ids Attributes Mixin

This mixin adds attribute getter/setter helper functionality to components

Closes https://github.com/infor-design/enterprise-wc/issues/428

### DIFF
--- a/src/mixins/ids-attributes-mixin/ids-attributes-mixin.js
+++ b/src/mixins/ids-attributes-mixin/ids-attributes-mixin.js
@@ -9,9 +9,9 @@ const capitalize = (word) => word.charAt(0).toUpperCase() + word.slice(1).toLowe
 
 const camelCase = (word) => word.split('-').reduce((curr, next) => (`${curr}${capitalize(next)}`));
 
-export const defineProp = (target, name, defaultValue, { changing, changed }) => {
-  const attrName = String(name?.attrName || name).trim();
-  const propName = String(name?.propName || camelCase(attrName)).trim();
+export const defineProp = (target, name, defaultValue, { propName, before, after }) => {
+  const attrName = String(name).trim();
+  propName = String(propName || camelCase(attrName)).trim();
   const propTypeCaster = target.propTypeCasters[typeof (defaultValue)] || Boolean;
   const stringCaster = (typeof (defaultValue) === 'object') ? JSON.stringify : String;
 
@@ -20,9 +20,9 @@ export const defineProp = (target, name, defaultValue, { changing, changed }) =>
   const getter = () => (propTypeCaster(target.getAttribute(attrName)) ?? defaultValue);
   const setter = (value) => {
     const stringValue = stringCaster(value);
-    changing?.bind(target)?.(stringValue);
+    before?.bind(target)?.(stringValue);
     target.setAttribute(attrName, stringValue);
-    changed?.bind(target)?.(stringValue);
+    after?.bind(target)?.(stringValue);
   };
 
   const propDescriptor = {
@@ -40,8 +40,8 @@ export const defineProp = (target, name, defaultValue, { changing, changed }) =>
     propDescriptor,
     propTypeCaster,
     stringCaster,
-    changing,
-    changed,
+    before,
+    after,
     defaultValue,
   };
 };
@@ -51,7 +51,9 @@ export const defineProp = (target, name, defaultValue, { changing, changed }) =>
  *
  * Possible usage could look like this:
  *
- * +  @attr(attributes.ALIGN, 'center', { changing, changed, })
+ * +  import { decorator as attr } from 'mixins/ids-attributes-mixin/ids-attributes-mixin';
+ * +
+ * +  @attr(attributes.ALIGN, 'center', { before, after, })
  * +  class IdsCustomComponent extends Base {}
  *
  * @param  {...any} args - attribute settings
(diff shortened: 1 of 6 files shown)


src/mixins/ids-attributes-mixin/ids-attributes-mixin.js
```diff
@@ -0,0 +1,66 @@
+const TypeCasters = {
```
clayinfor, line 1: This could be pluggable...

tmcconechy:
I think would need to see how this looks in one of the converted components... 

Also debating if it should go in IdsElement or be a mixin (i guess if not all elements have attributes then a mixin does sense).

Didnt do a full review yet...

src/mixins/ids-attributes-mixin/ids-attributes-mixin.js
clayinfor:
```javascript
// IdsDataGrid.js could do something like this to override typecasters
get propTypeCasters() {
  return {
    ...super.propTypeCasters(),
    boolean: stringToBool,
    object: this.mySophisticatedDataParser,
  };
}
```

tmcconechy: What would this be useful for? Not picturing it.

src/mixins/ids-attributes-mixin/ids-attributes-mixin.js
```diff
@@ -0,0 +1,135 @@
...
+  const propTypeCaster = target.propTypeCasters[typeof (defaultValue)] || Boolean;
+  const stringCaster = (typeof (defaultValue) === 'object') ? JSON.stringify : String;
+
+  target.setAttribute(attrName, stringCaster(defaultValue)); // initialize attribute
+
+  const getter = () => (propTypeCaster(target.getAttribute(attrName)) ?? defaultValue);
+  const setter = (value) => {
+    const stringValue = stringCaster(value);
+    changing?.bind(target)?.(stringValue);
+    target.setAttribute(attrName, stringValue);
```
EdwardCoyle, line 24: @clayinfor I wonder if this handles boolean attributes?  Typically for those, you would want to fully remove the attribute if it were to become `false`, per the spec: https://github.com/infor-design/enterprise-wc/blob/main/doc/COMPONENTS.md#boolean-attributes

We might also want to talk about enum-style attributes, where they could either be `null` (no attribute), or be defined with a string type.  This is currently how the `colorVariant` property passed from IdsColorVariantMixin behaves: https://github.com/infor-design/enterprise-wc/blob/main/src/mixins/ids-color-variant-mixin/ids-color-variant-mixin.js#L63-L67

EdwardCoyle (commented):
Just adding a comment on something we've run into regarding attributes in the past

src/mixins/ids-attributes-mixin/ids-attributes-mixin.js
clayinfor: It gives the component a way to override the default type-casters which as follows:
```javascript
const TypeCasters = {
  boolean: Boolean,
  number: Number,
  string: String,
  object: JSON.parse,
};
```

tmcconechy: ok, not sure how to work it in but for object/array we might want to use this deepCopy instead https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-data-grid/ids-data-grid.js#L5 of JSON.parse as that could loose something...

Just putting it here for now

src/mixins/ids-attributes-mixin/ids-attributes-mixin.js
clayinfor: Thanks for pointing this out, @EdwardCoyle!  I think we should be able to easily implement that spec here.  

If, however, it turns out to be more complicated that I'm thinking, we could consider continuing to use `get/set exampleProp()` as we're currently doing.

src/mixins/ids-attributes-mixin/ids-attributes-mixin.js
```diff
@@ -0,0 +1,145 @@
...
+  const stringCaster = (typeof (defaultValue) === 'object') ? JSON.stringify : String;
+
+  target.setAttribute(attrName, stringCaster(defaultValue)); // initialize attribute
+
+  const getter = () => (propTypeCaster(target.getAttribute(attrName)) ?? defaultValue);
+  const setter = (value) => {
+    const stringValue = stringCaster(value);
+    before?.bind(target)?.(stringValue);
+    target.setAttribute(attrName, stringValue);
+    after?.bind(target)?.(stringValue);
```
jmacaluso711, line 25: Just making a note here about the names of the callbacks.. Initially what came to mind for me was `beforeChange` and `changed` Maybe want to put it up for debate in a Teams survey or further discussion in next weeks wc meeting?

jmacaluso711 (commented):
@clayinfor This is looking really great.. Left a comment about some of the callback naming conventions. Also wondering how much, if any, this could be improved/streamlined by adding typescript to our stack. I guess we'd need to do a cost-benefit analysis there.

src/mixins/ids-attributes-mixin/ids-attributes-mixin.js
tmcconechy: Id rather not go add typescript to the whole stack. That seems over kill to me.

danielortiz1982:
@tmcconechy My outlook and opinion on implementation, I think this should be baked into `IdsElement` as it is core functionally that will built around the `attributeChangedCallback` and `observedAttributes` lifecycle functions. Just my thoughts on it. I feel like a mixin could work also.

src/mixins/ids-attributes-mixin/ids-attributes-mixin.js
danielortiz1982: @jmacaluso711 I like the idea, let Polly decided 😃

tmcconechy: For this i like beforeChange and changed as well.

src/mixins/ids-attributes-mixin/ids-attributes-mixin.js
```diff
@@ -0,0 +1,145 @@
...
+   * +      super();
+   * +
+   * +      this.prop(attributes.DATA, { fname: "John", lname: "Doe" });
+   * +      this.prop(attributes.AUTO_FOCUS);
+   * +
+   * +      // Reacting to prop changes using method-chaining
+   * +      this.prop(attributes.ALIGN_Y, 'top').before(this.alignYBefore).after(this.alignYAfter);
+   * +
+   * +      // Reacting to prop changes using the options argument
+   * +      this.prop(attributes.ALIGN_X, 'start', { after: this.alignXAfter });
```
tmcconechy, line 93: I like this approach the best so far. Might want to see if we can group them so its less verbose in an array or object or method properties.

And if we do call it properties i think the mixin would become propertyMixin (we do have a concept of adding attributes like id's / data-automation-id) so might want to reserve that name

tmcconechy: But probably lets first make a decision on how the API looks first

tmcconechy:
i agree @danielortiz1982 i did think at first it should be baked into IdsElement but the only thing is Not Every component does have attributes but if its that basic you probably just inherit from HTMLElement. So I guess i agree about baking it in as thats what i was originally thinking about this

tmcconechy:
@clayinfor 

Discussed a bit with @lucacolumbu and we think if we can do it like this... it would be 💯 

```
export default class IdsBadge extends Base {
  @prop('disabled', Boolean)
  @prop('color', String)
  @prop('shape', Object, onShapeChanged, ...otheroptions)

  constructor() { // Not sure if has to be first
      super();
  }

  // Later
  onShapeChanged() {
  }
  
  // Fires after every prop is set
  onAllPropertiesSet() {
  }
```

I think the decorators can be there like this just need to workout the syntax... it it could go here https://github.com/infor-design/enterprise-wc/blob/main/src/core/ids-decorators.js

Feature Use Cases
- Only properties (no attributes)
- Fire an event only when the last property is set on loading
- Two settings depend on each other and fire changed when both are set

EdwardCoyle:
Regarding this one:

> Only properties (no attributes)

Came to my mind that it could be handled this way... arguable whether or not "observed by attribute" is a default:
```js
@prop('shape', Object, onShapeChanged, {
  observed: false
})
```

tmcconechy:
Just to confirm: i like Eds addition so put it in

tmcconechy:
Will close this for now and can reraise

PR #529, opened 2022-01-29
350 - Add ids editor

### LINKED ISSUE #350: IdsEditor: New Web Component
Add a new SVG editor component MVP https://main-enterprise.demo.design.infor.com/components/editor/list
The new name should be IdsEditor (ids-editor). 

For this one we should try to mostly directly port the code over and use slots for the toolbar to provide the needed buttons. Maybe take the selection code and make it a util/mixin

**Phase 1 Features (MVP)**
- Toolbar
- Selections
- Pasting
- Editing
- States
- Rows
- Main features

**Not Needed Phase One**
- Empty State
- Modal
- Anything legacy 

**Possible Internal Markup**
```js
<ids-editor>
  <ids-toolbar>...</ids-toolbar>
  <ids-editor-content>
  ...
  </ids-editor-content>
  <ids-html-content>
  ...
  </ids-html-content>
</ids-editor>
```

**Suggested Dependencies**
- ids-text
- ids-toolbar
- ids-buttons
- ids-font-picker (if a seperate component)

**Design**
https://main-enterprise.demo.design.infor.com/components/editor/example-index.html

**Events**
- change

**Other**

deep7102:
Added Ids Editor component

Closes #350

- Navigate to:
- http://localhost:4300
- http://localhost:4300/ids-editor
- http://localhost:4300/ids-editor/sandbox.html
- http://localhost:4300/ids-editor/side-by-side.html
- See editor all variations and uplift styles 
- Added e2e and functional tests
- See the docs


src/components/ids-editor/ids-editor.js
```diff
@@ -0,0 +1,2203 @@
...
+   * @param {Event} e The event
+   * @returns {string|null} The updated pasted data
+   */
+  #pasteAsHtml(e) {
+    if (!e) return null;
+
+    const clipboardData = e.clipboardData;
+    let html;
+
+    if (clipboardData?.types) {
```
danielortiz1982, line 1450: This came out really clean, however I feel this can be optimized further if you perhaps used the `navigator.clipboard.readText()` and `navigator.clipboard.write(data)` API. I think using these API's will reduce the amount of checking you will need to perform, also it has a wider browser support and returns a promise. This feedback is under the condition you will not need to use the `Permissions API` 


https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/readText
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write

src/components/ids-editor/ids-editor.js
```diff
@@ -0,0 +1,2203 @@
...
+        html = e.clipboardData.getData('text/plain');
+      }
+      if ((typeof types === 'object' && types[0] === 'text/plain') && !types[1]) {
+        html = e.clipboardData.getData('text/plain');
+      }
+    } else {
+      const paste = window.clipboardData ? window.clipboardData.getData('Text') : '';
+      const paragraphs = paste.split(/[\r\n]/g);
+      html = '';
+      for (let i = 0, l = paragraphs.length; i < l; i++) {
```
danielortiz1982, line 1466: 100% code style option here but I think this should be changed to a `forEach` instead. In my opinion it's cleaner because it is a lot more readable and also forEach keeps the variable’s scope to the block which will help in future code updates that are less prone to errors. Having to use a temporary `i` variables can add a lot of confusion to your code. This change I am requesting is just my humble opinion.

danielortiz1982 (changes requested):
Please change all while statements on lines 571, 638, 784, 791 to `for` or `forEach`. The iterators for and forEach is a better solution because they are less prone to memory leaks and infinite loops locking up the IO and window.

src/components/ids-editor/ids-editor.js
```diff
@@ -0,0 +1,2203 @@
...
+} from '../ids-toolbar/ids-toolbar';
+
+import { debounce } from '../../utils/ids-debounce-utils/ids-debounce-utils';
+import { sanitizeHTML } from '../../utils/ids-xss-utils/ids-xss-utils';
+import { stringToBool } from '../../utils/ids-string-utils/ids-string-utils';
+import { isObject } from '../../utils/ids-object-utils/ids-object-utils';
+
+import styles from './ids-editor.scss';
+
+// List of view modes
```
danielortiz1982, line 24: Can you possible move the settings from line 24 to 72 into it's own file. You can see IdsButton as a reference. 

IdsButton Attributes:
https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-button/ids-button-attributes.js

danielortiz1982 (changes requested):
Overall I think this came really well, great job! One thing I would like to see on the `ids-editor.js` file. Ids-editor.js is currently over 2,000 lines so maybe we can break this down into different files to make future updates easier and more readable. Perhaps you can pick out core functionality and group them into there own file and import them. This will reduce the file size while making the logic more modular

deep7102:
> Please change all while statements on lines 571, 638, 784, 791 to `for` or `forEach`. The iterators for and forEach is a better solution because they are less prone to memory leaks and infinite loops locking up the IO and window.

agree with concern about `while` loops it can be wonky if not use carefully, so will update as `for` or `forEach`, but some will still there because thinking `while` is good place for it

src/components/ids-editor/ids-editor.js
deep7102: yes, I also had looked thru this api and personally liked it, but had few points as below which not supported that's why did not go with it.

1) it only support `https`, and it cannot be used from a content script running on `http:` pages
2) to paste just text content we can use `navigator.clipboard.readText()` which is fine, but in order to use images with clipboard it require to use `navigator.clipboard.read()` and which is not fully supported yet with new/major browsers like chrome/firefox https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read
3) this requires the permission to "clipboardRead" or "clipboardWrite" (I think this is fine)

deep7102:
> Overall I think this came really well, great job! One thing I would like to see, `ids-editor.js` is currently over 2,000 lines so maybe we can break this down into different files to make future updates easier and the codebase more readable. Perhaps you can pick out core functionality and group them into there own file and import them. This will reduce the file size while making the logic more modular

Great point, like this to be breakdown in smaller parts, will do as much can -thanks

demos/ids-editor/index.yaml
```diff
@@ -0,0 +1,11 @@
+---
+ component: Editor
+ link: ids-editor
+ description: An input for multi line editable text
```
tmcconechy, line 4:
```suggestion
 description: An input for multi line rich text
```

demos/ids-editor/index.yaml
```diff
@@ -0,0 +1,11 @@
...
+ category: Form Inputs
+ keywords: theme, color, switcher
```
tmcconechy, line 6:
```suggestion
 keywords: editor, rich text, content editable, RTE
```

doc/CHECKLIST.md
```diff
@@ -69,19 +69,19 @@ Will get a checkbox at minimum viable product. The rest of the details are cover
...
- - [ ] Editor (ids-editor)
+ - [x] Editor (ids-editor)
  - [ ] Emptymessage (ids-empty-message)
  - [x] Error Page (ids-page-error)
  - [x] Expandable area (ids-expandable-area)
  - [ ] Fieldset (ids-field-set)
  - [x] Fileupload (ids-file-upload)
  - [x] Fileupload Advanced (ids-file-upload-advanced)
- - [ ] Fontpicker (ids-font-picker)
+ - [ ] ~~Fontpicker (ids-font-picker)~~
```
tmcconechy, line 79:
```suggestion
 - [x] Fontpicker (added as part of editor)
```

src/components/ids-editor/README.MD
```diff
@@ -0,0 +1,282 @@
...
+    <p>Some sample text</p>
+  </div>
+</ids-editor>
+```
+
+Disabled editor component.
+
+```html
+<ids-editor label="Demo Editor" disabled>
+  <div slot="content">
```
tmcconechy, line 46: Is there more than one slot?

src/components/ids-editor/ids-editor.js
danielortiz1982: okay cool, just wanted to throw it out there... I will leave it up to your better judgement. 👍

tmcconechy (changes requested):
This looks really good Deepak 🥇 

Only small things from me for now. The only functional thing i saw in addition while testing was that on http://localhost:4300/ids-editor/sandbox.html on the dirty example i did a quick test to make it bold and then not bold and the dirty indicator stayed when it shouldnt.

src/components/ids-editor/README.MD
tmcconechy: Yeah thats what i was getting at but do we have a html view/normal view slot?

src/utils/ids-debounce-utils/ids-debounce-utils.js
```diff
@@ -0,0 +1,29 @@
...
+      timeout = null;
+    };
+
+    if (timeout) {
+      clearTimeout(timeout);
+    } else if (execAsap) {
+      func.apply(obj, args);
+    }
+
+    timeout = setTimeout(delayed, threshold || 250);
```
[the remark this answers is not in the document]
deep7102: Yes, did try and was first preferences `requestAnimationFrame` but did not help.

deep7102:
@tmcconechy @danielortiz1982 it's updated, and ready for review now.

tmcconechy (approved):

tmcconechy (approved):

danielortiz1982 (approved):
This looks great @deep7102 really nice work!

Issue #524, opened 2022-01-27
IdsMenuButton: Styling issues in RTL Mode


tmcconechy:
**Describe the bug**
The arrow styles are off in RTL mode.

**To Reproduce**
1. Go to http://localhost:4300/ids-tag or http://localhost:4300/ids-menu-button
2. In the html inspector find `ids-container` and change `language="ar"` to force RTL
3. Click ....

**Expected behavior**
Fixed styles in RTL mode for menu button/theme switcher

**Screenshots**
![Screen Shot 2022-01-27 at 1 52 23 PM](https://user-images.githubusercontent.com/814283/151424695-b368b01d-24fe-4fae-9286-5f313cde785d.png)

andriivarhanov:
@tmcconechy I would like to also add style issue of `ids-popup` arrow in RTL mode, same steps to reproduce just on the page http://localhost:4300/ids-menu-button
![image](https://user-images.githubusercontent.com/5204821/151548510-acf39f70-190e-4c0b-82de-8c623cd25b5e.png)

tmcconechy:
Right! Pretty sure a css fix would fox both cases..

PR #545, opened 2022-02-09
524 - Update locale mixin

jeffwa428:
Fix arrow styling issues in RTL Mode.

Closes https://github.com/infor-design/enterprise-wc/issues/524

- Go to http://localhost:4300/ids-tag or http://localhost:4300/ids-menu-button
- In the html inspector find ids-container and change language="ar" to force RTL
- Click ....
- Arrow style should work


src/mixins/ids-locale-mixin/ids-locale-mixin.js
```diff
@@ -12,6 +12,35 @@ const IdsLocaleMixin = (superclass) => class extends superclass {
     this.offEvent('languagechange.mixin');
     this.onEvent('languagechange.mixin', this.closest('ids-container'), async () => {
       this.setDirection();
+
+      const elements = [];
```
tmcconechy, line 16: My only comment would be we could make a util in https://github.com/infor-design/enterprise-wc/blob/main/src/utils/ids-dom-utils/ids-dom-utils.js#L44 for this in case its used again.

Maybe `getIdsElements` as a name? Also if we can avoid some loops somehow. Maybe @EdwardCoyle has an idea of other use cases for finding all elements of a certain type...

`querySelectorShadow(selector)` -> i dont there is a way to make a query selector that starts with `ids` but this would be useful.

tmcconechy (commented):
Check out my suggestion

tmcconechy (commented):
@jeffwa428 one more idea. If you add bubbles here:

https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-container/ids-container.js#L139

```
this.triggerEvent('languagechange', this, { bubbles: true, detail: { elem: this, language: this.language, locale: this.state.locale } });
```

Does the event fire on the children? If so might be a better approach

tmcconechy (changes requested):
@jeffwa428 @jdilag proposed this function (thanks @jdilag !!)

```
function getIdsElements(node, pierce = false) {
    if (node == null) return [];
    let idsElements = node.tagName.indexOf('IDS-') !== -1 ? [node] : [];
    let shadowChildren = pierce && node.shadowRoot ? [...node.shadowRoot.children] : [];
    let children = [...node.children].concat(shadowChildren);
    for (const child of children) {
      idsElements = idsElements.concat(getIdsElements(child, pierce));
    }
    return idsElements;
  }
```

See if it works 
```
example usage -> getIdsElements(document.body, true)
```

src/mixins/ids-locale-mixin/ids-locale-mixin.js
EdwardCoyle: Honestly, I'm not sure I like the idea of this mixin automatically scrubbing shadow root for the elements it needs to adjust...

It would be a bit more work to get all the integrations right, but I'd rather see something like an `onDirectionChange` callback [after this](https://github.com/infor-design/enterprise-wc/blob/main/src/mixins/ids-locale-mixin/ids-locale-mixin.js#L45-L49) that each component could implement for whenever a change to `dir` happens on a component.

jeffwa428:
@tmcconechy setting bubbles is not working.
I added new util function and used it in locale mixin.

src/utils/ids-dom-utils/ids-dom-utils.js
```diff
@@ -109,3 +109,21 @@ export function getEditableRect(rect) {
...
+ * @param {boolean} pierce indicate if this search will include shadowRoot
+ * @returns {HTMLElement[]} array of elements
+ */
+export function getIdsElements(node, pierce = false) {
+  if (node == null) return [];
+
+  let idsElements = node.tagName.indexOf('IDS-') !== -1 ? [node] : [];
+  const shadowChildren = pierce && node.shadowRoot ? [...node.shadowRoot.children] : [];
+  const children = [...node.children].concat(shadowChildren);
+  for (const child of children) {
```
danielortiz1982, line 125: @jeffwa428 This is a really nice solution, however in my opinion it would look cleaner with a forEach and a return idsElements statement as a one liner. This is just a recommendation for a cleaner code pattern approach. 👍

src/mixins/ids-locale-mixin/ids-locale-mixin.js
```diff
@@ -12,6 +13,16 @@ const IdsLocaleMixin = (superclass) => class extends superclass {
     this.offEvent('languagechange.mixin');
     this.onEvent('languagechange.mixin', this.closest('ids-container'), async () => {
```
danielortiz1982, line 14: @jeffwa428  @tmcconechy  This looks good, one thing. I noticed we are using an `async` fn however I do not see an await statement anywhere... Can this be changed to a normal callback or is the async really needed?

tmcconechy: Yeah your right good catch. None of these `this.onEvent('language...` handlers need `async` unless there is something to await in it. Could clean this/others up?

tmcconechy:
@jeffwa428 

Can you look at @EdwardCoyle suggest and then lets just change this so that.

- we dont scan teh whoel dom and automatically add it to all of them
- add a callback onDirectionChange that fires when it changes direction
- in the popupmenu implement onDirectionChange and copy the property down

We can talk more on the pair programming tomorrow if unclear

Issue #537, opened 2022-02-01
IdsCard: Add more examples


tmcconechy:
**Describe the Enhancement**
Add additional card examples for the secondary toolbar area and more variations (with hit points if not already present).

**Specs**
- [ ]  Add a secondary toolbar example (also called group action toolbar) see: https://main-enterprise.demo.design.infor.com/components/cards/example-group-action.html for the example. This would use ids-toolbar and scroll under (showing an ids-listview) and other example https://main-enterprise.demo.design.infor.com/components/bar-stacked/test-fit-height.html
- [ ]  Add variations example like the old version https://main-enterprise.demo.design.infor.com/components/cards/example-variations-hitboxes.html and any css if needed. This includes the option in IdsIcon to rotate the button.
- [ ] Make sure the buttons have 44px hit points when in a card as per the previous example that hangs off but doesnt change button size https://main-enterprise.demo.design.infor.com/components/button/example-button-with-hitbox.html (option name hitbox="true")

**Additional context**
#536 

```
<ids-card>
  <ids-toolbar slot="group-actions">
    <ids-text></ids-text>
    <ids-button></ids-button
```

PR #559, opened 2022-02-22
537 - Add ids-card examples

jeffwa428:
Add more ids-card examples
- Add a secondary toolbar example like this https://main-enterprise.demo.design.infor.com/components/cards/example-group-action.html
- Add variations like https://main-enterprise.demo.design.infor.com/components/cards/example-variations-hitboxes.html

Closes https://github.com/infor-design/enterprise-wc/issues/537

- go to http://localhost:4300/ids-card/toolbar.html to see the toolbar example
- go to http://localhost:4300/ids-card/info.html to see the variants


demos/ids-card/toolbar.scss
```diff
@@ -0,0 +1,17 @@
+.parent-container {
+  height: 300px;
+}
+
+#primary-toolbar {
```
tmcconechy, line 5: We shouldnt style by ID as this may change. I think need to use a class.

tmcconechy (commented):
This looks good..Only see a couple things

- maybe make the new cards on a new page
- i think they need a better name than "variant" but not sure what? will research
- check the lint errors
- can you add a visual test / check test coverage on the card code
- even tho the old version is like this i think it is wrong - can we make the title text line up 
Screen Shot 2022-02-22 at 10 35 54 AM

tmcconechy:
@jeffwa428 as discussed in the call lets try to do this without any special variant class. i.e.

- card gets auto-height setting to fit to contents
- add a card-actions (or similar) sub component that floats action buttons on the right (or left in RTL) with flex or vss grid
- see markup examples https://getbootstrap.com/docs/4.0/components/card/
- ids-text will just get proper padding between

tmcconechy:
```
<ids-card auto-height="true">
  <div slot="card-content">
    <ids-text color="green" font-size="14">Action Item</div>
    <ids-text font-size="14">Submit 360 Review</div>
    <ids-text font-size="14">Send review by May 2nd</div>

    <ids-card-actions> <!-- flex to the right / also test rtl-->
       <!-- look at ids-button -->
      <ids-button icon="..."></ids-button> 
    </ids-card-actions>
  </div>
</ids-card>

<ids-card auto-height="true">
  <div slot="card-content">
    <ids-image src="photo.jpg" alt="photo"></ids-image>
    <ids-text color="green" font-size="14">Action Item</div>
    <ids-text font-size="14">Submit 360 Review</div>
    <ids-text font-size="14">Send review by May 2nd</div>

    <ids-card-actions> <!-- flex to the right / also test rtl-->
       <!-- look at ids-button -->
      <ids-button icon="..."></ids-button> 
    </ids-card-actions>
  </div>
</ids-card>
```

demos/ids-card/index.yaml
```diff
@@ -17,4 +17,7 @@
     description: Showing cards with single selection
   - link: multiple-selection.html
     description: Showing cards with multiple selection
-
+  - link: toolbar.html
+    description: Showing cards with secondary toolbar
+  - link: variants.html
```
tmcconechy, line 22: Can we rename this to `info.html` - i.e. info cards.... (cards with various dynamic information)

demos/ids-card/toolbar.scss
```diff
@@ -0,0 +1,17 @@
...
+}
+
+.secondary-toolbar {
+  padding: 0 16px;
+  border-top: 1px solid var(--ids-color-palette-slate-20);
+  border-bottom: 1px solid var(--ids-color-palette-slate-20);
+}
+
+button::after {
+  content: 'a';
```
tmcconechy, line 16: Is this right?

demos/ids-card/variants.html
```diff
@@ -0,0 +1,118 @@
...
+<ids-layout-grid auto="true">
+  <ids-text font-size="12" type="h1">Cards</ids-text>
+</ids-layout-grid>
+
+<ids-layout-grid cols="3" gap="md">
+  <ids-layout-grid-cell>
+    <ids-card auto-height="true">
+      <div slot="card-content">
+        <div>
+          <ids-text labelFontSize="16" color="emerald-60">Action Item</ids-text>
```
tmcconechy, line 12: `labelFontSize` should be `font-size`

demos/ids-card/variants.scss
```diff
@@ -0,0 +1,82 @@
+.ids-card-content-variant {
```
tmcconechy, line 1: I dont think we need this anymore..

demos/ids-card/variants.js
```diff
@@ -0,0 +1,10 @@
+import './variants.scss';
```
tmcconechy, line 1: Remove?

demos/ids-card/variants.scss
```diff
@@ -0,0 +1,82 @@
...
+      margin-top: 2px;
+    }
+  }
+
+  ids-image {
+    padding-right: 16px;
+  }
+}
+
+[slot='card-content'] {
```
tmcconechy, line 77: I tried to remove this but both http://localhost:4300/ids-card/toolbar.html and http://localhost:4300/ids-card/info.html do get messed up.

The goal would be that the cards are dynamic and work without adding any special css so maybe this needs to go in the card somehow

src/components/ids-card/ids-card-action-base.js
```diff
@@ -0,0 +1,14 @@
+import IdsEventsMixin from '../../mixins/ids-events-mixin/ids-events-mixin';
+import IdsThemeMixin from '../../mixins/ids-theme-mixin/ids-theme-mixin';
```
tmcconechy, line 2: We probably only need IdsElement in here not the other mixins..

src/components/ids-card/ids-card-action.js
```diff
@@ -0,0 +1,31 @@
+import { customElement, scss } from '../../core/ids-decorators';
+import Base from './ids-card-action-base';
+import styles from './ids-card-action.scss';
+
+/**
+ * IDS Card Action Component
+ * @type {IdsCardAction}
+ * @inherits IdsElement
+ * @mixes IdsEventsMixin
```
tmcconechy, line 9: Same here - just IdsElement

src/components/ids-card/ids-card-action.scss
```diff
@@ -0,0 +1,8 @@
+:host {
+  margin-left: auto;
+}
+
+:host([dir='rtl']) {
+  margin-left: 0;
+  margin-right: auto;
```
tmcconechy, line 7: Is this needed? Not sure what auto would do in terms of margin? But it would be better to use the mixins here as it works with margin-start/end instead of left and right.. fx

```
@include ml-4;
@include mx-4;
```

tmcconechy (changes requested):
- Minor cleanup
- Rename variants to info for the example
- Try to do this without any needed example csss

src/components/ids-card/ids-card-action.scss
tmcconechy: The whole thing could be
```
host {
    margin-inline-start: auto;
}
```

tmcconechy: See https://github.com/infor-design/design-system/issues/478 for more info

demos/ids-card/toolbar.scss
```diff
@@ -0,0 +1,13 @@
+.parent-container {
+  height: 300px;
+}
+
+.primary-toolbar {
```
tmcconechy, line 5: @jeffwa428 the only question is is this needed? As the styles should be in the component so the end user doesnt need any css? Maybe it works without this?

src/components/ids-card/ids-card.scss
```diff
@@ -101,6 +101,15 @@
   }
 }
 
+:host([flex-content]) {
+  ::slotted([slot='card-content']) {
```
EdwardCoyle, line 105: @jeffwa428 are these styles needed to shape the component?  If so, we also need to consider standalone CSS usage (non-webcomponent, styles only).  Currently the way this is written, we can't use this new feature when only the styles are used, due lack of a host element.

Could these same rules instead be applied to a CSS class that we would add to the Card's container element (the first element in Shadow root that has the class `.ids-card`)?  Maybe we can use a css class `.flex-content` to match the attribute name?

EdwardCoyle (commented):
In addition to Tim's comments, I had one more request for this one.  Overall, looks good and will approve once comments are addressed 👍

src/components/ids-card/ids-card.scss
EdwardCoyle: For reference, run your local demoapp and see the markup for IdsCard on this URL:  http://localhost:4300/ids-card/standalone-css.html

tmcconechy (commented):
The only (last problem) i see is if you load http://localhost:4300/ids-card/info.html and look in the console you will see an error:

Uncaught TypeError: Cannot read properties of undefined (reading 'call')
    at HTMLElement.connectedCallback (ids-card-action.js:18:5)

tmcconechy (approved):

tmcconechy:
@jeffwa428 i synced this with main for you after the big merge. Looks good to me but you might want to take a quick look everything is as expected

EdwardCoyle (approved):
Looks good!

tmcconechy (approved):

PR #560, opened 2022-02-22
542 - Listbuilder editing fix

### LINKED ISSUE #542: IdsListBuilder: Loses Value when editing
**Describe the bug**
Its possible to make listbuilder loose the value you edited. (noticed doing some testing)

**To Reproduce**
1. Go to http://localhost:4300/ids-list-builder
2. Click on an item
3. click edit
4. type something
5. click a different item

**Expected behavior**
It only seems to save on Enter. I think we should just make it save each keydown instead so you wont loose values and have to hit enter every time.

Also add some tests for this case

jdilag90:
When editing a list item, clicking another list item applies user input to that clicked item.
Fix makes it so that user input updates on keyup rather than waiting for 'Enter' keypress/focusout.
Added e2e tests for arrow navigation for more code coverage.

Closes #542 

2. Click on an item, then click edit button
3. Type anything
4. Click a different list item
5. Check that original item has updated value


src/components/ids-list-builder/ids-list-builder.js
```diff
@@ -135,8 +134,8 @@ export default class IdsListBuilder extends Base {
    * Helper function to remove the editor element from the DOM
    */
   #removeSelectedLiEditor() {
-    this.selectedLi.querySelector('ids-text').style.display = 'list-item';
-    this.selectedLi.parentNode.removeAttribute('disabled');
+    this.offEvent('keyup', this.#selectedLiEditor);
+    this.#selectedLiEditor.nextElementSibling.style.display = 'list-item';
```
danielortiz1982, line 138: This is a nice solution @jdilag90. However I was wondering if this can be achieved by adding and removing a class with the `list-item` property instead of doing an inline style.

tmcconechy (approved):

danielortiz1982 (approved):

PR #568, opened 2022-03-01
487 - IdsSwappable / swaplist

### LINKED ISSUE #487: IdsDraggable: Use drag API / Mobile Support / Make it a Mixin
**Describe the bug**
Refactor the current ids-draggable to use the touch api and not mouseevents. If possible test mobile support for example IOS might now support it https://caniuse.com/?search=drag 

**Tasks**
- Does it work on mobile devices? If so which ones
- Remove all "strange" code and refactor drag api
- should we change to both a mixin and a component (so we dont need to wrap things in all cases)
- add to swaplist (and any other places)

**Documentation**
- https://www.w3schools.com/howto/howto_js_draggable.asp
- https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API
- https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable

**Additional context**
Swap list could not use it...

jmacaluso711:
This PR adds the IdsSwap list as well as IdsSwappable component. The abstracted component is currently named `ids-swappable` which can be used to contain any component you would like to with drag and drop.

closes #487 

- pull and run branch
- go to http://localhost:4300/ids-swaplist
- make sure you can select an item and drag it to another list.
- test multiple selection
- test select and swap with arrow buttons
- Reference enterprise: https://main-enterprise.demo.design.infor.com/components/swaplist/example-index.html
- check example usage of `ids-swappable` with `ids-cards` at http://localhost:4300/ids-swaplist/cards.html (WIP) missing come styling.


src/components/ids-swappable/ids-swappable.d.ts
```diff
@@ -0,0 +1,12 @@
+import IdsSwappableItem from './ids-swappable-item';
+
+export default class IdsSwappable extends HTMLElement {
+  /* Sets the `multi-select` property to enable/disable multi-select ids-swappable-items */
+  multiSelect?: boolean;
```
tmcconechy, line 5: On some other components we have `selection="multiple"` so maybe a slight inconsistencY

src/components/ids-swappable/ids-swappable.js
```diff
@@ -0,0 +1,246 @@
...
+  scss,
+} from '../../core/ids-decorators';
+
+import { attributes } from '../../core/ids-attributes';
+import { stringToBool } from '../../utils/ids-string-utils/ids-string-utils';
+import Base from './ids-swappable-base';
+import styles from './ids-swappable.scss';
+
+/**
+ * IDS SwapList Component
```
tmcconechy, line 12:
```suggestion
 * IDS Swappable Component
```

src/components/ids-swappable/ids-swappable.js
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  connectedCallback() {
+    this.setAttribute('dropzone', 'move');
+    this.attachEventListeners();
+  }
+
+  static get attributes() {
+    return [
+      attributes.ACTIVE,
```
tmcconechy, line 32: Do we need ` ...super.attributes,`?

src/components/ids-swappable/ids-swappable.js
```diff
@@ -0,0 +1,246 @@
...
+    return `<slot></slot>`;
+  }
+
+  /**
+   * Get the item that is dragged over
+   * @returns {HTMLElement} ids-swappable-item[over]
+   * @readonly
+   * @memberof IdsSwappable
+   */
+  get overEl() {
```
tmcconechy, line 47: Maybe we make this overElement? (spelled out vs el?)

tmcconechy (changes requested):
Looks great a couple smaller comments. Also on the examples:

https://main-enterprise.demo.design.infor.com/components/swaplist/example-search.html
- compared to http://localhost:4300/ids-swaplist there is a 1px gap between the items 
- the placeholder is grey/slate/black and using dashes vs blue dotted

Biggest two questions are:
- removal of the template
- multiselect vs selection in other places...

Thanks!

src/components/ids-swappable/ids-swappable-item.js
```diff
@@ -0,0 +1,278 @@
...
+      this.offEvent('click', this, this.#toggleSelect);
+      this.onEvent('click', this, this.#toggleSelect);
+    }
+  }
+
+  /**
+   * Handle the drag events
+   */
+  #handleDragEvents() {
+    this.removeEventListener('dragstart', this.#dragStart.bind(this));
```
danielortiz1982, line 258: If supported, I think we should use `onEvent` for all of these event handlers. If it's not supported I think it's fine for now.

tmcconechy: yeah good catch - as onEvent will kill the event listener on delete

jmacaluso711: I tried using onEvent for these but getting this console error: `target.addEventListener is not a function` not exactly sure what the issue it.

tmcconechy: There is another signature for onEvent..Did you get the second param?

```
this.onEvent('click', this, () => this.#dragStart());
```

src/components/ids-swappable/ids-swappable.js
EdwardCoyle: @tmcconechy there's a handful of mixins applied in `ids-swappable-base.js` so I'd say "yes" here

src/components/ids-swappable/ids-swappable-item.scss
```diff
@@ -0,0 +1,40 @@
+@import '../../core/ids-base';
+
```
EdwardCoyle, line 2: @jmacaluso711 I'll leave this one to your discretion because I'm not 100% sure if we need this...

If there is any chance at all that these styles might need to live in a standalone-CSS environment, we should try to figure out how to use them in CSS classes that would be applied to a shadow root container element instead.  However, I can see it's currently just a slot, so that would maybe change the design of this... 

I'm thinking it may not be necessary due to all of this just facilitating the swap behavior, but strikes me as something that may come up in the future.

EdwardCoyle (commented):
This is very nicely done @jmacaluso711!

I had one comment on functionality here that is minor and maybe ok to leave for a Phase 2:
- Open http://localhost:4300/ids-swaplist
- Highlight one or multiple items in List #0
- Click the "Swap Item Right" button in List #1
- See the highlighted item(s) jump from List #0 to List #2
- The old swaplist didn't do this, but we could potentially fix this by not allowing swapping via buttons unless items within those buttons' list are highlighted (disable the buttons otherwise?)

One other thing that may be unrelated to your changes... the two buttons in List #1 both say "Swap Item Left" in their text content (accessibility problem)

Just had another comment below that is also something we can push to later

jmacaluso711:
@tmcconechy @danielortiz1982 @EdwardCoyle - Ready for re-review.. I think I addressed all comments, let me know if I missed anything.

@EdwardCoyle I ended up finding an easy fix for the swap button bug you found

tmcconechy (approved):

tmcconechy (commented):
@jmacaluso711 the only thing im just noticing after reviewing the https://percy.io/00aaa841/enterprise-wc/builds/16355875/changed/920684642?browser=chrome&browser_ids=20&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280 (not as obvious on http://localhost:4300/ids-swaplist)

Is that there is an un-needed scrollbar?

tmcconechy (approved):

tmcconechy:
@jmacaluso711  i synced this with main for you after the big merge. Looks good to me but you might want to take a quick look everything is as expected.

One thing i wondered about is

- http://localhost:4300/ids-swaplist/cards.html -> seems like the cards are too big and the select state looks a little strange. I didnt notice this the first time i reviewed. you could use this smaller actionable card maybe for the example https://main.wc.design.infor.com/ids-card/
- should this example be better filled under ids-swappable ?

jmacaluso711:
@tmcconechy Yeah I was trying to show that you can put anything inside an `ids-swappable-item`, but didn't spend time making the example look right. I will move the demo to `ids-swappable` and simplify, maybe just use `ids-text` for now

jmacaluso711:
@tmcconechy @EdwardCoyle ready for review

EdwardCoyle (approved):
Looks great @jmacaluso711

Issue #551, opened 2022-02-16
IdsColorPicker: Remaining Features/Issues


tmcconechy:
**Description**
Lists the Remaining Bugs/Features for a 100% Component

**Major**

- [ ] https://main-enterprise.demo.design.infor.com/components/editor/example-index.html - editor version
- [x] Fix the keyboard (arrow down , enter, keys around the colors ect), and fix focus state around swatches
- [x] Tooltip on the colors
- [x] Fix all the states [See example](https://main-enterprise.demo.design.infor.com/components/dropdown/test-states.html) and [also see](https://main-enterprise.demo.design.infor.com/components/colorpicker/test-states.html)
- [x] Check Styles match with 4.x on the popup container - swatch/hover/spacing/selected state on swatches
- [x] Default color list should load from our theme palette `node_modules/ids-identity/dist/theme-new/tokens/web/ui.config.color-palette.js`
- [x] Check mark color should change based on swatch background (check mark color can get from colorlist -arrowClass)
- [ ] Standardize/Abstract out Popup/Picker/"return value" interface (to be used in this and other pickers)
- [x] Check JSDoc comments for everything that's missing
- [x] Add test for change event firing
- [x] Add test for setting color / clearing color
- [x] Change hide/show to open/close for consistency and make sure it fires events that are documented
- [x] Ability to set custom labels for selections. Should be able to show color-name `azure06` or `color-hex` `#ffffff` inside the input. [See example](https://main-enterprise.demo.design.infor.com/components/colorpicker/example-custom-labels.html)
- [x] Show label instead of hex. [See example](https://main-enterprise.demo.design.infor.com/components/colorpicker/example-show-label.html)
- [x] Add sizes widths and field heights. [See Ids Input](https://main.wc.design.infor.com/ids-input/)
- [x] Add dirty tracker example to [the example page](https://main.wc.design.infor.com/ids-color-picker/)

**Minor**

- [ ] Some text hard coded in component, should give setting/slots to user change those text strings ids-text audible="true"
- [ ] Enable/Disable standard HTML colorpicker independently of other colorpicker parts using a boolean attribute
- [ ] Click the swatch to select / option for the "advanced" picker
- [ ] Add side by side example

PR #569, opened 2022-03-01
520 - Fix inheritance on IdsInput/IdsTriggerField and related components

### LINKED ISSUE #520: IdsInput/IdsTriggerField:  Fix Inheritance and extend IdsInput
Many components that are built on-top of IdsTriggerField are currently still slotting an IdsInput.  Awhile back we made the [IdsTriggerField extend IdsInput](https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-trigger-field/ids-trigger-field-base.js#L3) so it could take advantage of all attributes of IdsInput instead of needing to duplicate/redirect-to IdsInput's properties.  However the following list of components is still doing the duplication/routing because an IdsInput is being [used within the components' templates](https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-search-field/ids-search-field.js#L72-L87) or is [being slotted in the template](https://github.com/infor-design/enterprise-wc/blob/main/demos/ids-trigger-field/example.html#L53-L58), instead of leaning on inheritance for IdsInput features:

- IdsColorPicker
- IdsDropdown
- IdsLookup
- IdsPager
- IdsSearchField
- IdsSpinBox
- IdsTimePicker

- [x] We should be properly inheriting IdsInput or IdsTriggerField (depending on the component) in the preceeding list, instead of using the IdsInput within the template of these components, since they are functionally all input fields that need the same attributes.
- [x] IdsTriggerField's slots should become named slots that target a `start` and `end` trigger element, and usage of the IdsInput field should no longer be slotted, but accessible via inheritance
- [x] Fix IdsColorPicker to extend IdsTriggerField (currently has input/trigger in shadow root) and examples
- [x] Check IdsDatePicker implementation once #510 is merged
- [x] Fix IdsDropdown's templated Trigger field (this is fine but test all attributes)
- [x] Fix IdsLookup's templated Trigger field (also slotting an ids-input)
- [x] Check IdsPagerInput for accuracy (nothing should have changed) 
- [x] Fix IdsSearchField to extend IdsTriggerField (currently has input/trigger in shadow root) and examples
- [x] Check IdsSpinbox implementation (maybe should be a trigger field internally)
- [x] Fix IdsTimePicker's templated trigger field (also slotting an ids-input)
- [x] All examples of these components should be corrected (most of them slot an IdsInput)
- [ ] Color Variants for extended components may not be necessary once inheritance is implemented, so these styles should be reviewed.  The end visual result should be exactly the same as before.
- [ ] Complete the IdsSearchField color variants (specifically toolbar/alternate styles) for both Dark and Contrast mode

If we are able to correct the inheritance problem it will reduce the amount of code we are writing in all Input-based components, since re-wiring the attributes in every class will not be necessary on the JS side, and the CSS will become simpler since most styles will be defined at the Input level.

**Additional context**
- Discovered while working on the Search Field component during #439
- Some exploration and work has been done on this against this branch: `ids-input-refactor`

EdwardCoyle:
This PR changes the inheritance model between IdsInput, IdsTriggerField, and any components that use/extend these.  The following major changes exist in this PR:

- IdsTriggerField directly inherits IdsInput and no longer allows it to be added via slot
- Some components, such as IdsSpinbox, IdsSearchField, and IdsColorPicker directly inherit IdsTriggerField, and no longer allow it to be added via slot
- All attributes from IdsInput are accessible in the extended components and should work without additional getters/setters/inclusion of attributes in those extended components (provided no additional change is needed).
- Some components are still wrapping IdsTriggerField, but in those components the templates have been cleaned up to remove any extraneous elements and use correct attributes.
- Some components that need to do additional work when IdsInput's setters/getters run now just extend the setters/getters by using `set [attributeName](value) { super.[attributeName] = value; }` or `get [attributeName]() { return super.[attributeName]; }`
- Trigger buttons can be slotted differently to apply it to the beginning or end of Trigger Fields (use `slot="trigger-start"` or `slot="trigger-end"`)
- Promoted some attributes like `no-margins` up to IdsInput so all extended components can use them.
- Separated some functionality into new mixins `IdsFieldHeightMixin` (controls the `compact` and `field-height` settings) and `IdsLabelStateMixin` (controls `label-state` attribute).
- Replaced `label-hidden` attribute with `label-state` which supports both "hidden" and "collapsed" mode.  This is used in IdsPager, but can also be used in the page.
- Moved some existing functionality in IdsInput into existing mixins.  For example, `validate` and `validationEvents` are now defined in the `IdsValidationMixin` instead of IdsInput.  This was done with other components and their respective attributes such as Clearable, Dirty Tracking, etc.
- `IdsFieldHeightMixin` now applies to IdsTriggerButton so it takes advantage of field height natively

Closes #520 

Pull/Build/Run... then examine all of the following examples and compare them to https://main.wc.design.infor.com.  Open dev tools and try changing some attributes/props as well.  

Functionally, all components should work as they did before with minor changes to make better use of shared code. Visually, things should remain largely the same, except for some padding/margin of inline trigger buttons which is now normalized among all trigger-field components.

- http://localhost:4300/ids-button
- http://localhost:4300/ids-trigger-field
- http://localhost:4300/ids-app-menu (check Searchfield)
- http://localhost:4300/ids-date-picker
- http://localhost:4300/ids-header (check Searchfield)
- http://localhost:4300/ids-lookup
- http://localhost:4300/ids-mask
- http://localhost:4300/ids-pager (check that labels are properly collapsed on inputs)
- http://localhost:4300/ids-search-field
- http://localhost:4300/ids-spinbox
- http://localhost:4300/ids-textarea (shares some mixins)
- http://localhost:4300/ids-upload


demos/ids-input/label-state.html
```diff
@@ -0,0 +1,31 @@
...
+      label="Regular Input"></ids-input>
+  </ids-layout-grid-cell>
+
+  <ids-layout-grid-cell>
+    <ids-input
+      id="input-2"
+      placeholder="This input's label is 'hidden'"
+      size="md"
+      label="Input with hidden label"
+      label-state="hidden"></ids-input>
```
tmcconechy, line 18: Is label state hidden the same as audible?

src/components/ids-button/ids-button-attributes.js
```diff
@@ -23,13 +23,13 @@ const BUTTON_DEFAULTS = {
 // Definable attributes
 const BUTTON_ATTRIBUTES = [
   attributes.CSS_CLASS,
```
tmcconechy, line 25: Intentional removal?

deep7102 (commented):
Looks all nice,
see one as below with addition to Tim's found.

http://localhost:4300/ids-input (Clearable)
-type something then hold mouse down on X-clear button in input
-see it adding validation errors

demos/ids-input/label-state.html
EdwardCoyle: Not exactly the same thing. audible visually hides text elements and does so without regard to form/input layout. label-state specifically hides the input field label, and will either leave empty space in its place to retain the field's position in the page ('hidden') or will hide the label completely with display: none; ('collapsed').

src/components/ids-button/ids-button-attributes.js
EdwardCoyle: Yeah in this case, the attribute is being pushed up from `IdsColorVariantMixin`

EdwardCoyle:
I've reviewed the comments on this one and I believe I've fixed the majority of them.  Some I punted on, primarily things that didn't regress as a result of this PR, and moved them to new issues.

> compact size stopped working

Compact worked fine but some examples hadn't been updated to use the boolean attribute syntax (just `compact`)

> looks like the bottom margin between fields changed everywhere (should be 16px below the field to the next)

This did change everywhere because pretty much every extended component was defining its own bottom margin :) They are all definitely 16px between fields now, with the exception of `compact` and `xs` field heights, which have 12px below them (which I think is correct).  This is also all defined in one place.

> placeholder is gone (not working?)

Some of the alternate variants had stray text color rules that were causing this, should be fine now.

> for the password one looks like the caps look indicator is showing when it isnt on?

I defaulted the caps lock indicator to hidden for now and kicked this to #579 because we need something more comprehensive than what I should add as part of this PR

> might need a better focus state here (white?) on the search field

In scope for this issue and is now present

> check on the selected color isnt centered

Added this as a bullet point on #551 because it was pre-existing before this PR: https://main.wc.design.infor.com/ids-color-picker

> - the cursor when you hover/click the field changed from a pointer to a text cursor
> - when opening the field the focus state no longer goes around the whole thing like before https://main.wc.design.infor.com/ids-dropdown/
> - when i select blank on the dropdown i get two validation messages

Dropdown should be good to go on all of these now.  I had to remove `IdsValidationMixin` from the Dropdown Base, since actual validation handling is completely covered by the inner IdsTriggerField.  The only part present in Dropdown now is an observed attribute that passes the value down to the trigger field.

> ids-header
> - missing the X
> - is it also much narrower? don't hate it but wondering

both of these were due to missing settings and is now working

> open the grid and select (select not working) but i see it was like this already...

yep, we don't have a remaining features issue for lookup yet but we can cover this here since it's unrelated.  I also raised #577 for the blurriness which is unrelated.

> http://localhost:4300/ids-input (Clearable)
> - type something then hold mouse down on X-clear button in input
> - see it adding validation errors

Raised #580 for this one since it was also present in `main` -- can be reproduced on https://main.wc.design.infor.com/ids-input

danielortiz1982 (changes requested):
@EdwardCoyle This PR looks great! Good job, my only comment and concern is can you please fix the DOM elements by making them inline. Breaking the attributes into their own lines can cause some confusion. Also it's creating a mixed pattern how we keep and maintain our templates. Other than that, looks good and I approve this PR. 👍

danielortiz1982:
> @danielortiz1982 do you mean..
> 
> Do this
> 
> ```
> <ids-component setting1="sxx" setting2="sxx">
> ```
> 
> Not This?
> 
> ```
> <ids-component
>  setting1="sxx"
>  setting2="sxx">
> ```
> 
> if so i do agree some are doing it one vs another way but i think it might depend how long the line is

@tmcconechy Yes sir... Thanks, I should have explained that a little better.

EdwardCoyle:
@danielortiz1982 Forgot we discussed the formatting, it's fixed on any pages I touched now 👍 
@tmcconechy would you mind another approval?

deep7102 (approved):
Thanks! Looks nice 🍌

tmcconechy (approved):

danielortiz1982 (approved):

Issue #526, opened 2022-01-28
IdsBreadCrumb: Remaining Features/Issues


tmcconechy:
**Description**
Lists the Remaining Bugs/Features for a 100% Component

**Major**

- [ ] Add expanded [hit box](https://main-enterprise.demo.design.infor.com/components/breadcrumb/example-with-hitbox.html)
- [ ] Add [truncated feature](https://main-enterprise.demo.design.infor.com/components/breadcrumb/example-navigation-breadcrumbs.html)
- [ ] Add example for [header bread crumbs example](https://main-enterprise.demo.design.infor.com/components/breadcrumb/example-navigation-breadcrumbs.html) https://github.com/infor-design/enterprise-wc/issues/527
- [ ] Add example for [sitting in a toolbar example](https://main-enterprise.demo.design.infor.com/components/breadcrumb/test-flex-toolbar.html) https://github.com/infor-design/enterprise-wc/issues/527

**Minor**

- [ ] Add a test for click events on the links (callback example) https://github.com/infor-design/enterprise-wc/issues/527
- [ ] In RTL mode are the slashes the right direction? https://github.com/infor-design/enterprise-wc/issues/527
- [ ] In dark mode the slashes are hard to see (confirm) https://github.com/infor-design/enterprise-wc/issues/527

**Additional context**
Original Issue #208

jeffwa428:
@tmcconechy 
I have checked bootstrap RTL breadcrumb example. It seems like slash direction is correct.

![image](https://user-images.githubusercontent.com/94501186/157146801-ae029360-1e80-440e-8259-fa1001d19ee2.png)

tmcconechy:
ok thanks @jeffwa428 lets leave this for now as im not sure https://material.io/design/usability/bidirectionality.html#mirroring-elements -> mentions not doing it for icons

Asking the translator team @cpinatel7 -> in RTL mode should slashes in the breadcrumb go the opposite way...

PR #593, opened 2022-03-10
526 - IdsBreadCrumb: Remaining Features/Issues

jeffwa428:
- Added hitbox feature
- Added truncated feature
- Added breadcrumb example in header and toolbar

Closes https://github.com/infor-design/enterprise-wc/issues/526


jmacaluso711 (changes requested):
@jeffwa428 I'm seeing a few display issues:

The ids-header example is displaying in a column instead of inline
Screen Shot 2022-03-10 at 2 06 23 PM

The first item in the toolbar breadcrumb is cutoff when selected:
http://localhost:4300/ids-breadcrumb/toolbar.html
Screen Shot 2022-03-10 at 2 05 44 PM

The truncated example is removing styles after un-truncating. Here's a screencap of the issue:
IDS Breadcrumb Component

src/components/ids-breadcrumb/demos/truncated.html
```diff
@@ -0,0 +1,29 @@
...
+<html lang="en">
+<head>
+  <title><%= htmlWebpackPlugin.options.title %></title>
+</head>
+<body>
+  <ids-container role="main" padding="8" hidden>
+    <ids-theme-switcher mode="light" version="new"></ids-theme-switcher>
+
+    <ids-layout-grid auto="true">
+      <ids-text font-size="12" type="h1">Breadcrumb</ids-text>
```
tmcconechy, line 11:
```suggestion
      <ids-text font-size="12" type="h1">Breadcrumb (truncated)</ids-text>
```

tmcconechy (commented):
In addition

- can you add the links in the future to steps to reproduce (it makes it easier)
- on http://localhost:4300/ids-breadcrumb/toolbar.html can we use a different icon as it makes it seem like you can change the theme but it doesnt work (maybe settings or something)

Otherwise looks good 👍🏻

jeffwa428:
@jmacaluso711 Could you send me the ids-header example URL? I can't find it on my side.

tmcconechy:
@jeffwa428  the header example is at https://main.wc.design.infor.com/ids-header/breadcrumb.html (src/components/ids-header/demos/breadcrumb.html) - seems like it was like that but i would include a fix here.. Also add a percy test for it?

Thanks

tmcconechy (dismissed):

tmcconechy:
@jeffwa428 just needs a couple updates to the tests - for the mutation observer you just need to import it in the func test like other examples

tmcconechy (approved):

src/components/ids-breadcrumb/ids-breadcrumb.js
```diff
@@ -18,6 +30,82 @@ export default class IdsBreadcrumb extends Base {
...
+      width: elem.offsetWidth,
+    })));
+    elementsToRemove.forEach((elem) => this.removeChild(elem));
+
+    // check if there is any element to show
+    while (totalWidth < widthLimit && this.#itemsCollapsed.length) {
+      const breadCrumb = this.#itemsCollapsed.pop();
+      if (totalWidth + breadCrumb.width < widthLimit) {
+        totalWidth += breadCrumb.width;
+        this.insertBefore(breadCrumb.element, this.firstElementChild);
```
EdwardCoyle, line 87: Wondering if this DOM manipulation occurring on resize is related to the problem.  This is basically rebuilding the Breadrcrumb DOM element by element on each resize... could be causing an endless resize loop if it constantly lands on a different size.

IdsToolbar has a similar use case as the truncation called "overflow".  Instead of being calculated on every page resize, overflow is calculated as the toolbar's Actions menu is opened [on `beforeShow`](https://github.com/infor-design/enterprise-wc/blob/52e631fa169d5fd74104b9c19b2c63aa210ece8f/src/components/ids-toolbar/ids-toolbar-more-actions.js#L330) to see if Toolbar buttons can be displayed in the section without being visually "cut off".  The edge of each button is [checked against the edge of its parent section](https://github.com/infor-design/enterprise-wc/blob/52e631fa169d5fd74104b9c19b2c63aa210ece8f/src/components/ids-toolbar/ids-toolbar-more-actions.js#L458-L459) to see if the boundary is crossed.  If the boundary is crossed, the menu's corresponding item is displayed ([marked with an `overflow` attribute](https://github.com/infor-design/enterprise-wc/blob/52e631fa169d5fd74104b9c19b2c63aa210ece8f/src/components/ids-toolbar/ids-toolbar-more-actions.js#L374)).

When building Toolbar I ran into a similar problem solved with a similar DOM-based approach and I found this one to work a lot better... It's a bit faster because you're not removing/adding DOM elements that should most likely remain where they are, and just comparing number values instead.  Also it provides state (hidden vs displayed) with an attribute that could be checked in a test more accurately.

EdwardCoyle (changes requested):
@jeffwa428 This PR is great and implements the feature pretty well.  The template and API seem pretty sound.

I noticed a problem with rendering that occurs when the page is set to a certain size -- you have to drag the browser viewport's corner a bit to get it to happen, but once you do, you see something like in this video occur:

https://user-images.githubusercontent.com/3249601/159346565-9a752ec1-e840-48a2-b7bc-8dae1d95cca0.mp4

After looking at the resize code, I'm thinking its related to the code below.  I think we should take another look at this before merging and see if we can do it differently.

In addition, I was wondering if you could add a bit of documentation to the README.md for Breadcrumb about the feature once it's finalized?

src/components/ids-breadcrumb/ids-breadcrumb.js
tmcconechy: @EdwardCoyle - @jeffwa428 will be off for two weeks and unable to finish this. Im not 100% sure I'm totally following this enough. Would you be able to add a fix on for it? if its easy?

tmcconechy:
Something that could help to fix this i noticed looking at the tests for TS. Some properties need to be added but instead of adding them i just put them in the markup to simplify it...soone way could be too add the hyperlink props in the example like this:

```
    <ids-breadcrumb>
      <ids-hyperlink href="#" color="unset" font-size="14" text-decoration="hover">Breadcrumb 1</ids-hyperlink>
      <ids-hyperlink disabled href="#" color="unset" font-size="14" text-decoration="hover">Disabled Breadcrumb</ids-hyperlink>
    </ids-breadcrumb>
```

EdwardCoyle:
I'm not sure what's up with most of the visual regressions here.  The breadcrumb's positioning was changed slightly so that component's new baseline images should be good.

tmcconechy (approved):

tmcconechy (approved):
Updated with main 👍🏻

PR #594, opened 2022-03-10
520 (follow up) - Fix change event on IdsInput not to fire more than once

### LINKED ISSUE #520: IdsInput/IdsTriggerField:  Fix Inheritance and extend IdsInput

EdwardCoyle:
This PR is a minor adjustment to IdsInput that fixes a bug introduced in #569, where it was possible for IdsInput to sometimes incorrectly fire change events multiple times. 

Closes #520 

Two tests were added that cover the cases for this issue, but to manually test:

- Open http://localhost:4300/ids-input
- The first IdsInput field labelled "First Name" is listening to change events and produces a console statement when they are captured.
- Change the value of the input field and press Enter/Return.  The change event in the console should only appear once.

You can also use dev tools to trigger a change event manually on the inner HTMLInputElement, like this:
```js
$('#test-input').input.dispatchEvent(new Event('change', { bubbles: true }));
```

Doing this simulates an operation some of our input-based components do when they have to manually adjust the contents of the inner input element.  When firing the event, it should be captured and only appear in the console once.

cc @deep7102


deep7102 (approved):
Thanks!

tmcconechy (approved):

tmcconechy:
@EdwardCoyle not sure if you saw but i see a validation test failed

jmacaluso711 (approved):
Looks good  👍

tmcconechy (approved):

PR #631, opened 2022-03-31
576 - Fix dirty tracker rtl and other stuff

### LINKED ISSUE #576: IdsDirtyTrackerMixin/IdsLocaleMixin: rotate dirty tracker icon when in RTL mode
**Describe the bug**
When switching to a locale that requires right-to-left reading mode (such as `ar-EG`) many components that implement the Dirty Tracker currently do not also "flip" the dirty tracker icon to the opposite side of the input field.

**To Reproduce**
1. Go to http://localhost:4300/ids-input
2. Change the contents of the IdsInput labelled "Dirty Tracking" so the yellow dirty-tracker mark appears
3. Open dev tools and in the Element inspector, find `ids-container` underneath the body tag.  Change its `language` attribute to `ar`
4. See that the input fields all flip and the reading mode changes, but the dirty tracker icon still appears on the left side of the input field.

**Expected behavior**
RTL mode activation should cause the dirty tracker icon in all components to flip to the opposite side of the field, as well as face the opposite direction.

**Additional context**
- As of #569, IdsLocaleMixin now provides the `.container` element with an `.rtl` class if RTL is active.  We should try to use this for the rotation rules where possible.
- Note that we should try to make the CSS for the dirty tracker as generic as possible (should be able to live [here](https://github.com/infor-design/enterprise-wc/blob/main/src/mixins/sass/ids-dirty-tracker-mixin.scss#L3-L10) without causing conflicts between components).  Some components may need slight tweaks to the dirty tracker location, and the CSS for those cases should be very specific to the component they modify, living in that component's source code.

deep7102:
- Fixed dirty tracker mixin for RTL language was not working properly.
- Fixed Spinbox for disabled and all RTL styles were missing
- Fixed lookup selection was not working

Closes #576

- Navigate to:
- http://localhost:4300/ids-checkbox
- http://localhost:4300/ids-date-picker
- http://localhost:4300/ids-editor/sandbox.html
- http://localhost:4300/ids-lookup
- http://localhost:4300/ids-radio
- http://localhost:4300/ids-spinbox
- http://localhost:4300/ids-textarea
- http://localhost:4300/ids-upload
- Test for `Dirty Tracking` example
- It should add/remove yellow triangle icon
- Change to some RTL language, for example `ar`
- See the `Dirty Tracking` icon should working properly

Spinbox RTL styles
- Navigate to: http://localhost:4300/ids-spinbox
- Change to some RTL language, for example `ar`
- See all the styles, it should not misaligned

Lookup selection
- Navigate to: http://localhost:4300/ids-lookup
- Open lookup modal
- Should be able to select/unselect

### DIFF
--- a/src/components/ids-color-picker/demos/index.html
+++ b/src/components/ids-color-picker/demos/index.html
@@ -12,7 +12,7 @@
 
     <ids-layout-grid auto="true">
       <ids-layout-grid-cell>
-        <ids-color-picker dirty-tracker id="color-picker-1" label="Regular Color Picker" value="#B94E4E">
+        <ids-color-picker id="color-picker-1" label="Regular Color Picker" value="#B94E4E">
           <ids-color hex="#1A1A1A"></ids-color>
           <ids-color hex="#292929"></ids-color>
           <ids-color hex="#383838"></ids-color>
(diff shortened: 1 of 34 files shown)


tmcconechy (dismissed):

andriivarhanov (approved):
looks good, the only thing is not related to the issue is failed build. Axe tests for date picker throws "Exceeded timeout of 12000 ms for a test" sometimes, I've skipped the test for now, will revisit and find out why it's failing in the next date picker related issue

PR #651, opened 2022-04-13
565 - IdsPopup: Improve scrolling behavior

### LINKED ISSUE #565: IdsPopup: Issues with popping up / edge cases
**Describe the bug**
There are some additional missing parts of the placement calculations done by IdsPopup:

- [x] When contained by a scrollable area, the `scrollLeft` and `scrollTop` values of the area that scrolls are not accounted for by IdsPopup, causing the coordinates of the Popup to be placed too far down/right.
- [x] Nested container elements that alter position (such as with `position: absolute;`, etc) can also affect the coordinates of the Popup
- [x] Both of the items above may be specific to IdsPopups that exist inside shadow root (but may not be -- needs testing to confirm)
 
**To Reproduce**
This will exist in the Datagrid after #486 is merged

**Expected behavior**
- IdsPopup should account for scrollable parent containers and adjust accordingly
- If detection of all scrollable parents becomes too complicated or non-performant, we may need to consider a more comprehensive solution that places IdsPopup components in their own stacking context outside of the rest of the page (similar to the `attachToBody` setting we had in 4.x, but would most likely have their own container elements for better control on our part).  If this is needed, we should spec out a better issue than what's described here.

**Additional context**
@deep7102 discovered a need for this during development of #486

EdwardCoyle:
This PR fixes issues related to the scrolling behavior of IdsPopup components:
- Top/Left CSS values are now applied to the IdsPopup's host element instead of its ShadowRoot container
- Extraneous CSS rules that change position style (like some `fixed` and `absolute`) have been removed
- In some cases, `position: relative` has been added to elements that contain an IdsPopup in order to simplify the way it calculates placement
- Some parts of the JS method used for placement have been simplified

Closes #565

Pull/Build/Run the branch.  Most components should behave exactly as they used to, but will be in the correct place when scrolled.

*New test cases:*
http://localhost:4300/ids-popup/scrolling.html
- Try scrolling the the draggable container, and dragging it to move. 
- Open/Close the Popups.  They should place and behave predictably

http://localhost:4300/ids-data-grid/filter-books.html
- shrink the browser viewport to make the IdsDataGrid scrollable
- scroll it to the right and open some of the filter pickers.  When they open they should appear relatively close to their trigger elements, and the arrow should be placed correctly.

http://localhost:4300/ids-data-grid/filter-custom.html
- Open the filter pickers in the Color and Product ID columns (which are "custom" and added via slot instead of template). When they open they should appear relatively close to their trigger elements, and the arrow should be placed correctly.

*Also smoke test these pages to ensure they still work properly*:
- http://localhost:4300/ids-popup/sandbox.html
- http://localhost:4300/ids-contextual-action-panel
- http://localhost:4300/ids-modal
- http://localhost:4300/ids-tooltip


src/components/ids-time-picker/ids-time-picker.scss
```diff
@@ -32,7 +32,7 @@
     &::part(trigger-field) {
       @include inline-flex();
 
-      width: fit-content;
+      min-width: 75px;
```
tmcconechy, line 35: Is this going to impact sizes? I guess we need to see as we dont have them yet

tmcconechy (changes requested):
A couple things in order of importance


1. on http://localhost:4300/ids-dropdown the dropdown no longer seems to open at all (or its in an invisible spot?)
2. on http://localhost:4300/ids-tooltip/sandbox.html make it so you have to scroll down to the text field and then hover it (its way far up)
Screen Shot 2022-04-14 at 9 39 52 AM
3. on http://localhost:4300/ids-data-grid/filter-books.html
- if you open the datepicker its cut off
Screen Shot 2022-04-14 at 9 36 26 AM

- the time picker is too close to the icon (but doesnt do it on http://localhost:4300/ids-time-picker)
Screen Shot 2022-04-14 at 9 36 19 AM

src/components/ids-time-picker/ids-time-picker.scss
EdwardCoyle: I wasn't sure it would.  I tested the dropdown field sizes by digging into the picker shadow root on http://localhost:4300/ids-time-picker and setting the field size and it can still change (something unrelated going on with the listbox size)

tmcconechy: OK skip it for now...

danielortiz1982 (dismissed):
Great job on this PR @EdwardCoyle - This css pattern is a nice way to make sure the positioning of any element inside of the popup is a lot more reliable and more importantly predictable. 👍 
I am approving this PR under the condition that @tmcconechy comments have been addressed.

tmcconechy:
@EdwardCoyle The only things i still see are

a) datepicker is cut off 
Screen Shot 2022-04-18 at 9 44 38 AM
b) if you go in the time field and type: 1:00 AM and hit enter the timepicker opens

If you want to punt on those for now let me know and i'll approve and we can make an issue? But the datepicker one does seem related

EdwardCoyle:
@tmcconechy @danielortiz1982 I think what's left of the visual regressions here are acceptable:
- The menu-based ones are displaying "more" text than before due to some adjustments in the menu item CSS
- The standalone CSS examples for IdsAbout and IdsTooltip needed some manual correction due to the Popup changes
- Not sure about the minor shifts in size/padding on IdsEmptyMessage or IdsWeekView

tmcconechy (approved):

deep7102 (approved):
LGTM 🍌

Issue #502, opened 2022-01-21
IdsAccordion: Remaining Features/Issues


tmcconechy:
**Description**
Lists the Remaining Bugs/Features for a 100% complete component

**Major**

- [ ] Add an allowOnePage setting so only one can be open at once. [see](https://github.com/infor-design/enterprise/blob/main/src/components/accordion/accordion.js#L31)
- [ ] Add a setting to disable accordion-header similar to the [old version](https://main-enterprise.demo.design.infor.com/components/accordion/example-disabled.html)
- [ ] Add a way to expand/collapse headers programmatically as setting `$('ids-accordion-header').expanded` doesnt seem to work.
- [ ] Add an example like the [old version](https://main-enterprise.demo.design.infor.com/components/accordion/test-different-header-types.html) showing how to set an icon in the slot and hide the chevron.
- [ ] Add events to replace `selected` `expanded` and `collapsed` events.
- [ ] Update `d.ts` and `readme` settings files for all accordion settings, events and methods (like ids-tag)
- [ ] The state styling is not correct in all themes [#501](https://github.com/infor-design/enterprise-wc/issues/501)

**Minor**

- [ ] Improve new navigation methods to account for disabled/hidden/collapsed (children inside collapsed panes should become `tabIndex="-1"`)
- [ ] App-menu/Sub-App-menu variants need theme modes/versions besides New Light

**Additional context**
Original Issue #168

tmcconechy:
@jmacaluso711 can you review these. I am wondering about the "Add a way to expand/collapse headers programmatically" in case i missed something?

PR #657, opened 2022-04-18
502 accordion features fix

jeffwa428:
- Add a setting to disable accordion-header that can be seen at http://localhost:4300/ids-accordion/disabled.html similar to the [old version](https://main-enterprise.demo.design.infor.com/components/accordion/example-disabled.html)
- Add an example like the [old version](https://main-enterprise.demo.design.infor.com/components/accordion/test-different-header-types.html) showing how to set an icon in the slot and hide the chevron.
- Add events to replace selected expanded and collapsed events. With tests
- Update readme settings files for all accordion settings, events and methods (like ids-tag)
- Improve coverage as much as possible
- Improve new navigation methods to account for disabled/hidden/collapsed (children inside collapsed panes should become tabIndex="-1")
- 
Closes https://github.com/infor-design/enterprise-wc/issues/502


src/components/ids-accordion/ids-accordion.ts
```diff
@@ -47,7 +47,8 @@ export default class IdsAccordion extends Base {
       ...super.attributes,
       attributes.ALLOW_ONE_PANE,
       attributes.MODE,
-      attributes.VERSION
+      attributes.VERSION,
+      attributes.DISABLED
```
tmcconechy, line 51: Minor but keep in order

tmcconechy (approved):
Looks good to me..

tmcconechy (changes requested):
@jeffwa428 sorry i thought of one thing looking at this again. As a general rule we should use colors for disabled not opacity? So can you change the css to use the color instead of the opacity rules (and check the themes in https://main-enterprise.demo.design.infor.com/components/accordion/example-disabled.html)

tmcconechy:
@jeffwa428 you will also have to add disabled colors for the other themes. If you try dark mode and high contrast ect then then component doesnt appear disabled there

andriivarhanov (commented):
looks good, some findings:
- comparing to the original version it looks like disabled accordion panel should not have cursor pointer and focused styles. Steps:
  - go to http://localhost:4300/ids-accordion/disabled.html
  - click on chevron icon
  - see box shadow and border
image

- please change chevron icon color in dark theme, same for slotted icon 

image

src/components/ids-accordion/ids-accordion-header.scss
```diff
@@ -334,3 +418,21 @@
     }
   }
 }
+
+// :host([disabled='true']) {
```
danielortiz1982, line 422: This PR looks good so far @jeffwa428 - If these styles are not going to be used or are not needed please remove on your next commit. Thanks.

tmcconechy:
There is a small lint issue https://github.com/infor-design/enterprise-wc/runs/6158003635?check_suite_focus=true#step:7:9

tmcconechy:
Also one other thing in wondering if these handlers are needed https://github.com/infor-design/enterprise-wc/pull/657/files#diff-5ca6055f470a0a306082a01e8e65e4cac0601bca8fc275bd80a72dc8fa6cb8a5R263 -> should only need the original one... These probably never fire?

jeffwa428:
> Also one other thing in wondering if these handlers are needed https://github.com/infor-design/enterprise-wc/pull/657/files#diff-5ca6055f470a0a306082a01e8e65e4cac0601bca8fc275bd80a72dc8fa6cb8a5R263 -> should only need the original one... These probably never fire?

Expanded/Collapsed events are fired when we toggle accordion. This is one of the requirements and we also discussed it before. So we have 3 events defined. Did I misunderstand something?

tmcconechy:
@jeffwa428 no you got it right that they should fire with `triggerEvent` but i dont think they need to be watched here https://github.com/infor-design/enterprise-wc/pull/657/files#diff-5ca6055f470a0a306082a01e8e65e4cac0601bca8fc275bd80a72dc8fa6cb8a5R263-R270

Unless that does something im not understanding? So maybe just remove those highlight lines if that makes sense to you?

andriivarhanov (approved):
requested changes have been fixed, looks good, thank you

EdwardCoyle (approved):
@jeffwa428 Looks great!

PR #670, opened 2022-04-25
610 - Expandable Area features/issues

### LINKED ISSUE #610: IdsExpandableArea: Remaining Features/Issues
# Ids Expandable Area TODOs

## Major

- [  ] Styling is not the same on https://main.wc.design.infor.com/ids-expandable-area/#0 as https://main-enterprise.demo.design.infor.com/components/expandablearea/example-index.html (use ids-data-label and button icon on wrong side)
- [  ] Styling design changes on https://github.com/infor-design/enterprise/issues/5642
- [  ] Make the font size fixed size so it works on side-by-side.html
- [  ] Add events for `beforeexpand/beforecollapse` (vetoable), `expand/collapse`,`afterexpand/aftercollapse`

jdilag90:
Addresses remaining expandable area features/issues.
Adds expand/collapse events.
Refactored spacing styles to match design document. (https://www.figma.com/file/egtQemBolcaYJ5dFKjdvaS/Expandable-Area?node-id=788%3A2918)

Closes #610 

2. Navigate to http://localhost:4300/ids-expandable-area

**Check styles**
1. Check that spacing match design document
2. Toggle button expander icon should be on right side of button
3. Check that Show More/Show Less link expanders have increased hitboxes

**Check Triggered Events**
1. Test that `beforeexpand`, `beforecollapse` are vetoable and can prevent expand/collapse 
Ex. `elem.addEventListener('beforeexpand', (evt) => evt.detail.response(false));` 
4. Test that `expand`, `afterexpand`, `collapse`, `aftercollapse` are triggered.


andriivarhanov (commented):
some minor findings:
- please add more space between content and Show Less, example of spacing https://main-enterprise.demo.design.infor.com/components/expandablearea/example-custom-text.html 
image
- increase font size to 16px in expandable toggle button example content and maybe `no-padding` attribute for the button

EdwardCoyle:
@andriivarhanov as far as the spacing between content and link goes, looks like its the same way in the Figma design: 
Screen Shot 2022-04-26 at 9 00 18 AM

andriivarhanov:
@EdwardCoyle @jdilag90 oh sorry, I don't have access to figma design, compared it to the enterprise version, ignore it then

tmcconechy:
what is your infor id @andriivarhanov i can get you added

EdwardCoyle:
@jdilag I think this PR has uncovered something off in the `triggerVetoableEvents` method...  this method checks [an array of event names](https://github.com/infor-design/enterprise-wc/blob/main/src/mixins/ids-events-mixin/ids-events-mixin.ts#L115-L119) to see if it can fire a vetoable event, and isn't supposed to fire one if the event name doesn't exist.  After debugging, it looks like [this check](https://github.com/infor-design/enterprise-wc/blob/main/src/mixins/ids-events-mixin/ids-events-mixin.ts#L129) is being bypassed.

In this component the array should probably look like:

```ts
vetoableEventTypes = ['beforeexpand', 'beforecollapse'];
```

@tmcconechy think we should fix that separately?

tmcconechy:
We dont have to do it seperate - we can include the fix if we want here @EdwardCoyle
Not sure the scope of it but if its easy and fixes this PR better might make more sense to just add it here.

jdilag90:
Hey guys, thanks for the feedback! I can fix triggerVetoableEvent in this commit. I'll scan the repo for anywhere else that might be missing the vetoableEventTypes array and using that method

andriivarhanov (dismissed):

tmcconechy (changes requested):
@jdilag90 Went to approve the images but i found it uncovered that http://localhost:4300/ids-error-page is no longer working and it should look like https://main.wc.design.infor.com/ids-error-page/

Only thing i see is could it be the ids-events-mixin.ts change? Or perhaps you have another idea?

jdilag90:
@tmcconechy I reverted my events mixin changes. Would be better to make a separate events-mixin issue. The problem is as Ed described. A component with undefined or empty `vetoableEventTypes` and using `triggerVetoableEvent()` will skip this check 

`if (this.vetoableEventTypes.length > 0 && !this.vetoableEventTypes.includes((eventType as never)))`

tmcconechy:
ok @EdwardCoyle or @jdilag can you make an issue for that? I dont fully understand that issue :)

tmcconechy (approved):

jdilag90:
@tmcconechy I created a separate ticket for IdsEventsMixin triggerVetoableEvent issue https://github.com/infor-design/enterprise-wc/issues/681

EdwardCoyle (approved):
Looks good @jdilag90

Issue #622, opened 2022-03-23
IdsHierarchy: Remaining Features/Issues


tmcconechy:
**Description**
Lists the Remaining Bugs/Features for a 100% Component

**Major**

- [ ] Let the legend items be passed in as colors from the palette
- [ ] Check the styles in dark mode they seem hard to see
- [ ] Add a template or slot for the popupmenu and option to have a popup menu https://main-enterprise.demo.design.infor.com/components/hierarchy/example-stacked-ajax-popupmenu.html
- [ ] Add the ability to lazy load items when expanding https://main-enterprise.demo.design.infor.com/components/hierarchy/example-lazy-load.html (see dropdown ajax example and tooltip)

PR #672, opened 2022-04-27
622 hierarchy features

jeffwa428:
- Let the legend items be passed in as colors from the palette
- Check the styles in dark mode they seem hard to see
- Add a template or slot for the popupmenu and option to have a popup menu https://main-enterprise.demo.design.infor.com/components/hierarchy/example-stacked-ajax-popupmenu.html
- Add the ability to lazy load items when expanding https://main-enterprise.demo.design.infor.com/components/hierarchy/example-lazy-load.html (see dropdown ajax example and tooltip)

Closes https://github.com/infor-design/enterprise-wc/issues/622

### DIFF
--- a/src/assets/data/hierarchy.json
+++ b/src/assets/data/hierarchy.json
@@ -1,0 +1,48 @@
+[
+	{
+		"id": "item-1",
+		"name": "Tony Cleveland",
+    "position": "Director",
+    "employmentType": "FT",
+    "parentItem": null,
+    "isLeaf": false,
+    "color": "azure-80"
+	},
+  {
+		"id": "item-2",
+		"name": "Julie Dawes",
+    "position": "Records Clerk",
+    "employmentType": "PT",
+    "parentItem": "item-1",
+    "isLeaf": true,
+    "color": "turquoise-20",
+    "picture": "/src/assets/images/headshot-1.jpg"
+	},
+  {
+		"id": "item-3",
+		"name": "Kaylee Edwards",
+    "position": "Records Manager",
+    "employmentType": "C",
+    "parentItem": "item-1",
+    "isLeaf": false,
+    "color": "amethyst-60"
+	},
+  {
+		"id": "item-4",
+		"name": "Julie Dawes",
+    "position": "Records Clerk",
+    "employmentType": "OP",
+    "parentItem": "item-3",
+    "isLeaf": false,
+    "color": "slate-60"
+	},
+  {
+		"id": "item-5",
+		"name": "Tony Cleveland",
+    "position": "Director",
+    "employmentType": "C",
+    "parentItem": "item-4",
+    "isLeaf": true,
+    "color": "amethyst-60"
+	}
+]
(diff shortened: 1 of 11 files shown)


tmcconechy (changes requested):
@jeffwa428 overall looks good - just some fixes

- see inline comments
- the lazy loading example i would expect the JSON array to be in the example code and set to some setting. not "hard coded" inside the component because the developer using it would pass it in
- in http://localhost:4300/ids-hierarchy/lazy-load.html if you expand you get a warming about `Added non-passive event listener to a scroll-blocking 'touchstart' event`. This can be fixed by adding passive to onEvent i.e.
```
    this.onEvent('touchstart', this.leaf, () => {
      this.setAttribute(attributes.SELECTED, true);
    },  { passive: true });
  ```

File: src/components/ids-hierarchy/ids-hierarchy-item.ts [hunk 1]
```diff
@@ -171,24 +211,69 @@ export default class IdsHierarchyItem extends Base {
    * @private
    * @returns {void}
    */
-  #hasNestedItems() {
+  async #hasNestedItems() {
     const nestedItems = this.container?.querySelector('[part="nested-items"]');
-    const hasNestedItems = !!nestedItems?.assignedElements().length;
+    let hasNestedItems = !!nestedItems?.assignedElements().length;
+
+    if (this.lazyLoad) {
```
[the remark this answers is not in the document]
tmcconechy: cc @jeffwa428

src/assets/data/hierarchy.json
```diff
@@ -0,0 +1,48 @@
+[
+	{
```
danielortiz1982, line 2: This PR looks great so far. Might be Github messing with the formatting, please fix the formatting if possible. 👍

src/components/ids-hierarchy/ids-hierarchy-legend-item.ts
```diff
@@ -76,4 +66,23 @@ export default class IdsHierarchyLegendItem extends Base {
...
+   */
+  set color(value: string) {
+    this.setAttribute(attributes.COLOR, value);
+
+    let color = value;
+    if (this.color.substring(0, 1) !== '#') {
+      color = `var(--ids-color-palette-${this.color})`;
+    }
+
+    this.container.style.setProperty('--background', color);
```
danielortiz1982, line 86: Can you possibly move this and all other inline styles into a css class that you can add or remove based on the condition being true or false. We would like to minimize the use of inline styles in the ts files.

tmcconechy: The only problem with that is:

a) you would no longer be able to pass a custom color
b) would need classes for every single palette color

So given that i think this part is ok IMO.

danielortiz1982: Thanks for clearing up @tmcconechy

PR #699, opened 2022-05-03
681 - Events Mixin triggerVetoableEvent fix

### LINKED ISSUE #681: IdsEventsMixin: triggerVetoableEvent Issue
**Describe the bug**
The [triggerVetoableEvent](https://github.com/infor-design/enterprise-wc/blob/6b0bf5f2cebfaec01c8107da47a27264d44d63f8/src/mixins/ids-events-mixin/ids-events-mixin.ts#L127) method checks [an array of event names](https://github.com/infor-design/enterprise-wc/blob/main/src/mixins/ids-events-mixin/ids-events-mixin.ts#L115-L119) to see if it can fire a vetoable event, and isn't supposed to fire one if the event name doesn't exist. Currently this check is bypassed if a component using the `triggerVetoableEvent` method is missing or has an empty `vetoableEventTypes` property

**To Reproduce**
1. Place a debug point at the start of IdsEventsMixin's [triggerVetoableEvent method](https://github.com/infor-design/enterprise-wc/blob/6b0bf5f2cebfaec01c8107da47a27264d44d63f8/src/mixins/ids-events-mixin/ids-events-mixin.ts#L127)
2. Temporarily remove/comment out property [vetoableEventTypes](https://github.com/infor-design/enterprise-wc/blob/6b0bf5f2cebfaec01c8107da47a27264d44d63f8/src/components/ids-modal/ids-modal.ts#L61) in `IdsModalComponent`
3. Go to http://localhost:4300/ids-modal
4. Click on `Trigger a Modal` button
5. Notice that the [vetoableEventTypes check](https://github.com/infor-design/enterprise-wc/blob/6b0bf5f2cebfaec01c8107da47a27264d44d63f8/src/mixins/ids-events-mixin/ids-events-mixin.ts#L128-L131) is bypassed as you step through `triggerVetoableEvent()` 
6. After `triggerVetoableEvent` function exits, notice that the modal was able to open

**Expected behavior**
If `vetoableEventTypes` is missing or empty in a component (ex. IdsModalComponent), invoking IdsEventsMixin's `triggerVetoableEvent` method should return `false` and prevent show/hide type operations

jdilag90:
Fixes [`triggerVetoableEvent` method's](https://github.com/infor-design/enterprise-wc/blob/6b0bf5f2cebfaec01c8107da47a27264d44d63f8/src/mixins/ids-events-mixin/ids-events-mixin.ts#L128-L131) check for `vetoableEventNames`.
The method should return false when vetoableEventNames is undefined or doesn't contain triggered event name.

Closes #681 

1. Checkout this branch, `npm run start`
2. Go to http://localhost:4300/ids-error-page
3. Make sure error modal is able to open and close 
4. Run this in dev tools console `$('ids-error-page').triggerVetoableEvent('fakeeventname');`
5. It should return `false` since 'fakeeventname' isn't defined as a vetoableEventName


tmcconechy (commented):
Im just making a note here: because the way the error message works should be like a 404 page. The fact it uses a modal shouldn't be noticeable. 

Its like its an error page/web page. So maybe for purposes of this example can we make the button not close the modal like https://main-enterprise.demo.design.infor.com/components/error-page/example-index.html ?

What its for is for the user to go someplace else like a home page or whatever makes sense. So maybe just make it do nothing?

tmcconechy (approved):

deep7102 (approved):

PR #711, opened 2022-05-03
275 - Add manually validation and custom rules

### LINKED ISSUE #275: IsValidationMixin: Add way to manually set validation
**Describe the bug**
While we can add custom validation routines you cant just set a validation manually on the field this would be convenient. Also we can handle multiple messages. Similar to this example https://main-enterprise.demo.design.infor.com/components/validation/test-message-types.html

**Suggestion**
- Add a method `addValidationMessage({ id: 'x', type: 'error', message: 'Error occurred', icon: 'error'}) can also accept an array of messages
- Add three inputs when these are all set call `addValidationMessage`
```
<ids-input ... validation-type="error" validation-message="Error occurred" validation-icon="errror"></ids-input>
```
- when not set call `removeValidation()` this can also accept an array of message id's to remove for multiple

**Additional context**
Email

deep7102:
- Added support for manually validation messages
- Added support for manually validation messages thru HTML markup only
- Added support for custom rules
- More tests and coverage
- Update docs and other cleanup

Closes #275

- Navigate to: http://localhost:4300/ids-input/validation-message.html
- Check each message which can add/remove manually
- Messages can be add/remove as single or multiple 
- Also see example to set thru the markup
- Navigate to: http://localhost:4300/ids-input/validation-custom.html
- See examples to add/remove custom validation rules
- Rules can be add/remove as single or multiple


src/components/ids-input/demos/validation-message.ts
```diff
@@ -13,77 +14,202 @@ document.addEventListener('DOMContentLoaded', () => {
...
+      id: 'icon-custom-multi',
+      icon: 'mail'
+    }]);
+  };
+
+  // Add multiple messages
+  const addMultipleMessages = () => {
+    const addBy: string = radioValidationAddType.value;
+    // Error
+    if (addBy === 'error') {
```
danielortiz1982, line 128: This looks great @deep7102 - Just a recommendation based on preference, personally if I see I am using more then three `if` statements I kind of lean more to a `switch` pattern. Also, you can write a quick wrapper fn  since `inputMultiple?.addValidationMessage` is being used in multiple places. That will make this part of the code nice and `DRY`. Since this is a demo it's just a  recommendation.

src/mixins/ids-validation-mixin/ids-validation-mixin.ts
```diff
@@ -114,6 +130,23 @@ const IdsValidationMixin = (superclass: any): any => class extends superclass {
       if (isRulesAdded) {
         this.handleValidationEvents();
       }
+
+      // Update to remove unused rule/s
+      const arrayValidate: string[] = this.validate?.split(' ');
+      let rules = this.useRules.get(this.input);
+
+      if (this.input && (rules?.length > arrayValidate?.length)) {
+        const removed: string[] = [];
```
danielortiz1982, line 139: This looks good, however filtering inside of a `forEach` does not seem like the best approach to solve this. Also, the `Array.prototype.filter`  returns a new array no need check if `r.id === id`, `push()` or `return false`

deep7102: I think here use of filter and loop is less preference and less steps and give more clean code, other wise we have to write few extra loops/if, because we need to filter two different lists and keep track each item got filtered which make third thing to do and now we getting all three done in one so I think this is more way to go

src/components/ids-input/demos/validation-message.ts
deep7102: Update demo example now

tmcconechy (approved):
Aside from @danielortiz1982 's comments im ok to approve. I also sometimes prefer ifs to switches. But i agree that maybe creating a new array in the filter command might not be needed but will defer to you guys.

src/mixins/ids-validation-mixin/ids-validation-mixin.ts
danielortiz1982: @deep7102 I think this is fine for a first pass. Since we are up against our beta release let's keep this as is and revisit later. Good job on this PR! 👍

danielortiz1982 (approved):

Issue #587, opened 2022-03-09
IdsDraggable/Swapable: Remaining Features/Issues


tmcconechy:
**Description**
Lists the Remaining Bugs/Features for a 100% Component

**Major**

- [ ] Support mobile touch
- [ ] Support for dragging datagrid columns (is clone needed)
- [ ] Does it need keyboard support
- [x] Document: ids-drag event: e.detail.parentRect, e.detail.dragDeltaX/Y, e.detail translateX/Y
- [x] Fix known issue (minor): in firefox, the cursor when moved quickly flashes between dragging/default as the positioning/detection isn't instant as it is in chrome.
- [x] Support for dragging with handles (row reorder)
- [x] http://localhost:4300/ids-list-view/sortable.html -> use the new one
- [x] http://localhost:4300/ids-list-builder -> use the new one
- [x] compare styling with the latest https://main-enterprise.demo.design.infor.com/components/swaplist/example-index.html  (header seems too big for one thing)

**Additional context**
Original Issue #272

jmacaluso711:
Can explore using this [shim](https://github.com/justinribeiro/html5-dragdroptouch-shim/blob/master/src/dataTransfer.js) for touch capability

tmcconechy:
Closing this as stale issue

PR #720, opened 2022-05-05
587 - Sortable ListView and ListBuilder todos

jmacaluso711:
This PR refactors the sortable `ids-list-view` and `ids-list-builder` to use `ids-swappable` which implements the native drag/drop api.

related to #587 

- pull and run branch
- go to http://localhost:4300/ids-list-view/sortable.html 
- select and drag an itemt to sort
- go to http://localhost:4300/ids-list-builder
- select and drag an item to sort
- make sure all toolbar features are working (add, up, down, edit, delete)

~- [ ] A note to the change log.~


jmacaluso711:
@tmcconechy -- Looks like in enterprise you have to select first so I was going off of that. On WC main though you don't need to select first.

I didn't notice any issues with the header color, but there were a few dark theme styles that were missing, so I added those.

tmcconechy:
I just mean that this image https://percy.io/00aaa841/enterprise-wc/builds/17912965/changed/1005054429?browser=chrome&browser_ids=23&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280 (for contrast) doesnt have the same toolbar color as the actual page when contrast mode is on

andriivarhanov (commented):
it's weird Percy didn't reflect recent toolbar color change in contrast mode, maybe we should re-run the build, but also could you please check color or width of toolbar separator lines in contrast mode
image.

tmcconechy:
I think i may have approved the wrong color - if we can see find the reason its wrong

andriivarhanov (approved):
yeah, it looks like it's out of scope since the PR is about sortable functionality and there is another todo item related to styling in https://github.com/infor-design/enterprise-wc/issues/587 . Sortable functionality looks good, approved

tmcconechy:
Ok its probably a timing issue - not that worried about it right now either. Approved

Issue #660, opened 2022-04-19
IdsListBuilder: Remaining Features/Issues


tmcconechy:
**Description**
Lists the Remaining Bugs/Features for a 100% Component

**Major**

- [ ] Better keyboard support for all actions
- [ ] Can't create a new item if no items in the list initially
- [ ] Can multiselect but delete removes one at a time (select one, delete, select another, delete, select another, delete)
- [ ] Editing style needs design review
- [ ] Add same events as the prod component `move`, `selected`, `click`, `doubleclick` `change`, `add`, `beforeAdd`, `afterAdd`, `delete` `beforeDelete`, `afterDelete`
- [ ] Hover state in high contrast mode is 100% black

**Medium**

- [ ] Correct aria [see standards](https://design.infor.com/code/ids-enterprise/latest/listview#accessibility)
- [ ] A way to save order / events
- [ ] Make focus state on buttons white with box shadow
- [ ] More test coverage

**Additional context**
Original Issue #308

PR #730, opened 2022-05-16
660 - List builder features

jeffwa428:
Major
- Better keyboard support for all actions
- Can't create a new item if no items in the list initially
- Can multiselect but delete removes one at a time (select one, delete, select another, delete, select another, delete)
-  Add same events as the prod component `move`, `selected`, `click`, ...

Medium
- Correct aria [see standards](https://design.infor.com/code/ids-enterprise/latest/listview#accessibility)
- A way to save order / events
- Make focus state on buttons white with box shadow
- More test coverage

Closes https://github.com/infor-design/enterprise-wc/issues/660

- go to http://localhost:4300/ids-list-builder
- use arrow keys to move around, space to select and enter to edit


tmcconechy (changes requested):
Just a couple things i think need a further look. Also please update the steps to reproduce if i missed anything

1) In single select the editing with the keyboard is a bit strange try this:

- click the first one
- arrow down twice
- hit enter -> it edits the first one
- i think it could first select the current one and then edit it

2) when entering edit mode the size of the item changes while being edited

3) In multiselect
- select 3 items
- hit delete (not all are deleted)
- also select two
- press edit and then clear
- edit state stays around

Screen Shot 2022-05-17 at 10 39 18 AM

src/components/ids-list-builder/ids-list-builder.ts
```diff
@@ -182,7 +205,6 @@ export default class IdsListBuilder extends Base {
 
         // update inner text on keyup
         this.onEvent('keyup', i, () => this.#updateSelectedLiWithEditorValue());
-        this.onEvent('blur', i, () => this.#unfocusAnySelectedLiEditor(), { once: true });
```
jmacaluso711, line 185: I'm seeing an issue where the edit field stays up when you `blur` the input. Then, when you focus back into the input the item goes into an unselected state. I see you removed the `blur` event here which could be the cause. See below gif for recreation of the bug:

IDS List Builder Component

jmacaluso711 (changes requested):

src/components/ids-list-builder/demos/example.html
```diff
@@ -8,12 +8,21 @@
   <ids-container role="main" padding="8" hidden>
     <ids-theme-switcher mode="light" version="new"></ids-theme-switcher>
     <ids-layout-grid auto="true">
-      <ids-text font-size="12" type="h1">List Builder</ids-text>
+      <ids-text font-size="12" type="h1">Ids List Builder</ids-text>
```
tmcconechy, line 11:
```suggestion
      <ids-text font-size="12" type="h1">List Builder</ids-text>
```

tmcconechy (changes requested):
Made some changes so it is on two pages. But seeing a couple things:

1. hit new - then type then hit enter -> expect this to "commit" the edit (enter works fine on edit)
2. edit second one - then type - then click third one -> seeing edit state still there
3. after 2 cant click on the first item anymore
4. select 3rd (or any) and hit delete -> then cannot select the first one. Also it should select something like the first one so you dont need to tab back in

in http://localhost:4300/ids-list-builder/selection-multiple.html
- select first one - then move it down 4 rows -> after that try to select row 1-3 and you cant
- after dragging should remain selected

tmcconechy (commented):
Looks good i made https://github.com/infor-design/enterprise-wc/issues/753 for what you said.

I do see one thing might worth fixing and not sure if its covered by that one?

- go to http://localhost:4300/ids-list-builder/selection-multiple.html
- select the 2nd item
- hit delete - now nothing is selected

Related/Additionally tab into the page
- keep tabbing and notice you can tab to the items

I think we need to have a focus state where one item has tabindex="0" so you can move through the list and arrow up and down and select with the keyboard

We can defer this to another issue if you think?

tmcconechy (dismissed):

danielortiz1982:
This looks good so far @jeffwa428  one small thing I am noticing. If you select multiple items then add a new one seems like the api is duplicating the records instead of adding a new one. 
https://www.loom.com/share/d300199996c940639e7a3cc005dcc7bb

tmcconechy:
@Jah25 please let us know when this is ready to review again by removing the WIP flag (after addressing the additional comments from Daniel/John)

jeffwa428:
@danielortiz1982 
I think this is the expected behavior. If the user don't type anything, it will use the previous one.

tmcconechy:
@jeffwa428 did you see the video? from @danielortiz1982 ? Watching it that doesn't seem like the right behavior as it does look like it is duplicating it. Rather than use the previous value i would expect the new value to say "New Value". Maybe you just need to commit the initial value or something?

Rest looks good enough to me here and i'm willing to approve if that is fixed

tmcconechy (approved):
I see a couple issues but i think we can take this to individual issues now. So approving.
Anything else we find can just make a new issue or QA will find them.

danielortiz1982 (dismissed):

jmacaluso711 (approved):
Only seeing one minor issue we can create a new issue for:  Items are not toggleable so there is no way to unselect current item. See EP example: https://main-enterprise.demo.design.infor.com/components/listbuilder/example-index.html

tmcconechy (approved):

jmacaluso711 (approved):

PR #731, opened 2022-05-16
N/A - Remove IdsRenderLoop dependency from IdsElement

EdwardCoyle:
This PR makes a small change to IdsElement that changes a time-delayed function from using a 1ms `IdsRenderLoopItem` with `requestAnimationFrame()`.  This change results in removing a dependency on IdsRenderLoop from all IdsElements. We previously made similar changes to some individual components, but missed this one instance.

N/A

Pull/Build/Run, then smoke test components that depend on the time-delayed `rendered()` method.  Behavior of these components should be the same as `main`:

- [IdsMasthead](http://localhost:4300/ids-masthead/example.html)
- [IdsSpinbox](http://localhost:4300/ids-spinbox/example.html)
- [IdsSwipeAction](http://localhost:4300/ids-swipe-action/example.html) - test with mobile browser or responsive dev tools to make swiping work
- [IdsTabs](http://localhost:4300/ids-tabs/example.html)

### DIFF
--- a/src/core/ids-element.ts
+++ b/src/core/ids-element.ts
@@ -1,6 +1,4 @@
 import { version } from './ids-attributes';
-import renderLoop from '../components/ids-render-loop/ids-render-loop-global';
-import IdsRenderLoopItem from '../components/ids-render-loop/ids-render-loop-item';
 import { camelCase } from '../utils/ids-string-utils/ids-string-utils';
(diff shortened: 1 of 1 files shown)


andriivarhanov (approved):
looks good, no issues found

tmcconechy (approved):

PR #765, opened 2022-06-13
708 - IdsModal Additional Features

### LINKED ISSUE #553: IdsContextualActionPanel: Remaining Features/Issues
**Description**
Lists the Remaining Bugs/Features for a 100% Component

**Major**
- [x] Mobile styles (need to rework Modal for this) [#5607](https://github.com/infor-design/enterprise/issues/5607)
- [x] #728

**Minor**
- [ ] Side by side example
- [ ] Make RTL Work better (icon buttons)

### LINKED ISSUE #708: IdsModal: Remaining Features/Issues
**Original Issue:** #118 

**Major**

- [x] Add breakpoints/"fullsize" support from 4.x [(example)](https://main-enterprise.demo.design.infor.com/components/modal/example-fullsize-responsive.html)
- [x] Add `keepOpen` behavior to prevent clicking outside to close (may also just need to ignore `onOutsideClick()`)
- [x] Keep keyboard navigation within the Modal's boundaries (don't focus on anything outside the active modal) ([#267](https://www.github.com/infor-design/enterprise-wc/issues/267))
- [x] `beforeshow`(vetoable) /`show`/`aftershow` events
- [x] `beforehide` (vetoable) / `hide` / `afterhide` events

**Minor**

- [ ] Examples with CSS parts
- [ ] Trigger an event on the "target" containing a value of some sort (useful for Lookup/etc)
- [ ] Add API for disable/enable all elements within the modal (is probably tied to `disabled` property of Modal)
- [ ] Shared overlay support (multiple modals with one overlay)
- [ ] Bring over the Modal Manager from 4.x
- [ ] Create top-level containment for all modals (used by the stack)

EdwardCoyle:
This PR improves the behavior of the IdsModal/IdsContextualActionPanel and adds some additional features:
- Implements the `fullsize` setting from 4.x on IdsModal, allowing the Modal to take over the full width/height of the browser viewport with responsive/scrollable behavior in the content area
- Adds ability for Modals to respond to breakpoint changes using new IdsBreakpointMixin
- Fixes small regression in Focus Capture behavior (missing as of a recent PR)
- Adds example of how to disable `onOutsideClick` default behavior
- Adds missing event handling for `aftershow`/`afterhide` to Modals
- Fix animation bugs in overlay/modal fading and scaling

Closes #708
Closes #553 

Pull/Build/Run, then:

http://localhost:4300/ids-modal/example.html
- Open the Modal
- Shrink the browser viewport to less than 600px.  At this breakpoint, the modal becomes full size.  Enlarge the browser viewport to disable fullsize mode

http://localhost:4300/ids-modal/fullsize.html
- Note the radio buttons that represent supported IDS Breakpoints
- Choose a radio and open the Modal.  Scale the browser viewport to below the specified width.  The modal should switch to fullsize mode when below the breakpoint, and switch back to normal when moving above the breakpoint.
- Select `Never`.  The modal should never change to fullsize in this case.
- Select `Always`. The modal should always be fullsize in this case.

http://localhost:4300/ids-modal/keep-open.html
- Open the modal
- Uncheck the checkbox and click out into the Overlay area.  The modal should not close
- Check the box again and click out into the Overlay area.  Now the modal should close

http://localhost:4300/ids-modal/focus.html
- Open the modal
- Use the "Tab" key and "Shift+Tab" to navigate contents within the Modal.  Focus should never leave elements inside the Modal - it should cycle back to the first/last elements inside if necessary.

Also, smoke test the following components which are all based on IdsModal.  All should switch to fullsize mode at the 600px (sm) breakpoint by default:
- http://localhost:4300/ids-contextual-action-panel/example.html
- http://localhost:4300/ids-error-page/example.html
- http://localhost:4300/ids-hidden/example.html 
- http://localhost:4300/ids-message/example.html


src/components/ids-hidden/ids-hidden.ts
```diff
@@ -1,5 +1,6 @@
 import { customElement, scss } from '../../core/ids-decorators';
-import { attributes, Breakpoints, breakpoints } from '../../core/ids-attributes';
+import { attributes, Breakpoints } from '../../core/ids-attributes';
```
tmcconechy, line 2: breakPoints maybe (lower case?)

tmcconechy: And would it be better in ids-breakpoint-utils so in some cases might just need one import?

test/helpers/match-media-mock.ts
```diff
@@ -3,6 +3,11 @@
  *
  * @see https://jestjs.io/docs/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom
  */
+/*
+Object.defineProperty(window, 'MediaQueryList', {});
```
tmcconechy, line 7: Minor: but should this be moved up to the other comment?

tmcconechy (changes requested):
Really only some minor things. Looks really good

1. build failing with lint issue(s)
2. curious what http://localhost:4300/ids-modal/keep-open.html is for - doesnt hurt but wondering the use case
3. similar issue to https://github.com/infor-design/enterprise/pull/6567 -> do we need to change to RGBA here instead of opacity. The issue im seeing on all modal pages is if you want carefully when it loads you see the color change after a moment (not sure if this video captures it good enough https://watch.screencastify.com/v/7y2mExHq7HNNTT1r3kyZ)

src/components/ids-hidden/ids-hidden.ts
EdwardCoyle: `Breakpoints` is the TS type.  `breakpoints` is the object that contains the breakpoints.  They're both needed

EdwardCoyle:
@tmcconechy the `keep-open` test demonstrates functionality that was in 4.x that isn't an explicit setting in 5.x -- just overrides the default `onOutsideClick` behavior to prevent the modal from closing.  It's described up at the top if you want more info

src/components/ids-hidden/ids-hidden.ts
tmcconechy: What about IdsBreakpoints for the type? I think thats what we had been doing? Super minor. I think im missing the point

EdwardCoyle:
@tmcconechy I see the slight darkening you're describing, but doesn't seem consistent.  I tried replacing the usage of opacity with an `rgba()` background color instead, and it seems we're getting the same result.  Not sure what's going on there but I see it on https://main.wc.design.infor.com/ids-modal/example.html as well

andriivarhanov (commented):
looks great, nothing major, just one thing to clean up a bit

src/components/ids-popup/ids-popup.ts
```diff
@@ -1306,8 +1318,8 @@ export default class IdsPopup extends Base {
    * @returns {void}
    */
   #renderPlacementWithTransform(): void {
-    this.style.left = `50%`;
-    this.style.top = `50%`;
+    // this.style.left = `50%`;
+    // this.style.top = `50%`;
```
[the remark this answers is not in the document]
EdwardCoyle: Thanks Andrii, good catch 👍

tmcconechy (approved):
ok then aside from that everything looks good (once the checks pass)

andriivarhanov (approved):

deep7102 (approved):
Great! looks nice 🍌

PR #768, opened 2022-06-16
636 - IdsTimePicker Remaining Features

### LINKED ISSUE #636: IdsTimePicker: Remaining Features / Issues
**Original Issue:** #151 

**Major:**

- [x] Make the picker popup's dropdowns display values that reflect what's in the IdsTriggerField
- [x] Add validation passthrough to the inner IdsTriggerField
- [x] Add option to limit hours [similar to this Enterprise issue](https://github.com/infor-design/enterprise/issues/5880)
- [x] Review theme colors for accuracy (should be inherited from IdsTriggerField)
- [x] Add ARIA-attributes
- [x] Fix onOutsideClick() (ids-time-picker width too-wide, but inline-flex breaks alignment) - maybe get rid of the div
- [x] Add config for enabling/disabling leading-zeros on hours, minutes, and seconds (depends on localized date string `h` vs. `hh` or `m` vs. `mm`)
- [x] Replace the `range()` with the Date object or some other API, so we don't have to assume manually. See [Ed's comment](https://github.com/infor-design/enterprise-wc/pull/432#discussion_r756304951).
- [x] Separate out the "picker contents" to a separate, embeddable form element (could potentially be shared by this component and the Date Picker's "Date/Time" view)
- [x] Allow reset of the Time Picker's value ([#623](https://github.com/infor-design/enterprise-wc/issues/623))
- [x] `useCurrentTime` setting/attribute and functionality
- [x] Day periods AM/PM from locace calendar

**Minor:**

- [x] Add Side-by-side Examples
- [x] Add Standalone CSS Examples
- [x] Fix skipped e2e and percy tests
- [x] Add custom id's/automation id's that can be used for scripting
- [x] Add "Themeable Parts" for `label`, `input`, and `popup`
- [x] Make private `this.is24Hours`, `this.is12Hours`, `this.hasSeconds` and `this.hasPeriod` (i.e. `this.#is24Hours`)
- [x] Make private `this.dropdowns` (i.e. `this.#dropdowns()`). See [Ed's comment](https://github.com/infor-design/enterprise-wc/pull/432#discussion_r756209961).
- [x] Align center hours/minutes/seconds dropdown label
- [x] Add `:` separator between hours/minutes/seconds dropdowns

andriivarhanov:
This PR add remaining features/issues to IdsTimePicker component:
- Added validation (required, time) and validation events
- Added option to limit hours
- Added day periods from locale calendar
- Added more examples including side by side and standalone css
- Added more tests

Closes #636 

- pull and build
- go to http://localhost:4300/ids-time-picker see a list of example pages
- go to http://localhost:4300/ids-time-picker/example.html
  - check `Time Picker - required ` example
  - check required and valid time validation
- go to http://localhost:4300/ids-time-picker/sandbox.html
  - check all `Hour range ...` examples
- go to http://localhost:4300/ids-time-picker/sandbox.html
  - check first column for different custom formats
  - see leading zeros in hours dropdown doesn't appear when `H` `h` in the format
- go to http://localhost:4300/ids-time-picker/sandbox.html
  - check `Use Current Time` examples
  - dropdown values in the popup should show your current time
- check sizes example http://localhost:4300/ids-time-picker/sizes.html
- check side by side example http://localhost:4300/ids-time-picker/side-by-side.html
- check standalone css example http://localhost:4300/ids-time-picker/standalone-css.html 

~- [] A note to the change log.~


EdwardCoyle (commented):
This looks great @andriivarhanov!  

I had a small question about the behavior of the `useCurrentTime` setting -- I noticed that the time values inside the dropdowns are set on the first opening, but are not updated on subsequent openings.  Just wondering if they should update to reflect the current time as long as there isn't a "real" value set in the field.  It's a minor behavioral change, so if we should fix it, maybe something we can push to a TODO for later.

Code/tests/docs all look good to me -- good stuff 👍

EdwardCoyle (dismissed):

andriivarhanov:
@EdwardCoyle I had same thoughts about the current time values if they should be updated each time popup opened, now when you mentioned it I think it makes more sense. Should be quick fix, will do it now

EdwardCoyle (approved):

tmcconechy (changes requested):
@andriivarhanov everything is looking really good. I do see a couple smaller things.

1) On http://localhost:4300/ids-time-picker/example.html the autoselect example. I actually thought of this before and i dont get what autoselect means in this example? Does it mean auto open? Since it opens when you click it? Also in this example if i press Set Time then the picker does not close
2) Can we show a time mask on the examples on http://localhost:4300/ids-mask/example.html i noticed that i can type numbers in the time field
3) On the required example can the validation also fire on change (and does the change event fire when you click to select a time in the picker?) The issue i see (which is minor). Is if i tab out and get the validation message and then go back and click the picker and pick a value, im expecting the validation message to go away at that point but it only goes away on tab. I think this can be fixed by making the validation for on change or input event and making sure that fires when i pick a value

tmcconechy (approved):

danielortiz1982 (approved):
@andriivarhanov This looks good, great job on this PR 🚀

PR #771, opened 2022-06-22
Chore: Update dependencies and incremental build improvements

tmcconechy:
In this PR there a few things:

1. Updated dependencies from the demo the other day
2. Removed nodemon since it was "vulnerable" and was only used for initial webpack work
3. Adds back the filter concept we had before for faster build and incremental build (reloads) `npm --components=alert run start`
4. Turned off writeToDisk for modest gains (not seeing any issues with it but if anything is found this could be reverted)
5. Turned on `transpileOnly` -> this was the most significant gain. It stops the typechecker from checking types during builds and only compiles. This is done in our dev environment anyways and does report an error situation.

NA

- pull
- `npm prune && npm i` -> should be no warnings or vulnerabilities
- `npm run start` -> a bunch faster now (3x?)
- Save a file so a page reloads and recompiles -> a bit faster now  (3x?)
- For even faster speed can now use the filter again as we had before. An example command is shown in the package.json as `npm run start:filter:example`. You need to run it like this `npm --components=data-grid run start` this will only include the component your working on and other pages wont work (it excludes them from the build during dev). Save a file and its very fast now on reload in this mode.


danielortiz1982 (changes requested):
This PR looks good so far, one thing tho. Can you please make a standalone PR for the dependencies. Besides that I think everything else looks good to go. 👍

tmcconechy:
ok i can - but i felt like these were closely related...

danielortiz1982 (approved):

andriivarhanov (dismissed):

andriivarhanov (approved):

PR #791, opened 2022-07-05
feat: wc docs generator

corpulent:


scripts/docs-generator/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "name": "docs-generator",
+  "version": "1.0.0",
+  "description": "",
+  "main": "main.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
```
tmcconechy, line 11: Maybe these should be in dev-dependencies. Also wondering why its in a sub folder?

Also how do i test this?

danielortiz1982 (changes requested):
@tmcconechy @corpulent If possible update these dependencies in there own PR.

PR #800, opened 2022-07-08
Proposal: Change IdsCheckbox `checked` behavior to be more like an HTML checkbox

EdwardCoyle:
I discovered when working in the Svelte examples that when I switched some old code from an `HTMLInputElement[type="checkbox"]`  to an IdsCheckbox, the change events were no longer responding the same way anywhere a checkbox was bound.  

After some digging, I discovered IdsCheckbox was doing the following:  When using the `checked` property of an IdsCheckbox in JS or the dev tools console, it was returning `"true"` (read: string true) when checked, and `null` when unchecked.  This is different from the behavior of an HTML checkbox, which returns a boolean in either case (See MDN: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#attr-checked)

In the interests of making our components act as closely to native components as possible, I'm proposing we change the current behavior.  This PR changes the following:

- IdsCheckbox's checked property type to a boolean when accessed via its JS property.  `checkbox.checked` now returns `true` / `false` 
- When using `checkbox.getAttribute('checked')` on IdsCheckbox, the attribute will still return a string if true, and `null` if false per our current `stringToBool()` behavior.
- Tests that depend on IdsCheckbox have been updated

Related to #792 

Smoke test http://localhost:4300/ids-checkbox, and make sure all tests pass

### DIFF
--- a/src/components/ids-checkbox/ids-checkbox.ts
+++ b/src/components/ids-checkbox/ids-checkbox.ts
@@ -209,7 +209,7 @@ export default class IdsCheckbox extends Base {
     this.#triggeredChange = false;
   }
 
-  get checked(): boolean | string { return this.hasAttribute(attributes.CHECKED); }
+  get checked(): boolean { return stringToBool(this.getAttribute(attributes.CHECKED)); }
 
   /**
    *  Sets the checkbox color to one of the colors in our color palette for example emerald07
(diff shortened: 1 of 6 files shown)


deep7102 (approved):
🍌 Looks good!

danielortiz1982 (approved):

Issue #504, opened 2022-01-24
IdsActionSheet: Remaining Features/Issues


tmcconechy:
**Description**
Lists the Remaining Bugs/Features for a 100% complete component

**Major**

- [ ] Rename `btnText` to `cancelBtnText` and make it update live (https://github.com/infor-design/enterprise-wc/issues/505)
- [ ] Keyboard is a bit wierd as you need to tab to the cancel button (all arrow keys should work) (https://github.com/infor-design/enterprise-wc/issues/505)

**Minor**

- [ ] Add Tray feature to Action Sheet but this may not be needed as could be replaced by floating action buttons.
- [ ] Add visual test for having no cancel button (https://github.com/infor-design/enterprise-wc/issues/505)

**Additional context**
Original Issue #406

celhalal:
Similar features/issues mentioned in closed ticket #505

PR #822, opened 2022-07-22
505 - IdsActionSheet remaining issues

### LINKED ISSUE #505: IdsActionSheet: Cancel Button Setting and tests
**Describe the bug**
The `btnText` button could better be called `cancelBtnText`. Also if you attempt to update it after the component is initialized it doesnt do so. Could also use a test for having no cancel button.

**To Reproduce**

1. Go to http://localhost:4300/ids-action-sheet
2. In Dev tools select a mobile simulator
3. Click ...
4. in the console type: `document.querySelector('ids-action-sheet').btnText = 'test'
5. See error (button text doesnt update)

**Expected behavior**
Lets do the following:

- [ ] Rename the setting to `cancelBtnText`
- [ ] Make the update in the setter update the DOM text when called
- [ ] If it is set to '' (empty) then hide the cancel button... or a dedicated setting for `cancelBtn = false` similar to https://main-enterprise.demo.design.infor.com/components/actionsheet/test-no-cancel.html
- [ ] Add a percy UI test for no cancel button
- [ ] Fix the keyboard behavior behavior so that down arrow goes to the button as cell (and up arrow back)

celhalal:
This PR includes:
- Rename `btnText` to `cancelBtnText` and make it update live (DOM)
- Add visual test for having no cancel button (when the button is not displayed)
 - Test it by going to the Dev tools on a mobile simulator and typing in the console:
     `document.querySelector('ids-action-sheet').cancelBtnText = ''`

closes #505

1. Checkout branch, `npm run start`
2. Go to http://localhost:4300/ids-action-sheet
3. In Dev tools select a mobile simulator
4. Click ... icon in the top left corner to open the menu
5. In the console type: `document.querySelector('ids-action-sheet').cancelBtnText = 'test'`
6. See error (cancel button text doesn't dynamically update the DOM from `Cancel` to `test`)
7. To remove the cancel button, type in the console `document.querySelector('ids-action-sheet').cancelBtnText = ''`
  - Should remove the display of the button in the menu sheet.


EdwardCoyle (changes requested):
Hi @celhalal!  This is looking good so far -- just a couple things:

- Please update the steps for testing this PR to reflect the `cancelBtnText` change.  I think I tested correctly and the feature seems to work, but not totally sure of the intention of the testing steps
- I think unless there are more features coming we should not have this PR close #504.  There are some other items unaddressed here -- #505 does cover this issue so that one is fine to close.  Note that when you use `closes #[issue number]` in a Github PR, automation occurs that will actually close the issue once this PR merges.

Hopefully that helps, please reach out if you need anything

jmacaluso711 (dismissed):
@celhalal Approved  👍

EdwardCoyle (approved):
Awesome, good stuff @celhalal 🎉

test/ids-action-sheet/ids-action-sheet-percy-test.ts
```diff
@@ -0,0 +1,19 @@
...
+  it('should not have visual regressions when cancel button is not visible', async () => {
+    await page.setViewport({ width: 450, height: 800 });
+    await page.goto(url, { waitUntil: ['networkidle2', 'load'] });
+    await page.evaluate(() => {
+      (document.querySelector('#icon-button') as any).click();
+    });
+    await page.waitForSelector('ids-action-sheet[visible]');
+    await page.evaluate(() => {
+      const idsButton = document.querySelector('ids-action-sheet')?.shadowRoot?.querySelector('ids-button') as HTMLElement;
+      idsButton.style.display = 'none';
```
EdwardCoyle, line 15: Instead of hiding with CSS, I'd suggest using the same API you're using in the functional tests to hide the button, just to make the test a bit more accurate:

```suggestion
      idsButton.cancelBtnText = null;
```

EdwardCoyle (commented):
@celhalal Glad you got the test running!  I think we just need one more tweak and this PR is golden:

src/components/ids-action-sheet/ids-action-sheet.ts
```diff
@@ -88,18 +88,22 @@ export default class IdsActionSheet extends Base {
    * Set the btn text attribute
    * @param {string} val the inner text of the cancel btn
    */
-  set btnText(val: string) {
+  set cancelBtnText(val: string) {
+    const idsButton = document.querySelector('ids-action-sheet')?.shadowRoot?.querySelector('ids-button') as HTMLElement;
```
EdwardCoyle, line 92: Each of the components should be "self-contained" as much as possible... For example, we already have a reference to "this" action sheet reference, so we want to use that as the starting point for reaching the button:

```suggestion
    const idsButton = this.shadowRoot?.querySelector('ids-button') as HTMLElement;
```

src/components/ids-action-sheet/ids-action-sheet.ts
```diff
@@ -164,3 +168,10 @@ export default class IdsActionSheet extends Base {
     this.onCancelClick();
   }
 }
+
+const btn = document.querySelector('#icon-button') as HTMLElement | null;
```
EdwardCoyle, line 172: Probably want to remove Line 172 through 177 -- I'm guessing this was targeting something for a test case?  This code existing outside the component class can cause memory leaks, so better off moving it to a test page's TS if needed.

EdwardCoyle (commented):
@celhalal Was looking at this again and found a couple additional items we should address.  Other than that, this is looking great 👍

EdwardCoyle (approved):
Looks really good now!  Thanks @celhalal 👍

celhalal:
@EdwardCoyle The percy test snapshot still includes the cancel button so im going to try it in the e2e test in the meantime. Appreciate all the help and suggestions!

tmcconechy (approved):

PR #833, opened 2022-08-02
780 - ids-date-picker issues

### LINKED ISSUE #780: IdsDatepicker: Double click does not enforce Range - Min (N) Days
**Describe the bug**
1. Double click does not enforce Range - Min (N) Days
2. Cannot select date(s) on FF and Safari

**To Reproduce**

- https://main.wc.design.infor.com/ids-date-picker/range.html
1. Click on datepicker with Range - Min (5) Days (all browsers)
    - Select a date and double click it
    - See error: user is able to input a single day on datepicker with range Min 5 days
2. Click on any of the range fields (Safari & FF)
    - Select first date
    - See error: popup directly closes before selecting second date

**Expected behavior**
should only accept minimum required value
popup should not close when selecting date

**Version**

**Screenshots**
(https://www.screencast.com/t/eyUpmLeQK)

**Platform**
 - Infor Application/Team Name: [WC IDS New Web Components]
 - OS Version: [iOS 8]
 - Browser Name: [safari]

**Additional context**
None

celhalal:
This PR fixes the following issues:
1. Double click does not enforce Range - Min (N) Days
   - Fixed issue by creating an if statement to compare the start date and end date and ensure that they are not equal
2. Cannot select dates on FF and Safari since the popup automatically closes if it detects a click event
   - Updated `e.path` to `e.composedPath()` since path is being deprecated and the browsers are not responding to it.

fixes #780 

1. Checkout branch, npm run start
2. Go to http://localhost:4300/ids-date-picker/range.html on any browser
  - Click on date picker field with Range - Min (5) Days
  - Select a date and double click it
  - See error: user is able to input a single day on date picker with range Min 5 days
3. Go to http://localhost:4300/ids-date-picker/range.html on Safari/FF
  - Click on date picker field
  - Select a date
  - See error: popup automatically closes


doc/CHANGELOG.md
```diff
@@ -4,7 +4,9 @@
 
 ### 1.0.0-beta.1 Fixes
 
-- `[Action Sheet]` Updated `btnText` to `cancelBtnText` and fixed the setter update to update the DOM text when called (which was not working). ([#505](https://github.com/infor-design/enterprise-wc/pull/505))
+- `[ActionSheet]` Updated event path for clicking event and ensured appropriate date input for Range - Min(N) Days. ([#780](https://github.com/infor-design/enterprise-wc/pull/780))
```
andriivarhanov, line 7: [DatePicker]

src/components/ids-date-picker/ids-date-picker.ts
```diff
@@ -82,6 +82,8 @@ class IdsDatePicker extends Base {
 
   #triggerField = this.container.querySelector('ids-trigger-field');
 
+  #rangePickerMin = this.container.querySelector('#e2e-datepicker-min');
```
andriivarhanov, line 85: the id can't be used here since it works only for the particular example and will be different for another ids-date-picker

src/components/ids-date-picker/ids-date-picker.ts
```diff
@@ -813,6 +816,12 @@ class IdsDatePicker extends Base {
 
     if (this.useRange) {
       if (this.rangeSettings.end || (this.rangeSettings.start && !this.rangeSettings.end)) {
+        if (this.#rangePickerMin && (val.minDays !== undefined)) {
```
andriivarhanov, line 819: `if (this.rangeSettings.minDays) {` would work same way here

src/components/ids-date-picker/ids-date-picker.ts
```diff
@@ -813,6 +816,12 @@ class IdsDatePicker extends Base {
...
+          if (this.rangeSettings.start === this.rangeSettings.end) {
+            this.rangeSettings.end = new Date(this.rangeSettings.end + (val.minDays * 24 * 60 * 60 * 1000));
```
andriivarhanov, line 821: let's select days forward here like on https://main-enterprise.demo.design.infor.com/components/datepicker/example-range.html 
rangeSettings.start + minDays

celhalal: Thank you @andriivarhanov for the feedback! I updated everything  👍

andriivarhanov: looks good, thanks

andriivarhanov (approved):

src/components/ids-date-picker/ids-date-picker.ts
danielortiz1982: Why are we using 4 `if` statements? The first statement can be true, then the second can be false, then nothing happens, maybe we can change this to `if else` ?

tmcconechy (changes requested):
I noticed a wierd thing about double clicking @celhalal 

1. in chrome to to http://localhost:4300/ids-date-picker/range.html
2. click min 5 days and open the picker
3. double click aug 16 and then 11 becomes selected and then you can hit apply

tmcconechy:
I still think double clicking a day in the picker acts weird in that it selects two days. I think maybe we should try to ignore double clicks (same thing as a single click should happen).

Maybe
```
this.onEvent('dblclick',  .....
    event.preventDefault();  
    event.stopPropagation(); 
  },
);
```

tmcconechy (approved):

andriivarhanov (approved):

tmcconechy (approved):

andriivarhanov (approved):

PR #835, opened 2022-08-03
736 - Improve tooltip behavior for pie chart

### LINKED ISSUE #736: IdsPieChart: Improve Tooltip Behavior and Popup API
**Describe the bug**
The IdsPieChart's tooltip placement logic could use some cleanup and improvement. For code location see `positionTooltip` in ids-pie-chart

**To Reproduce**

1. Go to http://localhost:4300/ids-pie-chart/example.html
2. Hover the pie in various places (works ok but could be better like https://main-enterprise.demo.design.infor.com/components/pie/example-index.html)

Also i had to set the logic in two places, this could be fixed in the popup code to just set x and y.

```
    tooltip.popup.onPlace = (popupRect: any) => {
      popupRect.x = e.clientX - 45;
      popupRect.y = e.clientY - 35;
     ....
      return popupRect;
    };
    tooltip.popup.x = e.clientX - 45;
    tooltip.popup.y = e.clienty - 35;
```

**Expected behavior**
Ideally it should show in the middle of the slice https://main-enterprise.demo.design.infor.com/components/pie/example-index.html
Because I'm using dash-array this would be a different calculation using Trigonmetry to show it half way around the arc and adjust the tooltip.

~We could consider adding a setting like `popup.placement='centerarc'` in case needs to be reused~

**Expected behavior (alternate)**
Once hover end fires the tooltip can just follow above the mouse (with or without the arrow)

**Additional context**
See old logic https://github.com/infor-design/enterprise/blob/main/src/components/pie/pie.js#L523-L571

deep7102:
Improved the tooltip placement logic and some cleanup for IdsPieChart.

Closes #736

- Navigate to: http://localhost:4300/ids-pie-chart/example.html
- Hover the pie in various places
- See the tooltip should show half way around the arc (middle of the slice)


andriivarhanov (approved):
looks good

tmcconechy (approved):

tmcconechy (approved):

PR #857, opened 2022-08-11
637 - Share IdsLabelStateMixin with IdsTextArea/other components

### LINKED ISSUE #637: IdsTextarea: Share IdsLabelStateMixin features
`IdsLabelStateMixin` was recently added, and it would be more DRY for IdsTextarea to share its label functionality with the mixin (and therefore IdsInput/TriggerField/etc)

- [x] Share [`ids-label-state-mixin`](https://github.com/infor-design/enterprise-wc/blob/main/src/mixins/ids-label-state-mixin/ids-label-state-mixin.js#L46-L74) functionality with [IdsInput](https://github.com/infor-design/enterprise-wc/blob/main/src/mixins/ids-label-state-mixin/ids-label-state-mixin.js#L46-L74)
- [x] Add example page similar to http://localhost:4300/ids-input/label-state.html for IdsTextArea
- [x] Remove [duplicate code](https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-textarea/ids-textarea.js#L240-L251) and [more duplicate code](https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-textarea/ids-textarea.js#L240-L251)
- [x] Consider also sharing the [`labelRequired`](https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-textarea/ids-textarea.js#L625-L639) attribute (probably also belongs in the mixin).  Also remove duplicate code related to `labelRequired` from IdsInput.

**Additional context**
We may want to see if we can do this same thing with IdsEditor, IdsCheckbox, IdsRadio, wherever else there's a label

EdwardCoyle:
This PR:
- Extends usage of IdsLabelStateMixin, which was previously only used in IdsInput/IdsTriggerField, into other IDS components that contain a label element in their shadow root
- Moves the `label` and `label-required` attributes into IdsLabelStateMixin, and enables their usage in all consuming components
- Creates a new mixin `IdsLabelStateParentMixin`, which delegates down label state attributes from a wrapping component to an IdsInput in the wrapping component's shadow root.
- Removes/normalizes some existing settings in some components that already controlled label state in different ways (For example, removed `label-audible` from IdsCheckbox in favor of `label-state="hidden"`)
- Adds tests/documentation for both mixins

Closes #637

Pull/Build/Run.  Then test these component pages that leverage IdsLabelStateMixin. The three examples on each page should all share a similar label state:
- http://localhost:4300/ids-color-picker/label-state.html
- http://localhost:4300/ids-checkbox/label-state.html
- http://localhost:4300/ids-editor/label-state.html
- http://localhost:4300/ids-input/label-state.html
- http://localhost:4300/ids-spinbox/label-state.html
- http://localhost:4300/ids-text-area/label-state.html
- http://localhost:4300/ids-trigger-field/label-state.html

Also test these component pages that leverage IdsLabelStateParentMixin.  The three examples should all be similar:
- http://localhost:4300/ids-date-picker/label-state.html
- http://localhost:4300/ids-dropdown/label-state.html
- http://localhost:4300/ids-lookup/label-state.html
- http://localhost:4300/ids-time-picker/label-state.html
- http://localhost:4300/ids-upload/label-state.html


EdwardCoyle:
Note to reviewers: After comparing some padding sizes in Web Components to their 4.x counterparts, I realized what we used for padding between labels and their fields was 4px across the board, where it should have been 8px.  This is the reason for the sheer number of Percy tests that aren't passing.  I believe all the ones that are left are legitimate, but please double-check.

tmcconechy (commented):
@EdwardCoyle looks good. I will check the images when this runs again as well.

I do wonder if instead of LabelParentStateMixin if it is somehow possible to include the logic in the LabelStateMixin as well so its parent aware. Not a big deal, just an idea based on i had to read what the LabelParentStateMixin was doing a couple times to follow. this could be completely off base of suggestion but putting it out there.

tmcconechy:
One thing in the images we should check is did the spacing between the labels and fields change? Seems like the majority of the fields https://percy.io/00aaa841/enterprise-wc/builds/20636617/changed/1155588220?browser=chrome&browser_ids=23&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280

Should be 8px (visually) -> just want to be sure

EdwardCoyle:
Yeah I mentioned that above - the reason for all the visual regressions is because previously the padding between the fields and labels was previously only 4px.  These changes make it 8px.

The IdsLabelStateParentMixin just passes state down into an IdsInput in shadow root.  I was trying to avoid having duplicated logic in the component, since the IdsInput (or trigger field) in shadow root already has the IdsLabelStateMixin itself applied.  For the components that have the input/trigger field in shadow root, all that's really needed on the host element is a way to pass the state down.

tmcconechy (approved):
Ok i had to do something similar in https://github.com/infor-design/enterprise-wc/pull/868 and i had some problems with the name, but i needed a smaller validation mixin with just part of it... ended up calling it `IdsValidationInputMixin` as it targets the `input` element but not really happy with the name...

EdwardCoyle:
@tmcconechy @danielortiz1982 FWIW this is the problem our former IdsAttributeProviderMixin was trying to solve for... it doesn't really make sense to have these slim mixins everywhere that just pass state down if we could generalize it.

Elix has a similar solution called AttributeMarshallingMixin, which does pretty much the same thing: https://github.com/elix/elix/blob/main/src/core/AttributeMarshallingMixin.js

tmcconechy (dismissed):

tmcconechy (approved):
All good from me - but does seem to be a new test error and another conflict. Sorry to keep doing that to you :) @EdwardCoyle

tmcconechy (approved):

tmcconechy (approved):

PR #868, opened 2022-08-19
357 - Add IdsForm Components

### LINKED ISSUE #357: Forms: New Web Component/Example
Add a new ids-form component MVP https://main-enterprise.demo.design.infor.com/components/fontpicker/example-index.html

This component will do a few things. TBD

- handle form submission (`submit`)
- hold a reference array (or map) to error and dirty fields within it
- handle input specific layouts (or his might be the ids-layout-grid's job) for example https://main-enterprise.demo.design.infor.com/components/form/example-forms.html
- exhibit form examples
- handle the ability to toggle compact mode on fields within it
- Use an html `form` inside it https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit

**Possible Internal Markup**
```js
<ids-form>
 <ids-layout-grid>
 ...
    <ids-input>....
 </ids-layout-grid>
</ids-form>
```

**Suggested Dependencies**

**Design**
https://main-enterprise.demo.design.infor.com/components/form/example-index.html

**Events**
- change (ability to a field changed)
- submit

**Methods**
- submit
- dirtyFields
- validationFields

**Other**

tmcconechy:

This PR adds an initial IdsForm Component as per #357 initial code includes:

- form layouts via ids-layout-grid
- ability to set compact mode
- ability to reset dirty tracker
- ability to get invalid form fields
- ability to check form isValid
- ability to check form isDIrty
- fix compact mode on some form inputs
- Add isValid and isDirty flags to components

More can be added later as requested

Fixes #357 

**Dirty**
- go to http://localhost:4300/ids-form/example.html
- change a field like Purchase Form
- in the console check `document.querySelector('ids-form').isDirty`
- reset the tracker with `document.querySelector('ids-form').resetDirtyTracker()`
- check output. in   `document.querySelector('ids-form').dirtyFormComponents`

**Validation**
- clear a field like Purchase Form or more
- in the console check `document.querySelector('ids-form').isValid`
- check output. in   `document.querySelector('ids-form').errorFormComponents`

**Submit**
- change a field like Purchase Form
- click submit
- check the console and notice the data is showing the values, dirty state and validation state

**Compact**
- in the DOM add "compact="true" to ids-form and notice this works


EdwardCoyle (commented):
Looks great @tmcconechy !  Just a couple of questions/minor stuff:

src/components/ids-form/ids-form.ts
```diff
@@ -0,0 +1,177 @@
...
+    formAttribs += this.name ? ` name="${this.name}"` : '';
+    return `<form${formAttribs} part="form"><slot></slot></form>`;
+  }
+
+  /**
+   * Sets the compact attribute
+   * @param {boolean | string} value string value for compact
+   */
+  set compact(value: boolean | string) {
+    const isCompact = stringToBool(value);
```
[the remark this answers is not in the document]
tmcconechy: How would we share that? Way i look at it was that this has to set every one of them... Not thinking of a way we can. Really only the `set compact` `get compact` part is "not shared".

Wasn't also allowing field height tio be set at the form level only compact

tmcconechy:
@EdwardCoyle I see what you mean, i made it use the FieldHeightMixin more (and added field-height so that works now too)

EdwardCoyle (approved):

EdwardCoyle (dismissed):

deep7102:
Looks all good!
only one thing like to point about validation `isValid` return the current validation-state for each component, but I think it should check for validation also so it can use before submit the form data. 

issue see on fresh page, if you run `document.querySelector('ids-form').isValid`  it will return `isValid: true`
but if you see there are few component are not valid like textarea, timepicker, upload so I guess will be good if it have some kind of method or with same name `document.querySelector('ids-form').checkValidation()` where it will check for validation -vs- validation-state on each component.

FYI: if run [checkValidation()](https://github.com/infor-design/enterprise-wc/blob/4f805e5a1fc2f60dcb8f50a15616e0b51db0c194/src/mixins/ids-validation-mixin/ids-validation-mixin.ts#L160) on each component it will check validation and add/remove validation message.

tmcconechy:
yes good idea with `checkValidation`. I added a method that fires on submit in the example, but i only log the info on submit- check it out @deep7102

deep7102 (approved):
Looks good! Thanks, @tmcconechy

danielortiz1982 (changes requested):
Please update the package.json in a independent PR.

src/components/ids-form/ids-form.ts
```diff
@@ -0,0 +1,202 @@
...
+  get submitButton(): string {
+    return this.getAttribute(attributes.SUBMIT_BUTTON) || '';
+  }
+
+  /**
+   * Returns an array containing only IdsElements
+   * @private
+   * @returns {Element[]} Array of IdsElements
+   */
+  get idsComponents(): Element[] {
```
danielortiz1982, line 119: @tmcconechy Should we turn this into a mixin? Thoughts?

tmcconechy: Maybe later how about? I could be a util but usually I do that later when/if its ever needed a second time.

tmcconechy:
@danielortiz1982 took the package.json change out will make a new PR

src/components/ids-form/ids-form.ts
danielortiz1982: Sounds good. 👌

danielortiz1982 (approved):
This looks 🔥 @tmcconechy great job on this!

deep7102 (approved):

Issue #878, opened 2022-08-23
All: Find and fix memory leak(s)


tmcconechy:
**Describe the bug**
Action panel is exhibiting some memory leak that is proving difficult to track down.

**To Reproduce**
1. Go to test/ids-action-panel/ids-action-panel-e2e-test.ts
2. Remove skip from `it.skip('should not have memory leaks`...
3. Run `npx jest -- action-panel-e2e`

Components:

- [x] popup-menu
- [x] axis-chart
- [ ] menu
- [ ] card
- [ ] text-area
- [ ] menu button
- [ ] input
- [ ] dropdown
- [ ] action-panel
- [ ] calendar
- [ ] form
- [ ] error page
- [ ] counts
- [ ] upload
- [ ] slider
- [ ] about (modal)?
- [ ] scroll view
- [ ] accordion
- [ ] search field
- [x] counts  https://github.com/infor-design/enterprise-wc/pull/927
- [x] container https://github.com/infor-design/enterprise-wc/pull/927
- [ ] checkbox
- [ ] checkbox group
- [ ] block grid
- [ ] pager
- [ ] process-indicator
- [ ] hidden
- [ ] modal
- [ ] multi select
- [ ] action sheet
- [ ] masthead
- [ ] draggable
- [ ] editor
- [ ] swap list
- [ ] spin box
- [x] theme switcher  https://github.com/infor-design/enterprise-wc/pull/927

**Expected behavior**
Now test in the browser and see if we can find what this leak is.

**Additional context**
Use these links for troubleshooting/profiling
- https://medium.com/coding-blocks/catching-memory-leaks-with-chrome-devtools-57b03acb6bb9
- https://developer.chrome.com/docs/devtools/memory-problems/
- https://addyosmani.com/blog/taming-the-unicorn-easing-javascript-memory-profiling-in-devtools/

EdwardCoyle:
I took a look at this one and pushed some changes here: [`792-more-sveltekit-examples`](https://github.com/infor-design/enterprise-wc/commit/c24b17af61cec78c6fa83269fcb6f003d314ad4b).  The commit in the link specifically addresses the leaks only.

The gist of the changes are:
- Added `disconnectedCallback` to components and mixins in the Action Panel stack that were missing
- In these callbacks, cleaned up members of each mixin/component that needed it (`null`-ed out some items, and in some cases unbound events and observers)

[...]

PR #890, opened 2022-08-25
878 - IdsActionPanel Memory Leak Adjustments 

EdwardCoyle:
This PR attempts to fix some of the memory leak issues we were previously seeing in the IdsActionPanel example page.  This PR includes changes to the following:
- Tries to remove element references and stop observer APIs on `disconnectedCallback` in many places. 
- Moved things like ResizeObserver/MutationObserver into connectedCallback/DisconnectedCallback instead of being "always on" by way of a class property.
- Adds a chainable `disconnectedCallback` to many mixins with additional teardown routines

A combination of these things seems to bring down object counts in the corresponding e2e tests and in the dev tools memory profiler, but behavior in the components stays the same.

Related to #878 

- Pull/Build/Run
- Open http://localhost:4300/ids-action-panel/example.html
- Use developer tools to search for memory leaks ( [use these instructions if needed](https://addyosmani.com/blog/taming-the-unicorn-easing-javascript-memory-profiling-in-devtools/) ).  Nothing related to IdsActionPanel, IdsModal, any of IdsModal's inherited mixins, or IdsPopup should be present in the results.


src/core/ids-element.ts
```diff
@@ -72,10 +72,12 @@ export default class IdsElement extends IdsEventsMixin(HTMLElement) {
    * in a component you can just call super.
    */
   disconnectedCallback() {
+    /*
```
tmcconechy, line 75: Could remove this but looks 👍🏻

tmcconechy (approved):

PR #892, opened 2022-08-26
861 - Add support for router link

### LINKED ISSUE #861: IdsDataGrid: Add option to override hyperlink href attribute
When using the datagrid in certain frameworks the user may want to override the default use of `href`. For example, Angular makes use of a custom attribute `routerLink` to initiate navigation.

Add an option to `ids-hyperlink` or the `ids-data-grid` hyperlink formatter to override the use of `href`

n/a

**Additional context**
n/a

tmcconechy:

This should be reviewed by @jmacaluso711  / angular experts. As im not sure about routerLink. https://github.com/infor-design/enterprise-wc/issues/861#issuecomment-1228923805

Closes #861 

- go http://localhost:4300/ids-data-grid/columns-formatters.html
- inspect the column Location With RouterLink
- since routerLink is a string processed by JS is this going to work in a real scenario?
- go to http://localhost:4300/ids-hyperlink/example.html
- this should work and `routerLink` will set the href which will feed down to the link

### DIFF
--- a/.htmlhintrc
+++ b/.htmlhintrc
@@ -1,6 +1,6 @@
 {
   "tagname-lowercase": true,
-  "attr-lowercase": ["viewBox"],
+  "attr-lowercase": ["viewBox", "routerLink"],
   "attr-value-double-quotes": true,
   "doctype-first": false,
   "tag-pair": false,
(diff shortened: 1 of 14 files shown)


EdwardCoyle:
Is there a way to test this with the Angular examples?  I think the code looks OK, but not sure of the test case

tmcconechy:
@EdwardCoyle i was hoping @jmacaluso711 would do that as he requested this. But im not sure how the grid routerLink might work (like if the routes are dynamic or something)

jmacaluso711:
@tmcconechy @EdwardCoyle After reviewing this it seems this may not be the best approach. Since `routerLink` is a directive, if it's added in our library it won't be processed by Angular and will just be a string. Best approach might be to add a listener for click on the Angular side and use the router to navigate. Just need to make sure to document it

tmcconechy:
Ok i wasnt sure either but didnt seem 100% right. I will close it

PR #913, opened 2022-09-07
897 - Replace IdsRenderLoop with Timer Utils and async/await

### LINKED ISSUE #897: Replace IdsRenderLoop with timeout/interval utilities
The IdsRenderLoop global instance is causing memory leaks in some cases.  Additionally, it's becoming clear that we don't need an "always-on" animation loop to handle IDS animation requirements.

We should be using async/await and waiting for specific conditions where possible.  However, in cases where we need specified time delays, we can replace the usage of IdsRenderLoop with simpler functions that are not global, but can still make use of requestAnimationFrame loops that are of finite-length:

- [x] We can make a small library/util containing rAF-based timeout/interval functions, and only use them when needed.
- [x] We should review each usage of IdsRenderLoop and see if non-rAF based solutions are possible

Some examples:
- https://gist.github.com/joyrexus/7304146
- https://medium.com/trabe/implementing-settimeout-using-requestanimationframe-20cc2f6e6b5d

**Additional context**
This is needed to complete #878

EdwardCoyle:
This PR:
- Removes any dependency on IdsRenderLoop in favor of async/await, CSS animations, and a new set of timer functions that behave similarly to `setTimeout/setInterval` and use `requestAnimationFrame` internally (instead of a global loop instance).  As a result this PR also completely removes the render loop from the library.
- Adds tests/docs for new features
- In IdsToastMessage, replaces the progress bar behavior (previously driven by IdsRenderLoop) with a combination of CSS animation and waiting for `animationend` events

Closes #897 
Related to #878 

- Pull/Build/Run this branch
- On http://localhost:4300/ids-button/example.html  - make sure button animation (ripple effect) starts and completes as expected.
- On http://localhost:4300/ids-modal/example.html - make sure the modal still opens and closes, and animation appears correct.
- On http://localhost:4300/ids-scroll-view/example.html - Use dev tools to check this page on mobile view, and try to scroll left/right on the image.  Behavior should remain as expected.
- On http://localhost:4300/ids-swipe-action/example.html - Use dev tools to check this page on mobile view, and try to swipe the left/right edges to activate actions.  Make sure this doesn't produce "passive" event errors in the console.
- On http://localhost:4300/ids-tooltip/sandbox.html - Use dev tools to check this page on mobile view, and press/hold the mouse button over one of the top four buttons (labeled "Tooltip on Top", etc) to simulate a longpress action.  After about 1000ms, the tooltip should be displayed (uses the new timing functions underneath)
- On http://localhost:4300/ids-toast/sandbox.html - Try all toast combinations (especially delay change) and ensure Toast behavior is correct.
- New functional tests for the timer utils should pass
- New docs for timer utils should be accurate
- Use steps outlined on #878 to test for IdsRenderLoop-related memory leaks, and to check new timing functions for memory leaks.  None should be present.


danielortiz1982 (changes requested):
Can you possibly refactor all of the parts that contain ` requestAnimationTimeout` so we can removed them. There is a few cases were I feel we really do not need them.

tmcconechy:
@EdwardCoyle looks good. Aside from the idea that maybe we can remove some of the cases that use this (i'm suure not all).

I can confirm:

a) It fixes the leaking issue when not even adding anything
b) seems to help a bit with memory leaks in general when i branched off this
c) things seem to be still working

Will need to also check the test output

EdwardCoyle:
Fixed some of these:
- Ripple mixin now waits for the end of its ripple's CSS transition to remove
- Got rid of the requestAnimationTimeout  in IdsScrollView and reworked the detector to not need any timeouts

I'm not sure how we can remove timeout-based stuff from the following.  If you guys have any ideas on how we could do these differently, we should talk:
- IdsModal - Using the CSS-driven approach for now (doesn't use a `rAF` but still "time"-based via `transitionend` event).  This timeout is here for accessibility-reasons -- it allows the user to see the reaction of the Modal buttons' pressed state before the modal is dismissed, and was purposefully added.  I'm not sure how to get away from an arbitrary timeout here
- IdsEventsMixin - all "timeouts" here are related to longpress/swipe gestures, and need to be cancellable.  For example, lifting your finger off the longpress early would cause the longpress to cancel and cease running. We can't use the CSS animation approach for this, and I'm not sure there's a way outside of these timing loops to do this -- seems like the right use case.
- IdsSwipeAction - don't quite understand why we can't just scroll this to the correct place on `connectedCallback`.  I used the CSS-driven approach here for now to stagger the timing in a way that scrolls the container to the correct spot without animating on open.  I'm not sure how to fix this one otherwise.

Just noting we should also recheck the memory leak problem once we settle on these changes

tmcconechy (approved):
Nice work @EdwardCoyle 

IdsModal - I think its a valid use case here too. We need it in this case.
IdsSwipeAction - I also dont know a better way to fix this. I fought really hard with this. The problem is you need to scroll beyond the first button initially. I like your new solution it works better than it did before and no RAF. So lets just go with this. The only thing i can think of is that we remove the example with a button on the left (its less common)

Tests will need fixing - but tentatively approving

tmcconechy (approved):
Just needs sync with main - but good for me 👍🏻

tmcconechy (approved):

danielortiz1982 (approved):

Issue #900, opened 2022-08-31
General: Review IDS Components for XSS Vulnerabilities


EdwardCoyle:
**Describe the bug**
This issue is for performing a general check on all Web Components for mitigation of XSS attacks.  Any component that:

- accepts user input
- has string-based attributes that are stored on the host element

... should be double checked for XSS vulnerabilities, and should have test cases added demonstrating XSS attack mitigation.

**Additional context**
- Related to #899 
- Audit checklist [available here (requires Infor login)](https://inforonline-my.sharepoint.com/:x:/r/personal/ed_coyle_infor_com/Documents/IDS%20XSS%20Vulnerability%20Audit.xlsx?d=w8bcc37c9ffd24d36b97cc67950bd8b2a&csf=1&web=1&e=DsRuSx)
- [Veracode Analysis of the current WC library (requires Infor login)](https://analysiscenter.veracode.com/auth/index.jsp#ViewReportsDetailedReport:47220:435900:20984967:20956725:20972375:::::2904602) that demonstrates no current known XSS vulnerabilities.  However, we are still proactively addressing any potential bad input as described above.
- [Great article from Veracode on XSS attack mitigation](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.md)
- [OWASP article on how to exploit XSS problems](https://cheatsheetseries.owasp.org/cheatsheets/XSS_Filter_Evasion_Cheat_Sheet.html)

tmcconechy:
Will take each issue as a seperate one as it comes. Initially all seems good on a base layer

PR #940, opened 2022-09-28
900 - Add guards against XSS vulernabilities in some components

EdwardCoyle:
This PR addresses some potential XSS vulnerabilities in the IDS WebComponents by preventing unwanted input in class methods and setters of some components:  This PR also adds a functional test helper that performs a set of assertions against a handful of known XSS attack strings that can be run against setters in test suites.

Partially addresses #900 

Pull this branch, build, and run.  Then, review these components -- no behavioral changes should be present, but all test suites should pass:
- http://localhost:4300/ids-accordion/example.html
- http://localhost:4300/ids-action-panel/example.html
- http://localhost:4300/ids-action-sheet/example.html
- http://localhost:4300/ids-popup/example.html


tmcconechy:
@EdwardCoyle looks good. Just a few tests failing. Also if you do see a memory test fail feel free to skip it for now.
I do wonder one thing, though I thought we determined that using innerHTML ect it wasnt needed to strip tags out like here [6686-searchfield-tabsmod](https://github.com/infor-design/enterprise-wc/pull/940/files#diff-7f358f35fcc210f9b477522397700e7594c19c8315675a366da0d4e431296f44R104)

EdwardCoyle:
@tmcconechy you're right, it doesn't have to do this for `innerHTML`.  In this case I'm just re-using the same ref that I already sanitized for use in `setAttribute`

src/components/ids-about/ids-about.ts
```diff
@@ -122,7 +123,7 @@ export default class IdsAbout extends Base {
    * @param {string} val productName attribute value
    */
   set productName(val: string) {
-    this.setAttribute(attributes.PRODUCT_NAME, val);
+    this.setAttribute(attributes.PRODUCT_NAME, stripHTML(val));
```
tmcconechy, line 126: To further my question.  i still wonder if this is necessary at all becuase:

1) Would stripHTML would make it impossible to add a productName like `Product X <Version 1.0>`

Also it doesn't really cause problems on setAttribute if you do..

```
$('ids-data-grid').setAttribute('tree-grid', `<script>alert("sup");</script><span>Component</span>`)
```

Also another case is 

```
$('ids-data-grid').setAttribute('tree-grid', `<div onclick="alert()"></div>`)
```

But neither seem to cause any problems that would require stripping <> from the attribute? I think these are already protected in plain JS?

EdwardCoyle: In some further research, it looks like `setAttribute` doesn't execute its contents the way we'd originally thought... so the action of storing one of these "suspect" strings as an attribute isn't enough to trigger an XSS attack alone. 

I guess my question would be -- is it ever appropriate to store HTML strings inside of attributes in this way?  It seems like it shouldn't be allowed or even an appropriate use of an attribute to me.

tmcconechy: I guess you never know. But i know often a text string may contain < or > so i think there it could cause issues.
But seems to me if it isn't a vulnerability then no reason for the extra code on every attribute..

tmcconechy:
Added the tag - i think this is still WIP?

EdwardCoyle:
I'm closing this PR for now... it seems [after some research](https://cheatsheetseries.owasp.org/cheatsheets/XSS_Filter_Evasion_Cheat_Sheet.html) that only specific DOM attributes are susceptible to these kinds of attacks.  The custom attributes we are creating for IDS Components don't appear to be affected the same as `on***`, `src`, and `href` attributes... therefore the work done here specifically for attributes wouldn't apply.

In addition, we recently had a Veracode scan against this code base that turned up zero XSS issues with what's currently in `main`.  I think we may still want to consider this issue for specific input, methods, etc... but after digging in further, I'm thinking an audit of every single setter/method/etc is probably not necessary.

PR #949, opened 2022-09-29
650 - General types fixes

### LINKED ISSUE #650: Types: Make Components Have Correct Types
**Describe the bug**
Try to make the types for elements have the right casting to types and work in certain situations with types.

**Problem One**
Edit any example like src/components/ids-about/demos/index.ts to something like:

```
import IdsAbout from '../ids-about';
const about: IdsAbout  | null = document.querySelector('#about-example');
// Type 'Element' is missing the following properties from type 'IdsAbout': connectedCallback, template, attachEventHandlers, #attachCloseButton, and 14 more.
```

or

```
const about = document.querySelector<IdsAbout>('#about-example');
// Type 'IdsAbout' is missing the following properties from type 'Element': attributes, classList, className, clientHeight, and 160 more.
```
**Problem Two**
A component inheriting from another. IdsMessage inherits from IdsModal and Mixins. How can we make all the methods and props in both the mixin and modal be visible inside the IdsMessage child. In the types

**Expected behavior**
How can we construct the code so this works in both cases?

**Additional context**
Would be nice to have - bounty 🎁

jdilag90:
Fixes types in the component library. Used combination of generics and interfaces to replace 'any' types in the mixins. Fixed/Added types to the component classes and some func tests.
There are still some scattered `any` types within functions, but our component classes should at least be recognized/typed as `IdsElements` , `HTMLElements` and whatever Mixins that are listed in `Base`

Closes #650 

2. Running `npx tsc` should not throw any errors
3. Go to http://localhost:4300/
4. Go through component demo examples and compare to https://main.wc.design.infor.com/ 
5. Try out the types within the codebase (ex. Inside `class IdsTag`, doing `this.onEvent()` or `this.querySelector()` should be statically typed with tooltip intellisense inside VSCode)
6. If possible, try testing this branch in our wc framework example projects (angular, react, svelte, etc.) to check that types are working


jmacaluso711 (changes requested):
Tested this in the Angular examples and noticed 2 types of TS errors, TS2339 and TS2611

The first one (TS2339) can be resolved on the NG side by removing unnecessary imports of the enterprise components. For example, assigning the enterprise component to the ElementRef in Angular causes the error because `nativeElement` does not exist on our web components:

```
import { Component, AfterViewInit, ViewChild, ElementRef } from '@angular/core';
import IdsModal from 'ids-enterprise-wc/components/ids-modal/ids-modal';

... 

export class ExampleComponent implements AfterViewInit {
  @ViewChild('modal', { read: ElementRef }) modal: IdsModal;
  
   ...
   
  handleShow() {
    this.modal.nativeElement.show();
  }

}
```

TS2611:  Seems like a bulk of these errors are being thrown due to a naming conflict between our setters and native attributes, events or elements. (input, id, label, scrollTop, value, hidden, tabIndex, etc)

In the case below I was able to resolve the error by changing the name of the setter in `IdsVirtualScroll` from `scrollTop` to `scrollTopPos` 

```
Error: node_modules/ids-enterprise-wc/components/ids-virtual-scroll/ids-virtual-scroll.d.ts:82:9 - error TS2611: 'scrollTop' is defined as a property in class '{ handledEvents: Map<any, any>; longPressOn: boolean; swipeOn: boolean; keyboardFocusOn: boolean; keyDownEndOn: boolean; hoverEndOn: boolean; isClick: boolean; timer?: FrameRequestLoopHandler; ... 326 more ...; focus(options?: FocusOptions): void; } & IdsElement', but is overridden here in 'IdsVirtualScroll' as an accessor.

82     set scrollTop(value: number | string | any);
```

It's possible there is something else going on because there are other setters and getters throwing the same error. For example:

```
Error: node_modules/ids-enterprise-wc/components/ids-input/ids-input.d.ts:101:9 - error TS2611: 'fieldContainer' is defined as a property in class '{ connectedCallback(): void; handleTooltipEvents(): void; readonly toolTipTarget: any; showTooltip(): void; tooltip: string; onEvent(eventName: string, target?: any, callback?: (evt: any) => void, options?: EventOptions): void; ... 309 more ...; focus(options?: FocusOptions): void; } & ... 12 more ... & IdsElement', but is overridden here in 'IdsInput' as an accessor.

101     get fieldContainer(): HTMLElement | undefined | null;
```

All errors are coming from our `d.ts` files.

tmcconechy (changes requested):
Looks fantastic. I only saw two bugs/issues

http://localhost:4300/ids-scroll-view/example.html#id-0
- scroll left and right (i use a mac magic mouse)
- dot does not update
- see https://main.wc.design.infor.com/ids-scroll-view/example.html

http://localhost:4300/ids-locale/example.html
- has an error(s) in the console

Let me know if you need any help with those

jdilag90:
@jmacaluso711 thanks for the feedback

Regarding the Typescript TS2611 errors, 
Issue is described here
https://github.com/microsoft/TypeScript/issues/41347
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#properties-overriding-accessors-and-vice-versa-is-an-error

Basically if a parent property's type is defined like this
`id: string`

TS will complain when a child class overrides `id` with accessors
`get id()`
`set id(value: string)`

I'm working on a work-around for our non-native attribute setters/getters. (ex. get fieldContainer, get dirtyTracker).

Still not sure how to get around the native attributes/properties such as `id`, `title`, `hidden`.
Unless, we do what you did and change the naming (ex. scrollTop -> scrollTopPos)

tmcconechy (dismissed):
turns out my two issues were already issues so made new issues and fixed on locally. Approving this.

danielortiz1982 (dismissed):

tmcconechy (approved):
Is this all good now @jdilag90 ? cc @jmacaluso711

jdilag90:
Hey guys,

I addressed the type TS TS2611 errors. (`'scrollTop' is defined as a property in class....but is overridden here in 'IdsVirtualScroll' as an accessor.`)

The types in angular need to be changed from 
`@ViewChild('myTag', { read: ElementRef }) tag: IdsTag;`
to
`@ViewChild('myTag', { read: ElementRef }) tag: ElementRef<IdsTag>;`

`this.tag.nativeElement` should be typed to the component now

src/components/ids-about/ids-about.ts
```diff
@@ -105,41 +105,40 @@ export default class IdsAbout extends Base {
...
 
     return `${appName || ''} ${this.productName || ''} ${this.productVersion || ''}`;
   }
 
   /**
    * @returns {string} productName attribute value
    */
-  get productName(): string {
+  get productName(): string | null {
     return this.getAttribute(attributes.PRODUCT_NAME);
```
clayinfor, line 117: should we have this return an empty string instead of null?
`return this.getAttribute(attributes.PRODUCT_NAME) || '';`

jmacaluso711 (approved):
Looks great @jdilag90

deep7102 (dismissed):
Great! work @jdilag90

danielortiz1982 (approved):
@jdilag90 Great job on this PR. This is a really big one. 👍

Issue #667, opened 2022-04-25
IdsPopup/IdsDataGrid: Popup is cutoff by Data Grid containment


EdwardCoyle:
**Describe the bug**
The current design of our Popups/Data Grids -- specifically containment and nesting of popups inside of components such as Date and Time Pickers -- is causing a tough-to-fix containment bug where the Popups appear "cut off" by the boundaries of the Data Grid:

![Screen Shot 2022-04-25 at 12 13 37 PM](https://user-images.githubusercontent.com/3249601/165130204-4c31b48f-154d-48da-bd9c-20bb4dd47de2.png)

**To Reproduce**
1. Go to https://main.wc.design.infor.com/ids-data-grid/filter-books.html
2. Click on the Date Picker in the "Pub. Date" column header
3. When the Popup opens, notice it's cut off by the bottom of the data grid

**Expected behavior**
The Popup should be completely visible (should break out of Datagrid's overflow).  How we fix this isn't clear because it needs some exploration, but some ideas could be:

- We may need to make it possible for pickers/trigger fields to set a "context" element outside of their own boundaries where the Popup element physically exists.  In 4.x we had an `attachToBody` setting for many pickers that would place Popups directly in the body tag.  The downside to this would be Popups wouldn't have the same context for scrolling that the Picker/Trigger Field elements have, causing placement desyncs.
- Maybe change how Data Grid overflow works (but this may have other impacts on Data Grid that I'm not sure of)

**Additional context**
- Somewhat related to #648
- Discussed this on [#651](https://github.com/infor-design/enterprise-wc/pull/651#issuecomment-1101421056)

EdwardCoyle:
In order to get the Date and Time pickers to follow the same strategy used to fix IdsPopupMenu in #954, we need to restructure the way they are containing their menus.  This work is outlined in #958 and is required to finish this issue.

tmcconechy:
Also M3 reported a missing popup menu - but im not sure how they got it to pop on top. And i tried this and i didnt hit the issue. But mentioning here.

![image](https://user-images.githubusercontent.com/814283/201369282-cd0473d8-05b6-4b94-bfb3-2a6a2e8c76fb.png)

glenlieorillo:
The bug still exists.


https://user-images.githubusercontent.com/102151352/211235198-f2012187-b0de-4960-96f3-17624d298b52.mov

Also, a broken image appeared before _Canada_ under **Location** column

[...]

PR #954, opened 2022-10-04
667 - Fix containment problems with IdsDataGrid filter menus

EdwardCoyle:
This PR adds functionality to IdsPopupMenus that enables their attachment to other elements in the page.  This allows them to be created/appended near their trigger element, but allows them to move around in the DOM and retain their connections to that trigger element:
- A new mixin `IdsAttachmentMixin` has been added which contains the append/revert functionality
- Changes have been made to IdsDataGridFilter/IdsMenuButton to accept the new behavior (menus connected to trigger buttons that are not necessarily adjacent to those trigger buttons)
- Added docs and fixed related tests

Note that this PR doesn't fix the Timepicker/Datepicker fields in filter rows.  Additional work is needed to completely fix the cutoff problem in those specific Popups.

Partially addresses #667 

Pull/Build/Run, then test the following cases:

"Built in" filters (located in shadow root)
- Open http://localhost:4300/ids-data-grid/filter.html
- Use the Pager's "number of records" button to switch the number of records to 5
- Open the filter menu attached to the "Color" column.  It should escape the data grid's boundaries and be usable.

Custom filters (located in slots/light DOM)
- Open http://localhost:4300/ids-data-grid/filter-custom.html
- Use the Pager's "number of records" button to switch the number of records to 5
- Open the filter menu attached to the "Product Id" column.  It should escape the data grid's boundaries and be usable.

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -16,6 +16,7 @@
 - `[DataGrid]` Fixed issues with virtual scroll and selection (including event data) and keyboard. ([#737](https://github.com/infor-design/enterprise-wc/pull/737))
 - `[DataGrid]` Added checkbox and custom formatter. ([#737](https://github.com/infor-design/enterprise-wc/pull/737))
 - `[DataGrid]` Changed alternate row shading feature's color theming to behave similarly to the 4.x DataGrid's. ([#925](https://github.com/infor-design/enterprise-wc/issues/925))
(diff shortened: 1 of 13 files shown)


tmcconechy (approved):
Looks great! The only thing i saw (just tell me if you think its out of scope). Is on http://localhost:4300/ids-data-grid/filter-custom.html when you open the color filter popup its not "checked" in the menu like it is on http://localhost:4300/ids-data-grid/filter.html

EdwardCoyle:
@tmcconechy I didn't change anything in here that would've affected the custom filters in that way... It probably should check the menu item but I wouldn't want to mess with that as part of this issue

PR #964, opened 2022-10-18
852 - IdsAppMenu make main example consistent with 4x

### LINKED ISSUE #852: IdsAppMenu: Header is not consistent with 4x
**Expected behavior**
Ids App Menu example should look like 4x.

**Version**

**Screenshots**
Screen Shot 2022-08-10 at 1 46 50 PM

**Additional context**

andriivarhanov:
This PR makes `ids-app-menu` main example look like 4x.

Closes #852 

- pull and build
- go to http://localhost:4300/ids-app-menu/example.html
- compare with https://main-enterprise.demo.design.infor.com/components/applicationmenu/example-index.html


tmcconechy (changes requested):
@EdwardCoyle do we have this ability yet? If so we might need to expand the scope a bit here? For discussion:

@andriivarhanov I think we need to make the menu "push" over a container the header is in like https://main-enterprise.demo.design.infor.com/components/applicationmenu/example-index.html when you open and close.

 I.E. button should stay visible when open. Might need to come up with an approach for that. But its sort of key to this example

Also can you add a change log in the new section https://main-enterprise.demo.design.infor.com/components/applicationmenu/example-index.html (and remove the extra s in the 1.0.0-beta.3s Fixes if you dont mind?)

andriivarhanov:
@tmcconechy made the ability for the menu to push the container in the example, let me know if it should be in the component

EdwardCoyle:
Hey @andriivarhanov, a few things about this one we should consider:
- The app menu will eventually be resizable via a splitter (larger/smaller than 300px) and the rest of the "contained" area should be able to resize to fill the remaining space
- The "contained" area is more than just an ids-header in the majority of cases.  Usually will contain an entire page.
- In 4.x, one of this feature's requirements was that it was also tied to a breakpoint, and the page shifting would only happen above that breakpoint.  To see the difference: shrink your browser to a small width and compare this example to an open App menu on https://main-enterprise.demo.design.infor.com/components/applicationmenu/example-open-on-large.html. The behavior we had before would still occur due to size constraints, but would shift when above that constraint.

**As far as structure is concerned:**  I think the IDS container might be the right place for slotting the App Menu, but I'm unsure of the best way to keep the entire page layout flexible.  Even though it's not currently a requirement, I'm thinking we probably should allow for different drawer/menu/aside elements to be pluggable into the container that might need to adjust the page in different directions, but all the "main" slotted page elements should shift away from those elements.  For example, we also have the Action Sheet component which pushes page contents "up" to make room for an occasional, touch-friendly context menu.  I can see teams coming to us and asking us to implement this page shifting feature with that component for the same reasons.  

Therefore, I think what might work best is to specify named slots for menus/asides in IdsContainer, and any elements in the main slot would be "shiftable" the way the header is currently.  Maybe something like:

```html
<ids-container>
  <ids-header></ids-header>
  <ids-layout-grid></ids-layout-grid>
  
  <ids-app-menu slot="app-menu"></ids-app-menu>
</ids-container>
```
... where the first two elements are placed in a default slot that is "size-controlled" by some wrapping element in shadow root:

```html
<ids-container>
  #shadow root:
    <div class="ids-container" part="container">
      <slot name="app-menu"></slot>
      <div class="ids-container--contents" ${this.scrollable === 'true' ? ' tabindex="0"' : ''}>
        <slot></slot>
      </div>
    </div>
</ids-container>
```

I don't think I'm totally tied to this design as long as we keep things flexible -- so open to suggestions?

andriivarhanov:
@EdwardCoyle thanks for the helpful info, I will make the change and add some content to the page so we see how it works

andriivarhanov:
Added the "pushing" functionality to the component, only css is for the container, also added a content to the container to showcase how it works with the content in it. Fixed button focus shadow is cutting off. Ready for review

tmcconechy (changes requested):
Looks good - but did you intend to add the typography to that page? I feel like the page should just be empty as it sort of mixes interests on the example. So maybe take it out and leave the page body empty?

Thanks!!

andriivarhanov:
removed the content from the page

tmcconechy (dismissed):

tmcconechy (approved):

EdwardCoyle (approved):

PR #1000, opened 2022-11-16
989 - IdsDataGrid: Allow custom selected cell colors

### LINKED ISSUE #989: IdsDataGrid: Show a custom background color and custom selection color
**Task**

M3 has the ability to customize row and cell color. We can do this via the cssPart feature as shown on example https://main.wc.design.infor.com/ids-data-grid/columns-custom-css.html

However, if they add selection the row selection color also needs to be customizable. So extend this example to show selection and add another setting(s) to set the custom color of the selected rows with another css part.

Should also have a callback s it can be set dynamically (takes a string or a function like the current one)

EdwardCoyle:
This PR adds a new IdsDataGridColumn setting, `cellSelectedCssPart`, which optionally allows a column to provide a dynamic [CSS part]() name that can be used to provide a custom background color -- or other CSS -- to selected/activated grid cells.

Closes #989

- Pull/build/run
- Open http://localhost:4300/ids-data-grid/columns-custom-css.html
- Click a cell without selecting the row (using the checkbox).  When the row is activated, slightly darker custom colors should appear on the cell.
- Select some rows using the checkboxes.   When the rows are selected, the same darker background colors are applied.

### DIFF
--- a/src/components/ids-data-grid/ids-data-grid.ts
+++ b/src/components/ids-data-grid/ids-data-grid.ts
@@ -1498,7 +1498,10 @@ export default class IdsDataGrid extends Base {
 
     this.state.selectedRows.push(index);
     this.updateDataRow(Number(row?.getAttribute('data-index')), { rowSelected: true });
-    if (row) this.updateRowCells(index, row);
+
+    if (this.rowSelection === 'single' || this.rowSelection === 'multiple') {
+      if (row) this.updateRowCells(index, row);
+    }
 
     this.triggerEvent('rowselected', this, {
       detail: {
@@ -1655,7 +1658,10 @@ export default class IdsDataGrid extends Base {
         let cssPart = columnData.cssPart || 'cell';
 
         // Updates selected rows to display the correct CSS part (also activated rows in mixed-selection mode)
-        if (row.classList.contains('selected') || row.classList.contains('activated')) {
+        if (
+          (this.rowSelection === 'mixed' && row.classList.contains('activated'))
+          || ((this.rowSelection === 'single' || this.rowSelection === 'multiple') && row.classList.contains('selected'))
+        ) {
           if (columnData.cellSelectedCssPart) cssPart = columnData.cellSelectedCssPart;
           else cssPart = 'cell-selected';
         }
--- a/src/components/ids-data-grid/ids-data-grid.ts
+++ b/src/components/ids-data-grid/ids-data-grid.ts
@@ -493,8 +493,8 @@ export default class IdsDataGrid extends Base {
    */
   #cssPart(column: IdsDataGridColumn, rowIndex: number, cellIndex: number) {
     const cssPart = column.cssPart || 'cell';
-    if (typeof column.cssPart === 'function') {
-      return column.cssPart(rowIndex, cellIndex);
+    if (typeof cssPart === 'function') {
+      return cssPart(rowIndex, cellIndex);
     }
     return cssPart;
   }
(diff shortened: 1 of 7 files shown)


test/ids-data-grid/ids-data-grid-func-test.ts
```diff
@@ -2478,25 +2483,4 @@ describe('IdsDataGrid Component', () => {
       expect(dataGrid.getAttribute('id-column')).toBeFalsy();
     });
   });
-
-  describe('Events Tests', () => {
```
andriivarhanov, line 2482: what's the reason to remove the test?

EdwardCoyle: Good question... I think I accidentally got rid of these during my last rebase.  Good catch 👍

andriivarhanov (approved):

tmcconechy (changes requested):
@EdwardCoyle just wondering about a few things :

1. The price column when activated looks slate? I guess this is just the example but can we make it like the other colors? makes it look not selected
2. Something seems wrong with the way it works for mixed selection. I.E. when i click the checkbox it should only activate the checkbox, clicking the row activates (its now activating and marking the checkbox) see http://localhost:4300/ids-data-grid/mixed-selection.html
3. Open src/components/ids-data-grid/demos/columns-custom-css.ts and you get a type error (or run `npx tsc`). I think just need to add the new setting cellSelectedCssPart to IdsDataGridColumn

tmcconechy:
@EdwardCoyle 

1. I changed it - so it has the same color -> super minor but just for esthetics
2. Yes i see it is a little incorrect for mixed -> click just the checkbox and in this case it shouldn't activate the row like http://localhost:4300/ids-data-grid/mixed-selection.html. But that being said  i think it would be ok to use multiple select in the example, then its correct. 
3. I dont see this anymore, sorry must have gotten my branches mixed

EdwardCoyle:
@tmcconechy I think I fixed the row highlighting logic when in mixed selection mode now... I misunderstood what you meant by "activate" since it seemed everything was working (just that highlighting the row was occurring on "select" in mixed selection mode when it shouldn't).  Can you take another look to confirm?

andriivarhanov (approved):

src/components/ids-data-grid/ids-data-grid.ts
```diff
@@ -1634,6 +1642,38 @@ export default class IdsDataGrid extends Base {
...
+
+    const cells = row.querySelectorAll('.ids-data-grid-cell');
+    if (cells?.length) {
+      [...cells].forEach((cell: Element, columnIndex: number) => {
+        const columnData = this.columns[columnIndex];
+        let cssPart = columnData.cssPart || 'cell';
+
+        // Updates selected rows to display the correct CSS part (also activated rows in mixed-selection mode)
+        if (
+          (this.rowSelection === 'mixed' && row.classList.contains('activated'))
```
clayinfor, line 1662: These conditionals could be made more readable as follows:
```javascript
const isMixed = this.rowSelection === 'mixed';
const isNotMixed = !isMixed; // this.rowSelection === 'single' || this.rowSelection === 'multiple'; 
const isActivatedAndMixed = row.classList.contains('activated') && isMixed;
const isSelectedAndNotMixed = row.classList.contains('selected') && isNotMixed;

if (isMixedAndActivated || isSelectedAndNotMixed) {
  cssPart = columnData.cellSelectedCssPart || 'cell-selected';
}
```

clayinfor (commented):
👍

PR #1002, opened 2022-11-17
993 - IdsDataGrid row navigation setting

### LINKED ISSUE #993: IdsDataGrid: CellNavigation with keyboard and setting
**Enhancement**

- Add setting `cellNavigation` / true or false
- when off it behaves like this https://main-enterprise.demo.design.infor.com/components/datagrid/example-key-row-select.html
- default is on as it is now
- will need to style row selection and row focus
- keyboard up/down / enter to select as the example

jdilag90:
Added a row-navigation setting for data grid. Allows the user to focus the entire, not the just column, when navigating the data grid via keyboard/clicks. Changed naming to `rowNavigation` instead of `cellNavigation` , since cell navigation is the default behavior.

Closes #993 

2. Go to http://localhost:4300/ids-data-grid/row-navigation.html
3. Click on any row, ensure entire row has focused styles
4. Navigate the rows via keyboard arrows, ensure the navigated rows have focused styles
5. Hit spacebar on focused rows to select them

Mixed Selection Row Navigation
1. Go to http://localhost:4300/ids-data-grid/mixed-selection.html
2. Add `row-navigation` attribute to <ids-data-grid>,  or run in console `$('ids-data-grid').rowNavigation = true;`
3. Navigated rows should have focused and activated styles

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -12,6 +12,7 @@
 - `[DataGrid]` Added suppress row click for selection functionality. ([#737](https://github.com/infor-design/enterprise-wc/issues/737)
 - `[DataGrid]` Added support for context menu. ([#963](https://github.com/infor-design/enterprise-wc/issues/963))
 - `[DataGrid]` Added rowclick and rowdoubleclick events. ([#994](https://github.com/infor-design/enterprise-wc/issues/994))
+- `[DataGrid]` Added rowNavigation functionality. ([#993](https://github.com/infor-design/enterprise-wc/issues/993))
 - `[Icons]` All icons have padding on top and bottom effectively making them 4px smaller by design. This change may require some UI corrections to css. ([#6868](https://github.com/infor-design/enterprise/issues/6868))
 - `[Icons]` Over 60 new icons and 126 new industry focused icons. ([#6868](https://github.com/infor-design/enterprise/issues/6868))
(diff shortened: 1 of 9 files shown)


tmcconechy (changes requested):
Some minor things

1. The colors / theme(s)
2. Might be able to use `:focus-within` instead of adding the class.
3. The border looks thicker on the bottom -> maybe can use box shadows?
4. If possible the box shadow effect in addition to the borders would be nice to have (like the single cells focus state)

tmcconechy (commented):
Oh also can you add a brief note to https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-data-grid/README.md#converting-from-previous-versions-breaking-changes about the setting being renamed...

clayinfor (approved):
Looks good 👍 ... the only thing I noticed was that pressing the Spacebar on a highlighted row selects (or deselcts) that row, but pressing the Enter/Return key does not.

tmcconechy:
@clayinfor i think thats correct enter does other actions like click a button cell or submit. Space is the right key for selection see https://www.htmlelements.com/demos/grid/datagrid-bind-to-json/ as one example

tmcconechy (dismissed):

tmcconechy:
@jdilag minor lint error

deep7102 (approved):
Looks good 🍌

tmcconechy (approved):

Issue #971, opened 2022-10-25
Bug: Error invoking component from the web pack build


tmcconechy:
**Describe the bug**
At the moment we get an error "newing up" a component exported from the web pack build.

```
ids_enterprise_wc_components_ids_tag_ids_tag__WEBPACK_IMPORTED_MODULE_1___default(...) is not a constructor
```

**To Reproduce**

1. Go to angular-ids-wc/src/app/components/ids-tag/ids-tag.component.ts
2. Edit it as below
3. run `npm run start`
4. See error in the console on http://localhost:4200/ids-tag/example

```
import { Component, AfterViewInit} from '@angular/core';
import { routes } from './ids-tag-routing.module';
import IdsTag from 'ids-enterprise-wc/components/ids-tag/ids-tag';
@Component({
  selector: 'app-ids-tag',
  templateUrl: './ids-tag.component.html',
  styleUrls: ['./ids-tag.component.css']
})
export class IdsTagComponent implements AfterViewInit {
  public routes = routes.filter(r => r.path !== '');

  constructor() { }

  ngAfterViewInit(): void {
    console.log('Ids Tag init');
    const dynamicTag = new IdsTag()
    dynamicTag.textContent = 'Dynamic';
    document.body.append(dynamicTag);
  }
}
```

**Additional context**
Could be solved by es modules work #814 
See also https://teams.microsoft.com/l/message/19:b649a7c426e64e3dbedb0cc5202ce2d4@thread.skype/1667492602166?tenantId=457d5685-0467-4d05-b23b-8f817adda47c&groupId=4f50ef7d-e88d-4ccb-98ca-65f26e57fe35&parentMessageId=1667492602166&teamName=IDS%20Enterprise%20Development&channelName=Web%20Components&createdTime=1667492602166&allowXTenantAccess=false

LilBiggs:
Another scenario that needs to be thought about is extending an Ids Web Component. While trying to extend, I was receiving a very similar error as when you tried to use **new IdsTag()**

As a POC, I built enterprise-wc locally with a small tweak to the webpack.config to specify the library parameters. 
https://webpack.js.org/configuration/output/#outputlibrary
```
output: {
    ....
    library: {
        name: 'EnterpriseWc',
        type: 'umd'
    }
```

Code snippet from project

```
import IdsCalendarEvent from 'ids-enterprise-wc/components/ids-calendar/ids-calendar-event.js';
[...]

Issue #814, opened 2022-07-18
IdsBundleSize: Lazy Loading


tmcconechy:
**Description**
Take a look at light house results (but in general we knew) that the components are not sharing code too well on the beta. For example:

- ids-text is included 100% in many components (should be shared i.e. lazy loaded)
- ids-locale is included in ids-container making it very large (should be only the current locale)

For light house:
- go to https://main.wc.design.infor.com/ids-rating/example.html in chrome
- in chrome dev tools pick light house (do a scan in mobile and desktop)
- look at some of the red items like "Reduce unused JavaScript"
- have do a separate issue for things like "add meta description" and "roboots.txt" (additions to demo app for better scans) keep this to lazy loading for the most part

**Task**

Pattern may be something like:
```
loadChildren: () => await import('./items/items.module').then(m => m.ItemsModule)
```

**To Read**
https://webpack.js.org/guides/lazy-loading/
https://uploadcare.com/blog/lazy-loading-in-webpack/
https://webpack.js.org/guides/code-splitting/

**To Test**
- works in frameworks
- output is not repeating code
- Uses Es Modules?

PR #1023, opened 2022-11-29
990 - IdsIcon add custom icons

### LINKED ISSUE #990: IdsDatagrid: Ability to add custom icons
**Task**

The m3 team has some of their own custom icons. And they use them in the data-grid. The example https://main.wc.design.infor.com/ids-data-grid/columns-formatters.html shows a button with an icon in.

- [ ] Device a way to extend the ids-icon lib with custom icons. Its ok if we prescribe what they need to do too add it
- [ ] For testing can take an existing icon and make a small "custom modification" to it to use as any example. Perhaps this is an `addIcon` method that appends to the internal svg data array so it can be used. Make an example
- [ ] Make sure this works in the datagrid example (add a new column for "custom icon"

jdilag90:
Extended IdsIcon to support adding a custom icon. The custom icon can then be used in data grid similar to existing icons.
Icons are added by invoking a static `IdsIcon.addIcon(name, data)` method where an array of objects containing svg element attribute names/values are passed. The method then builds and caches SVG element strings similar to our existing icons.

Closes #990 

2. Go to http://localhost:4300/ids-icon/custom-icon.html
3. Check to see that the custom icons look ok
4. Go to http://localhost:4300/ids-data-grid/columns-formatters.html and scroll to last column
5. Check to see that the custom airplane icon looks/works as expected
6. Look through the `IdsIcon.addIcon()` implementation for any fixes/improvements that can be made


EdwardCoyle (approved):
@jdilag90 Making this work via static method seems like a great choice.  I think it's worth noting this might not work as intended for end users until we fix Webpack import issues described in #814 and #971.  It clearly works here, though.

My only comment about the functionality is that you might also want to provide a way to drop in SVG markup directly, as opposed to only accepting the object-based structure.  As a designer delivering icons to a dev, I'd most likely deliver a file using a tool like Figma/Illustrator/etc that generates SVG markup, so I could see this requiring an additional step for translation into an object.  However I think it's useful to have both, considering it may come from a backend this way.

I'd be good with approving as-is assuming this satisfies the M3 requirement.  Good stuff 👍

tmcconechy (commented):
Aside from Ed's comments - looks good to me. I would be ok with waiting to see if we have additional requirements in the future to have it in markup.

The only super minor thing(s)
- is can we make the column the second last one in the datagrid. The last one is a "spacer" column example which is usually the last column. Again - super minor
- also there is an airplane icon already in the "real icons" http://localhost:4300/ids-icon/example.html which confused me. Maybe we can cross check those custom ones and just show ones that are not already there. (I think this one was recently added to our set)

EdwardCoyle (approved):

tmcconechy (approved):

PR #1038, opened 2022-12-08
1006 - IdsDataGrid setFilterCondition filtered event bug

### LINKED ISSUE #1006: IdsDataGrid: Filter customization bugs
**Describe the bug**
When setting filter values using this.idsDataGrid.filters.setFilterConditions(filters), it triggers the filtered event which m3 things it shouldn't.
 
**To Reproduce**

1. Make an example in the filter examples with a button that calls setFilterConditons to set the filter conditions inline and an event handler that logs `filtered` event
2. Click it
4. See error -> should only log after applyFilter
5. also m3 would like a setting to hide the filter conditions button. we have a setting `filterTerms` for this.
6. Also i think `filterTerms` doesnt seem like the right name for setting the conditions. Maybe change to `filterConditions` instead before we go live

**Expected behavior**
- `this.idsDataGrid.filters.applyFilter` should trigger the filtered event not setting the values
- setting `filterTerms=[]` should hide the conditions button on the column or an explicit option `hideConditions=True`
- rename `filterTerms` to `filterConditions`

jdilag90:
Fixed bug where 'filtered' event fired when calling setFilterCondition. It should only fire when applyFilter is called.
Now setFilterCondition() doesn't automatically call applyFilter. Also, passing an empty Array for `filterCondition`, formerly `filterTerms`, will remove the filter conditions popup menu button from the grid column header.

Closed #1006 

2. Go to http://localhost:4300/ids-data-grid/filter-custom.html
3. At the input on top of demo page, type in a color value for data grid of products
4. Click set filter btn, check that `filtered` event isn't logged on console and your color value is reflected in the `no-operator` grid header.
5. Click apply filter btn, check that `filtered` event is logged on console
6. Check that the pre existing grid header filters/btns work normally and log `filtered` event
7. Try setting an empty array to `filterConditions` prop for `productName` column data inside grid/demos/filter-custom.ts. It should remove the filter menu button inside that column's grid header.


andriivarhanov (approved):
looks good

tmcconechy (commented):
Looks good - not even really a problem but maybe worth checking is..

- if i type in the input and press the buttons
- then go back and clear the input -> expecting to be able to reset the filter 
- as it does work in the inputs to clear the values and hit enter to filter

jdilag90:
@tmcconechy updated example input to listen to Enter and Backspace keys. It should update the filter conditions when you press on those keys

EdwardCoyle (approved):
Looks good @jdilag90 👍

janahintal:
QA Passed. Verified in https://main.wc.design.infor.com/ids-data-grid/filter-custom.html 
image

PR #1052, opened 2022-12-16
667/958/959 - Create/Integrate standalone popup component for IdsDatePicker

### LINKED ISSUE #667: IdsPopup/IdsDataGrid: Popup is cutoff by Data Grid containment

### LINKED ISSUE #958: IdsDatePicker/IdsTimePicker - Separate Popups into their own components
The current structure of the "Picker" components isn't flexible enough to be contained inside of other components.  

For example, containing both the IdsTriggerField and IdsPopup in the same IdsDatePicker shadow root causes the IdsPopup to be bound by the IdsDatePicker's CSS containment rules, like overflow: hidden or contain: content;  

In #667, the IdsDataGrid has these CSS rules applied in order to control the grid's virtual scrolling and number of displayed records.  Any Date Picker components displayed in the Data Grid are cut off at its borders because of the Picker menu's containment.

Other Popup types, such as IdsPopupMenu, are designed in a way that allows the menu to be connected to a triggering element without being nested inside the component.  We should replicate this model in all picker-style components, and share as much code as we can across all pickers to facilitate.

- Maybe need a generic Picker component/interface to describe

**Additional context**
Required for #667 to be complete

### LINKED ISSUE #959: IdsDatePicker/IdsMonthView - Remove interdependency
**Describe the bug**
Currently the IdsDatePicker and IdsMonthView components implicitly depend on each other:

- Month View import (https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-month-view/ids-month-view.ts#L31)
- Date Picker import (https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-date-picker/ids-date-picker.ts#L23)

The IdsDatePicker currently "reconfigures" itself depending on its use case (displayed as a trigger field vs displayed as a dropdown attached a button for use in a calendar toolbar).  In some situations, this design causes the Date Picker to need to load another copy of itself into its own shadow root in order to display pieces like the PickList menu.  This is a circular dependency problem we need to address.

**Expected behavior**
We should revisit the component design and see if there's a way to separate out the piece that would be the trigger field from the piece that would be attached to calendars.  Sharing these two views requires mixing of logic that might be more easily understood by a new developer separately.

As part of this, we should see if there is better code sharing to be done between the Month View/Week View/Calendar/Date Picker components that could be off-loaded to common libraries or utils

**Additional context**
Discovered while working on #667

EdwardCoyle:
This PR makes a number of changes to the IdsDatePicker component and its child components to resolve some known issues:
- A new, standalone "popup" component, `IdsDatePickerPopup` has been introduced, which contains only the IdsDatePicker's "popup" contents. (see [#958](#958))
- This component is built on top of a new base component `IdsPickerPopup`, which is intended to be used for other "picker"-style components in the future.
- `IdsDatePickerPopup` is now used in the IdsDataGrid with a new IdsTriggerField-based column filter, in place of what used to be the entire IdsDatePicker.  This fixes a problem with the popup being cut off by IdsDataGrid due to its encapsulation in the parent component (see [#667](#667))
- There is now a separate component, `IdsMonthYearPicklist` for the Month/Year/Week number picklist that was previously handled directly in the IdsDatePicker.
- The toolbar that used to exist inside the `IdsMonthView` and `IdsWeekView` for controlling those views has been moved out of these components and into `IdsCalendar`.  Additionally, `IdsDatePickerPopup` has a smaller, but similar toolbar for controlling its own MonthView.  These two changes fix a circular dependency problem, where date pickers used to depend on month views, and vice-versa (see [#959](#959)).
- IdsCalendar's "date picker dropdown" is now directly using `IdsDatePickerPopup`, instead of implementing a full IdsDatePicker.
- Some attribute setters/getters, and some utility functions, that were previously duplicated in a handful of places now exist in shared mixins `IdsDateAttributeMixin` and `IdsMonthViewAttributeMixin`.  These mixins are used in IdsDatePicker, IdsDatePickerPopup, IdsMonthView, IdsWeekView, etc to handle setters/getters, and their callbacks are implemented for behavior specific to each component.
- The original IdsDatePicker component still exists for general in-page use, and has also been updated to use the IdsDatePickerPopup internally.

*NOTE:* For full transparency, this PR doesn't cover changing IdsTimePicker.  We need to push changing that component in this manner to a separate issue since the scope of this one is huge.

Closes #667
Closes #958
Closes #959

Pull/Build/Run... then test these scenarios:

http://localhost:4300/ids-date-picker/picker-only.html
- This page demonstrates a single IdsDatePickerPopup instance, shared across 4 trigger fields, updating its values/settings whenever a new trigger field is accessed.
- Make sure picking dates from the picker works, and dates are set on the trigger field accordingly.
- Make sure opening the picker against a new trigger field updates the date values accordingly.
- Try using the month/year/week number picklist as well.

http://localhost:4300/ids-date-picker/example.html
http://localhost:4300/ids-date-picker/sandbox.html
http://localhost:4300/ids-date-picker/disabled-dates.html
- All previous examples should work as expected, utilizing the new picker popup

http://localhost:4300/ids-month-view/example.html
http://localhost:4300/ids-week-view/example.html
- These examples should still work, but they no longer display their previous toolbars for navigation (API calls to these components will still cause their displayed date ranges to change).

http://localhost:4300/ids-calendar/example.html
- This example contains the promoted IdsToolbar for navigating all views (month/week/day).
- Test that navigation works.
- Test that its still possible to add/update calendar events.
- Test that the IdsDatePickerPopup appears when clicking the Toolbar button with the enlarged date.  Choosing a date here should cause any active view to switch to that date.
- Test that navigating between months inside the IdsDatePickerPopup doesn't cause the event details pane to change contents (this should only happen after a date is chosen).

http://localhost:4300/ids-data-grid/filter-trigger-fields.html
- This example contains a filterable date column.
- Click the calendar icon in the date column header's trigger field.  It should activate an IdsDatePickerPopup.
- Pick a date such as "4/23/2021" to make sure filtering works as expected.
- Also note that the IdsDatePickerPopup is not cut off by the IdsDataGrid's boundaries once filtering occurs.


src/components/ids-picker-popup/README.md
```diff
@@ -0,0 +1,2 @@
+# Ids Picker Popup
```
tmcconechy, line 1: Empty readme? Maybe something basic or exclude it..

tmcconechy (changes requested):
Looks great just did a little testing and came up with the following suggestions.

1) http://localhost:4300/ids-date-picker/picker-only.html i dont 100% get what picker only means as compared to http://localhost:4300/ids-date-picker/example.html 
2) on http://localhost:4300/ids-date-picker/picker-only.html (but not example.html) i fi click the trigger while its opened it doesnt toggle closed
3) on http://localhost:4300/ids-calendar/example.html i noticed the keyboard does allow you to get to the datepicker. But if you do get to it by clicking when you open it the keyboard stops working. I.E. click it down, arrow, use arrow keys
4) could add a quick note to change log

jdilag90:
Nice work with the refactor! Just found a few things

Inside an [IdsCalendar in MonthView](https://main.wc.design.infor.com/ids-calendar/example.html) mode, days with more than 3 events have an overflow text/btn in the day container element (ex. 10/22/2019 has `+4 More`) . Clicking it should switch the Calendar to IdsWeekView's day view. This still works, but the `viewPicker` display value doesn't change to `Day`. Could be fixed by programmatically setting `value` of the viewPicker's IdsMenuButton where `overflow-click.calendar-container` is handled inside IdsCalendar.

When clicking on a day inside calendar, it sets the active date in the monthview and the date picker popup. Looks like the date picker popup's active date doesn't match the Calendar MonthView's active date. 

In the [picker only first example](http://localhost:4300/ids-date-picker/picker-only.html). Open date picker popup, click on a date. Open popup again and change solely the time using time picker. Looks like updated time values aren't reflected in the trigger field. It does update though if you set the time with time picker and then click on a date

deep7102:
Looks relay nice! saw few small things:

Read only
http://localhost:4300/ids-date-picker/example.html
- see `Date Field (readonly)`
- it should not let edit it's value
- fyi: see 4.x https://main-enterprise.demo.design.infor.com/components/datepicker/example-index.html

Date picker not switch modes in-range -vs- normal
http://localhost:4300/ids-data-grid/filter-trigger-fields.html
- in `Pub. Date` column
- click on filter conditions popup and choose `In Range`
- open date picker in filter field
- it should be set as `in-range` date picker
- if choose any other option (ie. Equals) should switch back to normal date picker type

EdwardCoyle:
Thanks everyone for the feedback.  I think this one is ready for a re-review.  

@tmcconechy @deep7102 I've addressed all the feedback in your comments and I think I've covered it all.

@jdilag90 I addressed all of these except for the last point --  the Time Picker doesn't affect each time portion in the trigger field on its own on https://main.wc.design.infor.com/ids-date-picker/sandbox.html -- however it is inconsistent with the 4.x components.  I'm not sure what the right behavior is so I kept as-is for now.

deep7102:
Date picker read-only, switch modes `in-range` -vs- `normal` looks good.

JS error on page load http://localhost:4300/ids-data-grid/filter-trigger-fields.html

Screen Shot 2022-12-22 at 12 15 59 PM

deep7102 (approved):
Looks good 🍌

tmcconechy:
All good from me as well...

tmcconechy (approved):

Issue #1012, opened 2022-11-23
IdsButton: In Angular the buttons may not be rendered as icon buttons


tmcconechy:
**Describe the bug**
In an angular project the buttons seem not rounded when they should be. Assume that https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-button/ids-button.ts#L91 is firing before the child contents are loaded so the class is not getting added.

**To Reproduce**

1. Go to https://stackblitz.com/edit/angular-ivy-o8lcre?file=src%2Fapp%2Faccess-controls%2Fcomponents%2Faccess-list%2Fheader%2Faccess-list-header.component.html,package.json
2. Notice the icon buttons on the card are not rounded

**Expected behavior**
Should be rounded.

**Suggested Actions**
Make `type="icon"` a new button type - or a new setting (need to consider all the different settings)

EdwardCoyle:
This may be related to #941

EdwardCoyle:
Noting that @tmcconechy's hunch was right and the component is checking for child contents too early.

I discovered as a workaround that using the IdsButton's `icon` attribute instead of composing the icon HTML works in Angular and gets the desired outcome:

before:
```html
<ids-menu-button id="menu-button" [attr.menu]="'menu'">
  <ids-icon icon="sort-down"></ids-icon>
  <span class="audible">Sort</span>
</ids-menu-button>
``` 

after:
```html
<ids-menu-button id="menu-button" [attr.menu]="'menu'" icon="sort-down">
  <span class="audible">Sort</span>
</ids-menu-button>
```

We can recommend the second way for now, but we should try and make both work

EdwardCoyle:
This [StackBlitz fork](https://stackblitz.com/edit/angular-ivy-hktteg?file=src/app/access-controls/components/access-detail/access-detail.component.html) demonstrates the original example using the workaround on all buttons.

danielortiz1982:
@EdwardCoyle after viewing the StackBlitz I noticed the dropdown is being covered. Not sure if this issue is related or something new. Just wanted to point that out. 👍 
<img width="538" alt="Screen Shot 2023-01-10 at 12 44 22 PM" src="https://user-images.githubusercontent.com/5215954/211625300-f5c3c284-34f8-41a6-9b92-2183d7308f4a.png">

EdwardCoyle:
[...]

PR #1090, opened 2023-01-05
1012 - Minor changes to fix Angular builds

EdwardCoyle:
While trying to test #1012 I ran into an issue with using `npm link` to create a local build of our latest WC `main` branch to the [Angular component examples](https://github.com/infor-design/enterprise-wc-examples/tree/main/angular-ids-wc) repo.  I discovered that there's a compile-time error due to some recent additions to the WC library that prevents the Angular component examples from compiling.

This was the error I saw:

Screen Shot 2023-01-05 at 10 26 19 AM

After some digging I discovered many of these properties originate inside mixins as get/set pairs, and are extended inside some component classes for various reasons,  `tsc` compiles these two things without any errors, so it appears the WC build works as expected.  However, we see the errors above once we try to import the library in our Angular examples.  

This is because the get/set pairs called out in the above errors are converted to regular properties in the mixin `d.ts` at compile time.  After this happens, the classes extensions with get/set pairs are mismatched, and any future builds depending on them cause errors.  See the below example of `rangeSettings` in IdsDateAttributeMixin on the left, compared to its `d.ts` output on the right:

| source | output |
| ---- | ---- |
| Screen Shot 2023-01-03 at 11 39 16 AM | Screen Shot 2023-01-03 at 11 44 42 AM |

Now compare this with the source of IdsDatePickerPopup on the left, compared to its `d.ts` output on the right:

| source | output |
| ---- | ---- |
| Screen Shot 2023-01-03 at 11 43 21 AM | Screen Shot 2023-01-03 at 11 42 39 AM |

This type of mismatch seems to cause the error.  In the interests of getting the NG examples building for dev/testing purposes, I created this PR to fix the five issues called out above by altering the way they're implemented.  In some cases I've put in helper methods for getting/setting these properties where nested.  In the case of the `scrollTop` property in IdsVirtualScroll, I used an enumerated prop since it's a similar case to [one that was needed in IdsButton](https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-button/ids-button.ts#L47-L52)

Related to #1012
Related to #958 

- Pull this branch
- Use `npm run publish:link` to push this branch to local NPM repo for linking
- Also pull the latest copy of `main` from the [WC Examples Repo](https://github.com/infor-design/enterprise-wc-examples)
- Switch to the `angular-ids-wc/` examples folder, and run `rm -rf node_modules && npm i && npm link ids-enterprise-wc`
- Use `npm start` to run.  The project should compile without errors.
- Smoke test http://localhost:4200/ids-date-picker
- Smoke test http://localhost:4200/ids-virtual-scroll


tmcconechy (approved):
These are all getters that arent settings right?

EdwardCoyle:
@tmcconechy only `activeDate` was a standalone getter.  `useRange`, `rangeSettings`, and `firstDayOfWeek` come from the date/month mixins and were originally get/set pairs.

andriivarhanov (approved):

deep7102 (approved):

PR #1103, opened 2023-01-19
1044 - IdsButton/IdsTabs Add hide focus mixin

### LINKED ISSUE #1044: IdsButton/IdsTabs: Focus State on Click
**Describe the bug**
Feedback on higher ups is they dont like the focus state activating when clicking. So we should go back to the "old" way.

**To Reproduce**

1. Go to https://main.wc.design.infor.com/ids-button/example.html and https://main.wc.design.infor.com/ids-tabs/example.html
2. Click on any button. it should focus but not show the focus state
3. Key to the buttons with tab. it should show the focus state
4. Same also for trigger fields (possibly other places)
5. Make this a setting hideFocus... true by default in dark and light mode, off in contrast (shows focus in contrast mode)

**Additional context**
Possibly use a hideFocus mixin - so it can be added to more components. (Can also look at EP code if needed)

Test accessibility further
- contrast
- screen reader 
- zoom (down up 35%)

andriivarhanov:
This PR adds `IdsHideFocusMixin` to toggle `hide-focus` CSS class for a component's container depends on how the component receives focus, click or a keyboard key.
The mixin was added to `ids-button` and `ids-tab`.
The mixin was tested on checkbox, radio, switch, input and related components.

Note: it doesn't work in Safari browser same as EP implementation and it looks like there are no workarounds since Safari doesn't focus buttons

Closes #1044

- pull and build
- go to http://localhost:4300/ids-button/example.html
- press Tab on your keyboard and see there is visible focus on every non disabled button
- click on a button and see there is no focus (for light and dark mode)
- switch to contrast mode and see the focus is visible on click
- go to http://localhost:4300/ids-button/focus-on-click.html where `hide-focus` is turned off
- see the focus is visible on click
- go to http://localhost:4300/ids-tabs/example.html
- see the focus is visible when tabbing to the active tab and right/left arrow key, the focus is not visible on click on a tab
- compare to EP example https://main-enterprise.demo.design.infor.com/components/button/example-index.html 

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -16,6 +16,7 @@
 - `[Breadcrumb]` Fixed popup menu being cutoff in truncated example. ([#906](https://github.com/infor-design/enterprise-wc/issues/906))
 - `[Button]` Removed named `text/icon` slots, re-worked the `iconAlign` setting to use only the default slot, and updated all examples/tests/docs to use only the default slot ([#839](https://github.com/infor-design/enterprise-wc/issues/839))
 - `[Button]` Updated all button style variants to reflect new IDS designs. ([#1046](https://github.com/infor-design/enterprise-wc/issues/1046))
+- `[Button/Tabs]` Added hide focus mixin. ([#1044](https://github.com/infor-design/enterprise-wc/issues/1044))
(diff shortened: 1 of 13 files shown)


tmcconechy (approved):
Well Done! Looks great 👍🏻

EdwardCoyle (approved):
Awesome!

Issue #1021, opened 2022-11-29
IdsDataGrid: Additional Formatters for Tag, Alerts, Image, Favorite, Card (templates), progress chart etc.


tmcconechy:
https://main.wc.design.infor.com/ids-progress-chart/example.html

PR #1120, opened 2023-02-06
1021 IdsDataGrid additional formatters

clayinfor:

Adds the following cell-formatters to IdsDataGrid:
- `IdsDataGridFormatters.alert()`
- `IdsDataGridFormatters.color()`
- `IdsDataGridFormatters.icon()`
- `IdsDataGridFormatters.favorite()`
- `IdsDataGridFormatters.tag()`
- `IdsDataGridFormatters.progress()`
- `IdsDataGridFormatters.rating()`
- `IdsDataGridFormatters.slider()`
- `IdsDataGridFormatters.stepChart()`
- `IdsDataGridFormatters.image()`
- `IdsDataGridFormatters.card()`

Closes https://github.com/infor-design/enterprise-wc/issues/1021

1. Run tests to ensure they all pass: `npx jest -- data-grid-func-formatters-test`
2. Run tests to ensure they all pass: `npx jest -- data-grid-func-test`
3. Check out branch, and do: `npm --components=data-grid run start`
4. @see http://localhost:4300/ids-data-grid/columns-formatters.html
6. @see http://localhost:4300/ids-data-grid/m3-features.html


doc/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
 ### 1.0.0-beta.3 Fixes
 
+- `[DataGrid]` Added more cell-formatters. ([#1021](https://github.com/infor-design/enterprise-wc/issues/1021))
```
tmcconechy, line 10: There is a beta 4 section now (actually if https://github.com/infor-design/enterprise-wc/pull/1121 gets merged its beta 6) depending which goes first

tmcconechy (changes requested):
Good Work Looks Nice but i think you did too much : ) plus some minor things we can fix.

1. We only really needed "Tag, Alerts, Image, Favorite, Card (templates), Targeted Achievement"
2. Im torn between leaving some of these and removing them so see below comments... for each
3. Can you add a row height picker on the toolbar about the http://localhost:4300/ids-data-grid/columns-formatters.html example so we can test that? Some other examples have it
4. Count: I think is un-needed because you can just use an integer formatter it doesn't look right and not sure about styling
5. For Progress bar this could be useful But lets remove the "your progress" text when in the grid cell and center the progress bar in the cell vertically and horizontally (maybe align: center may be used)
6. Rating Looks good -> can it be either readonly or clickable? and please cell vertically and horizontally
7. Slider -> is possibly useful but doesnt seem to render properly so can either remove, or just only show the slider part and center both vertically and horizontally (again align: center may be required)
8. Step Chart -> Looks good, lets again remove the step text when in the grid  and vertically and horizontally (again align: center may be required)
9. Favorite -> looks good, i think should be "gold" when "on" 
10. Badge -> is this centered? Lets make the cell wider and resizable too
11. Tag -> looks good, maybe show more types  and center both vertically and horizontally 
12. Alert -> looks good, and center both vertically and horizontally 
13. Color -> could be useful, will later maybe need an editor for it but can skip for now, for now lets center both vertically and horizontally 
14. Icon -> we did have a button where you can do this. I dont think you need an icon by itself but lets keep it since you did it. And just  center both vertically and horizontally 
15. Image -> looks good, can we make it reszable for testing?
16. Last Column (button with boat) -> seems un-centered, not sure if this was caused by the fix?

doc/CHANGELOG.md
clayinfor: Hey, thanks for the feedback, @tmcconechy!  I added changes to address each of the suggestions you made, except for the slight vertical-alignment on the `Tag` and `Last Column`.

tmcconechy:
These are smaller nit picks at this point

- in xtra small rowheight -> favorite, tag, alert and custom icon column (last one) could be more centered vertically
- in small rowheight -> alert and custom icon column (last one) could be more centered vertically
- in medium rowheight -> tag and custom icon column (last one) could be more centered vertically
- in large rowheight -> custom icon column (last one) could be more centered vertically
- in all row heights -> progress bar and custom icon should have the same padding on each side as the row height / text columns do
- thinking we can remove the slider actually -> i think it seems weird it would just be a number (sort of too complicated for a cell)
- i noticed on http://localhost:4300/ids-data-grid/m3-features.html the ones here didnt get the alignment class/fix

Good Job...

tmcconechy (approved):

deep7102 (approved):
Looks good 🍌

PR #1133, opened 2023-02-14
1065 - Separate IdsDropdown's Popup into its own component / integrate with IdsDataGridFilters

### LINKED ISSUE #1065: IdsDropdown: Dropdown is cutoff by Data Grid containment
This issue is related to #667, with a specific focus on IdsDropdown being cut off by the containment of the IdsDataGrid.  To test:

- Open https://main.wc.design.infor.com/ids-data-grid/filter-trigger-fields.html
- Open the "Location" dropdown
- Pick the menu item containing the "broken" image next to "Canada"
- After the grid filters, re-open the Location dropdown and see how it's cut off by the bottom of the grid:

Screen Shot 2022-12-23 at 11 48 03 AM

- [ ] Review if the IdsDropdown can have a "separate" list component outside its own containment.  If not, we need to implement this somehow in a way that's similar to how `IdsPopupMenu`, `IdsDatePickerPopup`, etc are able to connect to their triggering elements with an attribute (currently done with an ID reference)
- [ ] Re-implement the "dropdown" filter to account for the separate list component
- [ ] Fix examples/tests/docs to match

**Additional context**
- Discovered while working on #667

EdwardCoyle:
This PR makes some changes to IdsDropdown to allow it to break out of IdsDataGrid's overflow containment while being used as a filter:
- Separates the IdsDropdown's popup into its own new component, IdsDropdownList.  This new component is the new home to IdsListBox/IdsListBoxOptions
- Reconnected IdsDropdownList into IdsDropdown and restored its previous behavior
- Provided examples/docs of how to use IdsDropdownList separately
- Integrated IdsDropdownList's usage into IdsDataGridFilters wherever dropdown is used.

Closes #1065

_Test dropdown integration:_
- open http://localhost:4300/ids-data-grid/filter-trigger-fields.html
- Use the columns that filter with Dropdowns.  Test that filtering works as expected
- Use the "Price" columns and filter on `13.99` (one option present in the grid).  Try opening the dropdown columns -- the dropdowns should not be cut off by Datagrid containment.

_Test separate list example:_
- open http://localhost:4300/ids-dropdown/separate-list.html
- use the Dropdown/DropdownList and make sure they behave like the others

_Smoke test previous behavior:_ check all these and make sure they work as expected:
- http://localhost:4300/ids-dropdown/example.html
- http://localhost:4300/ids-dropdown/sizes.html
- http://localhost:4300/ids-dropdown/typeahead.html
- http://localhost:4300/ids-dropdown/sandbox.html
- http://localhost:4300/ids-multiselect/example.html


andriivarhanov (commented):
looks good, here is a few issues I've found:
- popup's position when it's at the top or when validation error is triggered 
image
- there is something with click outside to close the popup, on the page http://localhost:4300/ids-dropdown/example.html click on the `Dropdown with Allow Blank` example and click on the `Dropdown Using Ajax`, the previous one didn't close
image
- it looks like the popup should be 1px closer to the field
image
- http://localhost:4300/ids-dropdown/typeahead.html `Dropdown typeahead with blank option` doesn't have blank option
- when you open typeahead dropdown input's text should be selected
- typeahead's suggestions can't be selected
- http://localhost:4300/ids-multiselect/clearable.html 
  - not all items are loaded in the popup
  - can’t select the items
  - when it’s focused you can’t unfocus it by clicking on another element on the page 
- http://localhost:4300/ids-multiselect/groups.html group label being focused when selecting an item

sorry it's not much detailed feedback, I can clarify on some items, let me know

.stylelintrc.json
```diff
@@ -37,6 +37,7 @@
     "keyframes-name-pattern": null,
     "value-keyword-case": ["lower", {
       "ignoreKeywords": ["currentColor"]
-    }]
+    }],
+    "string-quotes": "single"
```
tmcconechy, line 41: Did this get re-added unintentionaly? This is the one i think got removed because its deprecated i spoke about.

tmcconechy (changes requested):
i see a couple of issues now

1. on http://localhost:4300/ids-dropdown/example.html - move the mouse slowly down the right side away from the component as you pass the tooltip one the tooltip will show and never hide
2. click on one near the bottom and it will open upward. And its not connecting with the field
image
3. more noticible on http://localhost:4300/ids-dropdown/sizes.html the popup is a bit lower than it used to be
image
vs
image

see: https://main.wc.design.infor.com/ids-dropdown/sizes.html

.stylelintrc.json
EdwardCoyle: I added this before you mentioned the change in the call -- without it the build had been failing on all SCSS files.  I'll rebase again and see if it helps without it

tmcconechy: Ok i thought it was one that caused the deprecation warnings when you run `npm run lint` so maybe just confirm that.... That was the only reason i removed it

tmcconechy (approved):

tmcconechy (dismissed):

andriivarhanov (commented):
still seeing some issues:
- there is something with click outside to close the popup, on the page http://localhost:4300/ids-dropdown/example.html click on the Dropdown with Allow Blank example and click on the Dropdown Using Ajax, the previous one didn't close
- when you open typeahead - dropdown input's text should be selected
- typeahead's/dropdown's list items can't be selected with keyboard
- http://localhost:4300/ids-multiselect/clearable.html `Clearable Dropdown (max choices 3)` example
  - not all items are loaded in the popup
  - can’t select an item except for the first one
  - when it’s focused you can’t unfocus it by clicking on another element on the page

andriivarhanov (approved):

tmcconechy (approved):

Issue #1131, opened 2023-02-14
IdsCalendar: Ability to get view picker change event control outside Calendar component


manoj-kothuri:
**Is your feature request related to a problem or use case? Please describe.**
The IdsCalendar component view picker change event listener control is not emitting to outside class where we can override the event pills Y_OFFSET position and set the MAX_EVENT_COUNT dynamically.

**Describe the solution you'd like**
We have our own class to handle the overriding of event pills position (Y_OFFSET) and handle the maximum number of event pills displaying in a day cell dynamically instead of fixed value of 3. As of now we implemented the logic in new class which is working fine for the initial load of Month view but when we change the views and comeback to Month view again the overriding logic is not working.

**Describe alternatives you've considered**
We tried to override the event pill Y_OFFSET position and event overflowing by adding the logic in CustomCalendarEvent class which is meant to be purely override the event pill template (multi line). As it's not ideal to use this class to handle the event pill positions and other logic.

**Additional context**
<img width="428" alt="image" src="https://user-images.githubusercontent.com/113168229/218730017-255f2392-c224-498b-a416-69cdd49f1b57.png">

And what application do you work on? Infor WFM

tmcconechy:
Not sure I understand this issue....

tmcconechy:
Maybe this needs an example setup to see it better?

manoj-kothuri:
> Maybe this needs an example setup to see it better?

Under ids-calendar/demos folder example-custom.html is having the example wherein we can replicate the issue.

tmcconechy:
Ok thanks @manoj-kothuri this is a little unclear to me. Do you think you can reframe the issue in terms of steps to reproduce/expected behavior ect?

manoj-kothuri:
> Ok thanks @manoj-kothuri this is a little unclear to me. Do you think you can reframe the issue in terms of steps to reproduce/expected behavior ect?

As suggested I used the steps recorder to showcase the issue, please find the below attached zip file for the same.
[IdsCalendar Steps Recording.zip](https://github.com/infor-design/enterprise-wc/files/10786074/IdsCalendar.Steps.Recording.zip)

So our main concern is we have implemented a logic to position the event pills accordingly based on the size (if fits in the day cell), it works on the initial load of the calendar. But, when we select the Week view or Day view and navigate back to Month View the event pills are getting overlapped as our logic is not getting called.

So could you suggest any approach to overcome this issue.

PR #1143, opened 2023-02-21
1131 - Adds events that render each time an event is added/removed

tmcconechy:

Adds a `beforeeventrendered` and `aftereventrendered` respectively to the weekview, monthview and calendarview component. In addition renamed the weekview events from render to rendered.

CC @LilBiggs -> this needs confirming it can solve the use cases. I tried setting the height on the event element and it seemed to work

Fixes https://github.com/infor-design/enterprise-wc/issues/1131

- go to http://localhost:4300/ids-calendar/example.html -> whatch the console fire for each event  (it can fire a lot)
- switch between day and month view -> whatch the console fire for each event (it can fire a lot)
- go to http://localhost:4300/ids-week-view/example.html -> whatch the console fire for each event  (it can fire a lot)
- switch between day and month view -> whatch the console fire for each event (it can fire a lot)
- for http://localhost:4300/ids-month-view/example.html the events are not rendered for this month but it is covered by tests, can be confirmed if desired by changing the events dates to match the presented date (year = 2018 in the DOM)


LilBiggs:
@manoj-kothuri - I think these changes will work for the positioning of our event pills. However we may still have an issue with the isOverflowing logic.

jdilag90:
@manoj-kothuri regarding the overflowing issue, is it that since the pill sizes can vary, the calculation for `isOverflowing` needs to be dependent on the sum of pill heights rather than # of event pills in a day cell?

clayinfor (approved):
👍

jdilag90 (approved):

manoj-kothuri:
> @manoj-kothuri - I think these changes will work for the positioning of our event pills. However we may still have an issue with the isOverflowing logic.

@LilBiggs - I have used these listeners and now we can position the custom event pills including isOverflowing logic as expected when switched between views.

manoj-kothuri:
> @manoj-kothuri regarding the overflowing issue, is it that since the pill sizes can vary, the calculation for `isOverflowing` needs to be dependent on the sum of pill heights rather than # of event pills in a day cell?

@jdilag90 - Yes we are "**isOverflowing**" calculation logic is based on the height of event pills and will be decided how many pills can be fitted in a day cell.

PR #1145, opened 2023-02-22
1102 - Datagrid no longer emits scroll event add scroll to bottom and append

### LINKED ISSUE #1074: IdsDataGrid: Scroll to bottom and add rows
**Describe the bug**
M3 (maybe) wants to see an example where when you scroll to the bottom a callback happens and you can append more rows. May only need it if virtual scrolling isnt really fast.

**Expected behavior**
Should work like https://main-enterprise.demo.design.infor.com/components/listview/example-footer-navigation.html

**Additional context**
M3 Call - But needs discussion
May also explore - load data from the backend during virtual scrolling... vs infinite scroll

### LINKED ISSUE #1102: IdsDatagrid: No longer emits 'scroll' event / add scroll to bottom and append
**Describe the bug**
On our app, it listens to the IdsDatagrid scroll event. It works on 1.0.0-beta.2 but it no longer works on 1.0.0-beta.3. 

We need the scroll event to check if the scroller is at the end, when it is the application request for the next set of records to the websever and adds it to the datagrid.

If a AtScrollEnd event is sent by the idsDatagrid, we can use it instead of scroll event.

**To Reproduce**

1. Open \enterprise-wc\src\components\ids-data-grid\demos\virtual-scroll.ts
2. Add code to listen to scroll event
```
dataGrid.addEventListener('scroll', (e: Event) => {
  console.info(`scroll`, (<CustomEvent>e).detail);
});
```
3. Run npm start
4. open virtual-scroll.html
5. scroll down
6. error, scroll event is not triggered

**Additionally**
7. Also test the scrollRowIntoView in junction with adding rows
8. Make an example that adds new rows when scrolling to bottom.
9. Add 33 to the bottom each number each time.
10. Try a tree with 33 rows...
11. add this new scroll to bottom to http://localhost:4300/ids-data-grid/m3-features.html

This works on 1.0.0-beta.2
**Expected behavior**

Scroll event should be triggered when scrolling on the idsDatagrid

**Version**
ids enterprise wc 1.0.0-beta.3
**Screenshots**
image

**Platform**
 - Infor Application/Team Name: M3 UX
 - OS Version: Windows 11
 - Browser Name: [chrome
 - Browser Version: 109.0.5414.75 (Official Build) (64-bit)

**Additional context**

clayinfor:

Add events for `scrollstart` and `scrollend`.

Closes #1074 and #1102 

1. Run tests to ensure they all pass: `npx jest -- data-grid-func-test`
2. Check out branch, and do: `npm --components=data-grid run start`
3. @see http://localhost:4300/ids-data-grid/infinite-scroll.html
4. @see http://localhost:4300/ids-data-grid/virtual-scroll-infinite-scroll.html


tmcconechy (changes requested):
Only a few things

1) does the new event virtualScrollTop work in both virtualscroll mode and or not? If not i think it should work on either setting and just act as a scroll event that tells you details on how far. So should be renamed scrollend or scrollstart or just scroll something without the name "virtual" as part of the setting. This is why i suggested on the issue to just fire the normal scroll event on the component so it looks standard then pass the ratio in the event details but willing to wiggle on that. Just think virtual isnt right.
2) For the example to match what m3 is doing can you make the number of records initially a lot lower (at least you so you dont need to scroll a long time down to test). They use 33 rows each time but thats based on what fits in a page. So might need to adjust this (maybe 100 or whatever works visually)
3) Would like another api name than `moreData` when setting but not sure what? Any other name ideas cc @EdwardCoyle ?
4) As i scroll down and append more data it doesn't seem to keep the scroll position and re-set it. the old example could do this on https://main-enterprise.demo.design.infor.com/components/listview/example-footer-navigation.html (middle example). Basic Idea was: save scrollTop then add rows then reset scrollTop 
 
I think this also solves https://github.com/infor-design/enterprise-wc/issues/1074

Lemme know if you need to discuss and im in any way unclear

tmcconechy (approved):
Very Nice. Thanks @clayinfor

deep7102 (dismissed):
Looks good 🍌 

on demo examples looks it loading same data again and again
```
return data.slice(0, 33);
return data.slice(0, 100);
```

tmcconechy:
Thats true @deep7102 might want to make it "count up" / loop...

clayinfor:
Good call @deep7102 . I've updated the demos @tmcconechy so that it doesn't look like it's repeating same items.

tmcconechy (dismissed):

deep7102 (approved):

src/components/ids-data-grid/ids-data-grid.ts
```diff
@@ -1038,6 +1038,15 @@ export default class IdsDataGrid extends Base {
 
   get columnGroups() { return this.state?.columnsGroups || null; }
 
+  moreData(value: Array<Record<string, any>>) {
```
EdwardCoyle, line 1041: Per @tmcconechy's comment, I'd say maybe this should be `addData` or `insertMoreData`, or just some indication of the action of "adding additional data" and not replacing it.  `moreData` doesn't bother me personally but could be confusing.

EdwardCoyle (commented):
@clayinfor This looks really great!  I'm just following up on this one item below, but otherwise ready to approve

tmcconechy (approved):

EdwardCoyle (approved):

PR #1155, opened 2023-02-27
1064 - Use IdsTimePickerPopup in IdsDataGridFilter, fix containment/cutoff issues

### LINKED ISSUE #1064: IdsTimePicker: Popup is cutoff by Data Grid containment
This issue is a followup to #667, with a specific focus on breaking the IdsTimePicker component's popup outside of the containment of the IdsDataGrid

- [x] Need to first finish ([#1063](#1063))
- [ ] Replace "time" IdsDataGridFilter's usage of `ids-time-picker` with an IdsTriggerField connected to an `IdsTimePickerPopup` component.  Follow the "date" filter's implementation of `IdsDatePickerPopup` for guidance.
- [ ] Fix examples/tests/docs to match

**Additional context**
- Follows up on #667
- Needs #1063 to be finished first

EdwardCoyle:
This PR follows up #667's work on separation of IdsPopup-based filter popups from their host components and into standalone components that can be moved around freely in the DOM.  This PR enables IdsTimePickerPopup to work in place of the "built-in" picker component, fixing the stacking context issue in IdsDataGrid.

Closes #1064
Related to #667

- Pull/build/run
- Open http://localhost:4300/ids-data-grid/filter-trigger-fields.html
- Use the "Pub. Time" column filter to slim down the results set (I used 2:25 PM)
- Re-open the timepicker popup.  When it opens, the entire popup should be accessible.

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -9,6 +9,7 @@
 - `[Build]` Updated `esbuild` to fix issues with import order. ([#1140](https://github.com/infor-design/enterprise-wc/issues/1140))
 - `[Calendar]` Added a `beforeeventrendered` and `aftereventrendered` event. ([#1131](https://github.com/infor-design/enterprise-wc/issues/1131))
 - `[DataGrid]` Fixed tooltip to show without text overflow. ([#1126](https://github.com/infor-design/enterprise-wc/issues/1126)
+- `[DataGrid/TimePicker]` IdsDataGrid now uses the IdsTimePickerPopup component inside Time-based filters. ([#1064](https://github.com/infor-design/enterprise-wc/issues/1064))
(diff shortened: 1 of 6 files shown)


andriivarhanov (commented):
looks good for the data grid. I noticed there is some popup position issue for `md` `lg` width size on the page http://localhost:4300/ids-time-picker/sizes.html , could you please check or I can create another issue
image

tmcconechy (approved):
For me aside from any other comments this is good.

EdwardCoyle:
@andriivarhanov after some investigation I think we should make a new issue around the position problem.  It's less to do with the popup and more to do with the box model of the trigger fields -- they all stretch to 100% of their containing grid columns and it sometimes affects the position.  I'd think these should be more like standard HTML Input fields and be `display: inline block;` by default, but I think changing the display type could have downstream effects, and I'd rather keep it separate.

andriivarhanov (approved):
@EdwardCoyle makes sense, works for me

clayinfor (approved):
looks good 👍

PR #1165, opened 2023-03-08
Chore: Add new deps

tmcconechy:
Updates a few deps. Mainly a test to confirm https://github.com/percy/cli/issues/1194#event-8694897940 by running a build.

NOTE: Percy Support might be looking at this. Hold until WIP is removed.

https://github.com/percy/cli/issues/1194

- checks pass
- `npm i` is clean of errors


package.json
```diff
@@ -53,42 +53,42 @@
     "@axe-core/puppeteer": "^4.6.0",
     "@babel/eslint-parser": "^7.19.1",
     "@babel/preset-env": "^7.20.2",
-    "@percy/cli": "1.18.0",
+    "@percy/cli": "1.20.3",
     "@percy/puppeteer": "2.0.2",
     "@types/jest": "29.4.0",
     "@types/jest-axe": "3.5.5",
-    "@types/node": "^18.14.4",
+    "@types/node": "^18.14.6",
```
EdwardCoyle, line 60: Wondering if this should match the recommended Node version in `.nvmrc`, which is lower: https://github.com/infor-design/enterprise-wc/blob/new-deps/.nvmrc#L1

EdwardCoyle (approved):
Everything seems fine, but a question below about one item:

package.json
tmcconechy: Not 100% sure but the latest is 18.15.0 there https://nodejs.org/en/download/ will just make them match for now. Having some wierd test issues. Might have to fix this later

tmcconechy: Yeah 18.14.6 doesnt exist so i guess they cant match (different packages at the end of the day)

PR #1172, opened 2023-03-10
153 - IdsDataGrid/IdsZip excel export feature

### LINKED ISSUE #153: General: Add xlsx and csv compatible export
Having a true xlsx export would solve a lot of issues users have with exported data. For example, in M3/GLT container number is a text field. It is not uncommon for that value to be a very long string of numbers. Currently a container number value of '3050000010100000' is exported correctly from the datagrid but when excel opens that file, be it CSV or XLS, it interprets it as a big integer and renders it as '3.05E+15'. When the the user subsequently saves that csv or xls, it saves the value as '3.05E+15'. This can also be the case for lot numbers in M3, as another example. I could give lots more. I'm sure this issue is prevalent in other Infor MRP products - long text values with only numeric characters. Xlsx would let you explicitly designate those long fields as text values and avoid Excel's 'wizardry'. We realize this is an Excel issue but that's not how customers see it. When they do an export and open it in Excel and the data is scrambled, it's not a good look for us. We also realize that there is import functionality in Excel that can circumvent this. In reality, users don't want to go through 10 steps (or whatever) every time they open an export. Our application is essentially a querying one so exports are commonly used to send data to FDA, etc.

Add 'xlsx' as a export option to the datagrid/excel.js component. 

1. As a workaround, you can prepend a tab or single quote character to each value for xls or csv exports. That forces Excel to interpret the value as text but in reality you are changing the data from what it truly is. Not good. It looks odd when displayed. If something needs to programmatically read the xls file, it will likely not like a leading tab or quote char. This is our current workaround solution but it is lacking.

2. Write/obtain our own xlsx code. Not great. It would be inconsistent with other applications that use IDS. Also, we would need to modify the standard grid behavior/code.

**Additional context**
1. I'm only mentioning csv here because that's what we've advised users to do. It gets rid of the 'The file format and extension of 'xxxxx.xls' don't match' warning. They would gladly go to xlsx if it was available. In our opinion, csv should be left alone as it's more likely to be read programmatically and should have raw data in it.
2. Ref this issue: https://github.com/infor-design/enterprise/issues/2995 
And what application do you work on? [e.g. {Infor XYZ}] 
GLT and by extension M3
4. Excel export works "properly" in DataTables https://datatables.net/extensions/buttons/examples/initialisation/export.html
and it does - although they export in xlsx format.
5. Check out https://github.com/exceljs/exceljs/blob/master/spec/browser/exceljs.spec.js for a possible start

**Additional context**
Nested datagrid data?

jdilag90:
Adds feature to export data to csv and xlsx excel files.
.xlsx are zip files, so an IdsZip util module was also added

Relevant doc links are provided in class/function comments, two important ones listed below
Office open xml docs - https://www.ecma-international.org/publications-and-standards/standards/ecma-376/ (Part 1 pdf)
Zip file format docs - https://docs.fileformat.com/compression/zip/

Closes #153 

2. Go to http://localhost:4300/ids-data-grid/export-excel.html
3. Try both `Export to .xlsx` and `Export to .csv` buttons
4. Check that excel files successfully download
5. Check that excel files open without an error message
6. Check that datagrid data matches up with excel data

**Test special characters/locales**
1. The last 3 rows of data grid have special characters/different languages
2. Check that those special characters are rendered correctly in excel
3. Go to http://localhost:4300/ids-data-grid/example.html
4. Change locale to something like Bulgarian to change date/time formats
5. Run `$('ids-data-grid').exportToExcel('xlsx', 'test');` and `$('ids-data-grid').exportToExcel('csv', 'test');` in console
6. Check that the excel file date/times keep the format from datagrid in selected locale

Note: Columns with number/decimal/integer formats are passed as raw number values.
So, formatting from data grid for numeric type values aren't reflected in excel sheet.
This allows excel to parse those values as numeric values, rather than formatted strings, which allows 
excel user to perform their own excel calculations/formatting. It also prevents excel `numberStoredAsText` 
warnings to convert potential numeric string cells into numeric value cells.
This can be changed if requested.

**Test .xlsx zip**
1. Try extracting the .xlsx zip file. (I used 7zip -> extract to)
2. It should successfully unzip without error and reveal internal directories/files of .xlsx file
****


tmcconechy (dismissed):
Beautiful Work! Looks great!

EdwardCoyle (dismissed):
I used `unzip -l [filename]` from the terminal on the XLSX file, for anyone else that doesn't have 7zip installed.

@jdilag90 This is really impressive, great work!

PR #1178, opened 2023-03-14
1102-idsdatagrid-infinite-scroll-padding-bottom-fix

clayinfor:

Fixes negative padding-bottom value for the ids body, and #appendMissingRows() method adds additional rows for virtual-scrolling.

Bugfix for #1102 

1. Run tests to ensure they all pass: `npx jest -- data-grid-func-test`
2. Check out branch, and do: `npm --components=data-grid run start`
3. @see http://localhost:4300/ids-data-grid/infinite-scroll.html
4. @see http://localhost:4300/ids-data-grid/virtual-scroll-infinite-scroll.html


clayinfor:
This PR fixes the issue where data is not loaded for smaller row-size due to miscalc on padding-bottom.  However, there is a problem where the scrollbar jumps when new data is loaded so the user loses their place in the datagrid.  Will have to revisit this later...

tmcconechy (approved):
LGTM 👍🏻 cc @GMInf

Issue #1179, opened 2023-03-14
IdsDatagrid: Opening Ids contextmenu does not close Soho Contextmenu and vice versa


GMInf:
**Describe the bug**
I am using Soho and Ids together in a project. When I right click on a soho control/component to open a contextmenu and right click on IdsDatagrid to open a contextmenu, the soho contextmenu does not close and vice versa.

**To Reproduce**

1. Open enterprise-ng-quickstart
2. Add a soho control with a context menu on the app.component.html
3. Add a ids component with a context menu on the app.component.html
4. Add ids js file dependencies ids-input and ids-popupmenu on angular.json

**Expected behavior**
The previously opened context menu (soho) should be closed when opening a new one (ids).

**Version**

- ids-enterprise: [e.g. v4.9.0 or v4.10.0]

**Screenshots**
Add Soho context menu on app.component.ts
![image](https://user-images.githubusercontent.com/53523076/225015094-b72d8c92-070b-4476-9cde-d423043c841d.png)
Add Ids context menu on app.component.ts
![image](https://user-images.githubusercontent.com/53523076/225015402-8270afc6-3437-4f96-b769-49af97c62209.png)
Add ids contextmenu to angular.json
![image](https://user-images.githubusercontent.com/53523076/225015790-d2bbcc81-b957-42ab-a60d-d983a1434b8c.png)

**Platform**
 - Infor Application/Team Name: M3 H5A
 - Device:
 - OS Version: Windows 11
 - Browser Name: chrome
 - Browser Version: Version 111.0.5563.64 (Official Build) (64-bit)
 
**Additional context**

tmcconechy:
So your using the Enterprise 4.x Context menu mixed with the Web component context menu? Maybe thats the issue as they don't talk to each other. Perhaps you can switch to exclusively use the new context menu then this should work?

GMInf:
Yes, we are using Enterprise 4/ EnterpriseNg 15 for our M3 H5 Angular. We need to replace the sohodatagrid with idsdatagrid for virtualization.
We can't replace the soho context menus because we have a lot of implementation of it and it will take a lot of work and may break a lot of stuff. It would be preferable for us to use sohocontextmenu inside the idsdatagrid but I do not think it is possible.
Our application could trigger an event to close the context menu on contextmenu beforeopen, would that be possible? If yes, what would be the event to close the sohocontextmenu and idscontextmenu?

tmcconechy:
One Idea (working idea)...

[...]

PR #1186, opened 2023-03-15
1179 - Add Close menu example

tmcconechy:

For discussion adds one simple solution for #1179. But it has issues.

a) if the menu is open it highjacks the right click event
b) requires jquery for the old component

So for now pushing for future ideas as an example. Suggestions?

- Part solution for 1179

- go to http://localhost:4300/ids-popup-menu/side-by-side.html
- right click the input then right click the text and see that each menu closes eachother going back and forth

### DIFF
--- a/src/components/ids-popup-menu/demos/side-by-side.ts
+++ b/src/components/ids-popup-menu/demos/side-by-side.ts
@@ -1,0 +1,15 @@
+// Add Event For Web Component Menu
+document.querySelector('ids-popup-menu#popupmenu')!.addEventListener('show', () => {
+  console.info('On Show', $('input#input-menu').data('popupmenu'));
+
+  const input = $('input#input-menu');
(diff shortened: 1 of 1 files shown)


deep7102:
Looks good!
guessing this fix only for this demo example, because on other type of WC components need different target element
like `IdsDataGrid` contextmenu etc.

```
document.querySelector('ids-popup-menu#popupmenu')?.setAttribute('hidden', '');
```

andriivarhanov (commented):
instead of setting `hidden` attribute we could use `hide()` method, works for me, don't see the right click event issue

`document.querySelector<IdsPopupMenu>('ids-popup-menu#popupmenu')?.hide();`

tmcconechy:
ok @deep7102 @andriivarhanov yeah everything looks good. this is just an example not a permanent fix for this issue yet

deep7102 (approved):

PR #1188, opened 2023-03-16
1144 - IdsDropdown default keyboard behavior

### LINKED ISSUE #1144: IdsDropdown: Default Keyboard Behavior Stopped Working
**Describe the bug**
Can no longer type on default dropdowns in a similar way to a standard select element.

**To Reproduce**
1. Go to https://main.wc.design.infor.com/ids-dropdown/example.html
2. tab to first dropdown
3. type h
4. see error -> nothing happens should select Hawaii

**Expected behavior**
The default behavior should be similar to https://main-enterprise.demo.design.infor.com/components/dropdown/example-no-search-lsf.html i.e.

- when closed typing a character or series of characters selects the next one found without opening
- when open typing a character will go to the item in the list but not select it

Also like https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select

- this method https://main.wc.design.infor.com/ids-dropdown/typeahead.html used on multiselect should still work / multiselect behavior should not change
- should not happen if disabled or readonly

**Additional context**
This worked when this component was introduced but stopped working at some point.

andriivarhanov:
This PR adds keydown event to the dropdown by default allowing to select an option and update the value by keyboard input

Close #1144

- pull and build
- go to http://localhost:4300/ids-dropdown/example.html
- focus `Normal Dropdown with Dirty Tracker` example either with Tab or clicking to the label
- type a text with the keyboard
- see the input value updates if an option exists by text entered
- open the popup
- type a letter with the keyboard
- see an option focuses by text entered
- go to http://localhost:4300/ids-multiselect/example.html 
- type a text with the keyboard 
- see the input multiple value updates if an option exists by text entered and doesn't exceed the max setting


tmcconechy (changes requested):
@andriivarhanov i do see a problem on http://localhost:4300/ids-multiselect/example.html

Issue 1:
- click to open
- select one
- click to open again and it doesn't open

Issue 2:
- type a when its focused and closes and it selects 
- try to type again a again i would think it would select
- when you open it and type a key it doesn't go down to it

Im not sure about the multi select doing this. Perhaps we turn this keyboard as a setting and make it only for dropdown?

tmcconechy (dismissed):

jdilag90 (dismissed):

PR #1196, opened 2023-03-20
1154 - Add support for double event click with data grid

### LINKED ISSUE #1154: IdsDatagrid: Request for event onHeaderColumnDoubleClick
M3/H5A executes a reset column size function when the header is double clicked. Requesting to implement the onHeaderColumnDoubleClick event.
Currently, there is no way to listen to the header double click event.

When the column header is double clicked, the event onColumnDoubleClick should be triggered similar to the 'rowdoubleclick' event.

An alternative would be specifying a more specific event.target on the 'dblclick' event. Currently the event.target is always the idsdatagrid.

**Additional context**
And what application do you work on? M3 H5A

deep7102:
Added support for double click event with data grid

Closes #1154

- Navigate to: http://localhost:4300/ids-data-grid/double-click.html
- Double click on different parts of grid body cells and header cells
- See console for "doubleclick" event details.

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - `[Button/Input/Dropdown]` Added posibility to show loading indicator. ([#1048](https://github.com/infor-design/enterprise-wc/issues/1048))
 - `[Datagrid]` Added feature to export data grid to excel xlsx or csv file. ([#153](https://github.com/infor-design/enterprise-wc/issues/153))
+- `[DataGrid]` Added support for double click event. ([#1154](https://github.com/infor-design/enterprise-wc/issues/1154))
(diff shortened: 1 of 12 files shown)


andriivarhanov (commented):
looks good and makes sense, I'm just not sure if it fits to the request from the issue, looking for @tmcconechy review

deep7102:
> looks good and makes sense, I'm just not sure if it fits to the request from the issue, looking for @tmcconechy review

@andriivarhanov 
yes, changed so it can use single event now `doubleclick` for whole grid.

tmcconechy (changes requested):
@deep7102 this looks good. But can we call the event dblclick instead of doubleclick.

https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event

I do like it better that there is one event vs row/headerdblclick. But think it would make it more natural if the name was the same as the native event.

andriivarhanov (approved):

tmcconechy (approved):

PR #1197, opened 2023-03-21
1114 - IdsDatePicker date localization

### LINKED ISSUE #1114: IdsDatePicker: Some languages are not parsing correctly
**Describe the bug**
After recent popup refactoring the ids-date-picker is no longer able to parse some locales

**To Reproduce**

1. Go to https://main.wc.design.infor.com/ids-date-picker/example.html 
1. switch to german
1. select a value in the first picker
1. clear value
1. try again

**Expected behavior**
Should work in all locales..

**Additional context**
Maybe need to format the value here against locale after it changes? https://github.com/infor-design/enterprise-wc/blob/main/src/components/ids-date-picker/ids-date-picker.ts#L582

jdilag90:
Fixes localization of dates in datepicker when selecting dates in different langauges

Closes #1114 

2. Go to http://localhost:4300/ids-date-picker/example.html
3. Select a different language
4. Select different dates from picker popup
5. Check that date value matches locale's date formatting

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -9,6 +9,7 @@
 - `[Datagrid/Dropdown]` Separated IdsDropdownList into its own component, re-integrated the new component into IdsDropdown, and fixed containment/cutoff issues in IdsDataGrid using the new list component. ([#1065](https://github.com/infor-design/enterprise-wc/issues/1065))
 - `[Data Source]` Changed the default sort function to behave more like Excel, sorting separately numbers, strings, and empty space. ([#1158](https://github.com/infor-design/enterprise-wc/issues/1158))
 - `[Datagrid]` Fixed broken validation styling in editable cells. ([#1171](https://github.com/infor-design/enterprise-wc/issues/1171))
+- `[DatePicker]` Fixed broken date parsing in different locales. ([#1114](https://github.com/infor-design/enterprise-wc/issues/1114))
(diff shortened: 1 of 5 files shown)


andriivarhanov (commented):
looks good, seeing three issues: 
- changing locale triggers invalid date validation
  - go to the http://localhost:4300/ids-date-picker/sandbox.html
  - change locale
  - select a date in `Valid Date Validation` example and see the issue
- ranges stopped working, check the page http://localhost:4300/ids-date-picker/range.html
- locale changing doesn't work when picker only, check the page http://localhost:4300/ids-date-picker/picker-only.html

tmcconechy:
Looks good! I like that it changes when you change locale now. Will approve when above is complete.

jdilag90:
@andriivarhanov @tmcconechy thanks for the feedback

* Fixed the range.html example. They should be working again.
* Fixed the invalid date validation triggered in the sandbox.html example.

The locale change doesn't affect the `picker-only.html` demo because a `format` is provided for each example.
You can see it in `picker-only.ts` inside `const fieldDefs`. Same behavior on [main](https://main.wc.design.infor.com/ids-date-picker/picker-only.html)

Btw, there's a maximum `callstack error` in [sandbox example](https://main.wc.design.infor.com/ids-date-picker/sandbox.html).
It's also on main. I think it's something to do with the theme mixin

tmcconechy (approved):
@jdilag i'll probably take care of https://main.wc.design.infor.com/ids-date-picker/sandbox.html on my theme work since that whole mixin is gone. Will make sure and test it after. The rest looks good!

andriivarhanov (approved):

PR #1204, opened 2023-03-23
1062 - IdsButton: change 'type' attribute to 'appearance'

### LINKED ISSUE #1062: IdsButton: change "type" setting to "appearance", support HTML "type" attribute
There's currently a naming convention issue on IdsButton, where its `type` attribute  is used to define the IDS button style (default/primary/secondary/destructive/etc).  

In addition, we don't currently have a way to set the `type` attribute on the inner HTMLButtonElement in shadow root.  For example, `type="submit"` for form buttons won't work properly.

- [x] Give a new name to our current `type` setting on IdsButton... We can't use "style" because it conflicts with yet another HTML attribute.  Maybe something like `appearance`, but open to suggestions.
- [x] Implement an attribute getter/setter for `type` that simply sets/gets the HTMLButtonElement's `type` attribute
- [x] Probably will need to update quite a few examples/tests to reflect this change

**Additional context**
Discovered while working on #839

EdwardCoyle:
This PR changes IdsButton's API in a couple ways:
- The previous `type` attribute has been renamed to `appearance`.
- A new `type` attribute has been added that simply passes the value through to the HTMLButtonElement used as the shadow root container (for example, `type="submit"` now works on buttons).
- All files in this project referencing the type attribute (components/examples/docs/tests) have been updated to reflect these changes

Closes #1062 

_Test appearance changes:_
Smoke test all button-related components for accuracy:
- http://localhost:4300/ids-button/matrix.html
- http://localhost:4300/ids-button
- http://localhost:4300/ids-toggle-button (should appear "default")
- http://localhost:4300/ids-trigger-field/example.html (trigger buttons should appear "default")
- http://localhost:4300/ids-message/example.html (two separately-styled buttons)
- http://localhost:4300/ids-action-sheet/example.html (see cancel button on mobile view)

_Test the new `type` setting:_
- Open http://localhost:4300/ids-form/example.html
- Use dev tools to inspect the Submit button.  The host element and the inner HTMLButtonElement should both have a `type` attribute of "submit":

Screen Shot 2023-03-23 at 12 57 32 PM

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -4,12 +4,13 @@
 
 ### 1.0.0-beta.9 Fixes
 
+- `[Button]` Renamed "type" attribute to "appearance", mapped a new "type" attribute that sets HTMLButtonElement's "type" attribute. ([#1062](https://github.com/infor-design/enterprise-wc/issues/1062))
+
 ## 1.0.0-beta.8
 
 ### 1.0.0-beta.8 Fixes
 
 - `[Button/Input/Dropdown]` Added ability to show loading indicator. ([#1048](https://github.com/infor-design/enterprise-wc/issues/1048))
-- `[Button]` Renamed "type" attribute to "appearance", mapped a new "type" attribute that sets HTMLButtonElement's "type" attribute. ([#1062](https://github.com/infor-design/enterprise-wc/issues/1062))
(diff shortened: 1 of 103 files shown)


andriivarhanov (commented):
looks good, the only thing I see there is `type` references left in `ids-button.scss` file

andriivarhanov:
@EdwardCoyle there was another place where `type` is used, pushed it to the PR if you don't mind

andriivarhanov (approved):

tmcconechy (approved):

Issue #1199, opened 2023-03-22
IdsDatagrid: VirtualScroll true - Instantly display data using an index similar to ScrollRowIntoView without scrolling


GMInf:
**Is your feature request related to a problem or use case? Please describe.**
Our application needs to display a lot of rows and in some cases we need to show the starting from index 150 for example. Using ScrollRowIntoView(), it is visible that the data is loaded starting from index 0 and then scrolled to index 150. We do not need to show index 0 and the scrolling rows. We just need to display the rows that includes index 150. 

**Describe the solution you'd like**
We would like to request a new function to display/load only rows that includes index X. Or a new parameter on ScrollRowIntoView that skips 0 upto the row close to X (example, ScrollRowIntoView(150, true /*disable scrolling*/),  it will render only rows  staring only in index 110 to 150 (depending on the viewable rows) 

**Describe alternatives you've considered**
A new function or a new parameter in ScrollRowIntoView will do

**Additional context**
And what application do you work on? M3 H5 Angular - Using IdsDatagrid Virtualization

clayinfor:
hey @GMInf , how do you feel about adding another IdsDataGrid attribute as such:
```html
<ids-data-grid start="150"></ids-data-grid>
```

And the datagrid's result-set would first render with record 150 at top... 

cc: @tmcconechy

tmcconechy:
I like that idea for the api @clayinfor 👍🏻

GMInf:
That would be great, thank you!

PR #1213, opened 2023-03-31
1198 - IdsDataGrid.scrollRowIntoView() does not go the the expected row

### LINKED ISSUE #1198: IdsDatagrid: ScrollRowIntoView does not go the the expected row
**Describe the bug**
When IdsDataGrid.ScrollRowIntoView() does not go the the expected row

**To Reproduce**

1. Open http://localhost:4300/ids-data-grid/virtual-scroll-infinite-scroll.html
2. Scroll down up to the last record (363) and go back to the top
3. On the console log, run document.querySelector("#data-grid-virtual-scroll").scrollRowIntoView(329)
4. The expected record show should be 330 but I am getting row 298

**Expected behavior**
The datagrid needs to show the row 330 

**Version**

- ids enterprise wc 1.0.0-beta.8

**Screenshots**

**Platform**
OS Version: Windows 11
Browser Name: [chrome
Browser Version: 109.0.5414.75 (Official Build) (64-bit)

**Additional context**

clayinfor:

After triggering an infinite-scroll, `IdsDataGrid.ScrollRowIntoView()` was not going the the expected row.

This PR fixes `IdsDataGrid.ScrollRowIntoView()` so that it finds correct row after infinite-scrolling.

Closes #1198

1. Open http://localhost:4300/ids-data-grid/virtual-scroll-infinite-scroll.html
2. Scroll down to the last record (300).
3. In the developer console, run:  `document.querySelector("ids-data-grid").scrollRowIntoView(0)`
4. The expected record shown should be 1.
4. In the developer console, run:  `document.querySelector("ids-data-grid").scrollRowIntoView(300)`
5. The expected record shown should be 300.


doc/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ### 1.0.0-beta.9 Fixes
 
 - `[Button]` Renamed "type" attribute to "appearance", mapped a new "type" attribute that sets HTMLButtonElement's "type" attribute. ([#1062](https://github.com/infor-design/enterprise-wc/issues/1062))
+- `[Datagrid]` Fix `IdsDataGrid.ScrollRowIntoView()` so that it finds correct row after infinite-scrolling. ([#1198](https://github.com/infor-design/enterprise-wc/issues/1198))
+=======
```
tmcconechy, line 9:
```suggestion
```

tmcconechy (changes requested):
@clayinfor i found a bug: for reference the page is 711x806 but i tried a few sizes and the issue seems to not depend on that.

- reload the page (fresh page)
- in the console type `document.querySelector("ids-data-grid").scrollRowIntoView(11)` -> in this case 11 is just out of sight
- then do `document.querySelector("ids-data-grid").scrollRowIntoView(300)` and in this case its only at 58

Whats the rule for where it should scroll to? Should it make it the first row visible? Or just someplace in view? (might need research on scrolIntoView html api)

clayinfor:
Thanks for your feedback @tmcconechy !  The rule is `scrollRowIntoView(x)` basically takes the number you pass in, and multiply it by the configured rowHeight to determine where in the resultset you should be taken to.  If it's not working, then the math was thrown off somewhere.

clayinfor:
hey @tmcconechy, upon further consideration, I think it best to label the latest issue you pointed out above a "won't fix" ... the reason being is because row 300, the one you're trying to scroll to, hasn't been loaded by infinite scroll yet, and so there's no way for our Web Components to know in-advance what the total result-set from the backend will be.

In the case of our `virtual-scroll-infinite-scroll.html` example, we know there are `> 300` rows before we scroll all the way down.  But in a real-world scenario, there's no telling what the last record would be.  And so this is something the end-user should take care of using a `for` or `while` loop etc to keep calling scrollRowIntoView(300) until the infinite-scroll reaches row 300+... but even this would be tricky unless the backend returns a `totalRows` value that the end-user could reference.

tmcconechy:
ok, i wonder if its related to https://github.com/infor-design/enterprise-wc/issues/1199 but we can leave it for now. What about `document.querySelector("ids-data-grid").scrollRowIntoView(11)` where it doesnt fully get it in view?

clayinfor:
Hey @tmcconechy, `scrollRowIntoView(11)` worked fine for me (see screen recording), so... unable to reproduce.

https://user-images.githubusercontent.com/93789065/229194413-0f732d76-487f-41df-81c6-8118eeb2f97f.mov

tmcconechy:
Here is what i see:  https://watch.screencastify.com/v/qvNK2NczY6Su1mgs29DL
So it might depend on screen size....

Maybe it doesn't even need to scroll in this case?

clayinfor:
Oh I see what's going on @tmcconechy ... `scrollRowIntoView(x)` is 0-index based, so when you ran `scrollRowIntoView(11)` in your screen recording, row # 12 is correctly displayed, because index-11 points to row 12... And `scrollRowIntoView(0)` would display row # 1, because row # 1 is `idsDataGrid.data[0]` under the hood.

tmcconechy:
Except in the first case it only went to 8 on the top in the video? So is that correct it just gets it "somewhere" in view? 
Lets see what @GMInfor thinks when testing

tmcconechy:
https://github.com/infor-design/enterprise-wc/issues/1198#issuecomment-1493794516

tmcconechy:
Probably closing this in favor of https://github.com/infor-design/enterprise-wc/pull/1239

PR #1252, opened 2023-04-25
1179 - Add fixes for closing menu and cleanup

### LINKED ISSUE #1179: IdsDatagrid: Opening Ids contextmenu does not close Soho Contextmenu and vice versa

tmcconechy:

Takes the code in side-by-side that was implemented to hide menus when a menu in the older framework is opened/closes and implements it in the component as a more permanent solution.

Uses mutation observer and setting classes. Also found a mutation observer that may not be cleaned up and fixed it and some JSDoc warnings.

Note: Includes https://github.com/infor-design/enterprise-wc/pull/1251 as it caused testing issues otherwise.

Fixes #1179 

- go to http://localhost:4300/ids-popup-menu/side-by-side.html
- right click between each component (old and new) and alternate seeing that one closes the other and vise versa


andriivarhanov (dismissed):
looks good

andriivarhanov:
@tmcconechy it looks like initial fix for the context menu doesn't work for some other menus (tests are failing for that reason), I've made another change adding hide only for context menu https://github.com/infor-design/enterprise-wc/pull/1251/files , tests should pass now

andriivarhanov (approved):

andriivarhanov (approved):

src/components/ids-popup-menu/ids-popup-menu.ts
```diff
@@ -155,12 +161,35 @@ export default class IdsPopupMenu extends Base {
...
 
+  /** Hide any older (enterprise) menus */
+  hideOtherMenus() {
+    const openMenu = (document.querySelector('.popupmenu.is-open') as HTMLElement);
+    openMenu?.classList.remove('is-open');
+    const openMenuElem = (document.querySelector('[data-popupmenu].is-open') as HTMLElement);
+    openMenuElem?.classList.remove('is-open');
+
+    // Use the is-open class to close it
+    this.#mo = new MutationObserver((mutations) => {
```
EdwardCoyle, line 179: We might not want to rebuild this on every 'show' event.  These fire often, so I'd say try and set this up once in `connectedCallback`.  The above code for closing the menus can run on each `show`

src/components/ids-popup-menu/ids-popup-menu.ts
```diff
@@ -155,12 +161,35 @@ export default class IdsPopupMenu extends Base {
...
     // Set up all the events specifically-related to the "trigger" type
     this.refreshTriggerEvents();
   }
 
+  /** Hide any older (enterprise) menus */
+  hideOtherMenus() {
+    const openMenu = (document.querySelector('.popupmenu.is-open') as HTMLElement);
+    openMenu?.classList.remove('is-open');
+    const openMenuElem = (document.querySelector('[data-popupmenu].is-open') as HTMLElement);
+    openMenuElem?.classList.remove('is-open');
```
EdwardCoyle, line 176: I think it's better to fire the [popupmenu's `close()` API](https://github.com/infor-design/enterprise/blob/main/src/components/popupmenu/popupmenu.js#L2384-L2390):
```suggestion
    openMenuElem?.data('popupmenu')?.close(true, true);
```

I think this is necessary because there are events bound during `open()/close()`, among other component interactions that wouldn't be accounted for simply by using the class to hide.

EdwardCoyle (changes requested):
If we add detection directly into the Popup Menu I think we might need a safer approach.  Please see below

src/components/ids-popup-menu/ids-popup-menu.ts
tmcconechy: @EdwardCoyle while this would be better it would require using the jQuery API and introduce a dependency. Since this is about cancelling it only I think it would be fine so i did it this way.

src/components/ids-popup-menu/ids-popup-menu.ts
tmcconechy: What if we do it only if elements are found in the page on the show? Think that would fly? I was thinking that maybe we dont add it unless we need it sort of thing.

EdwardCoyle (approved):

deep7102 (approved):
LGTM 🍌

Issue #1230, opened 2023-04-14
IdsScrollView: Add HideFocus Mixin


tmcconechy:
**Describe the bug**

We want the focus state hidden on https://main.wc.design.infor.com/ids-scroll-view/example.html so that clicking the circles does not show the focus state but tab does. I've tried adding it as per the mixin instructions but couldnt get it to work.
Maybe because its `a` tags. Could you take a look? (Not urgent)

**To Reproduce**

1. Go to https://main.wc.design.infor.com/ids-scroll-view/example.html
2. Click on the circles
3. Expect no focus state (only want it on tab)

**Additional context**
Noticed during #1118

PR #1264, opened 2023-05-01
904, 1230 - Fix to remove hash links for navigation with scroll view

### LINKED ISSUE #904: IdsScrollView: Remove hash links for navigation
**Describe the bug**
Find an alternate solution to using hash links in the ids-scroll-view component to avoid issues within SPA frameworks routing systems

**To Reproduce**

1. Pull angular-ids-wc example branch https://github.com/infor-design/enterprise-wc-examples/tree/316-angular-examples
2. `npm i && start`
3. go to http://localhost:4200/ids-scroll-view/example

**Version**

**Screenshots**
n/a

**Platform**
 - Infor Application/Team Name: [e.g. Infor XYZ, Infor ABC]
 - Device: (if applicable) [e.g. iPhone 6 or Samsung Galaxy S6]
 - OS Version: [e.g. Windows 10 or iOS 8]
 - Browser Name: [e.g. chrome, safari, stock browser]
 - Browser Version: [e.g. 22, 66.0.3359.181 (Official Build) (64-bit)]

**Additional context**
n/a

deep7102:
Fixed to remove hash links for navigation with scroll view
- Changed the logic to move the slides
- Add api methods (play, stop, toggle, first, last, previous, next, slideTo)
- Add events (some veto able)
- Add tests/covrage

Closes #904
Closes #1230
NG Examples: https://github.com/infor-design/enterprise-wc-examples/pull/32

- Navigate to:
- http://localhost:4300/ids-scroll-view/example.html
- http://localhost:4300/ids-scroll-view/sandbox.html
- See url when move slides, does not use hash links
- See the api calls should work
- Click on the circles, should not focus state (only should focus with keyboard)
- Should work fine with other platforms(angular etc..)


src/components/ids-scroll-view/ids-scroll-view.ts
```diff
@@ -64,105 +75,488 @@ export default class IdsScrollView extends Base {
...
+      const current = timestamp();
+      const diff = current - start;
+      if (diff >= this.delay) {
+        start = timestamp();
+        callback();
+        if (isOnce) cancelAnimationFrame(this.#slideRafId);
       }
-      this.#activateLink(event.target, true);
-      this.isClick = true;
+    };
```
jdilag90, line 285: Can we use a setTimeout to handle delay instead of a raf? If you insert a `console.log` inside `repeatOften()` you can see the function gets run ~1k times on a 60hz display.

jdilag90: Everything else looks good, one other thing. `currentSlideNumber()` sometimes returns a decimal since `scrollLeft` is used in the calculation

deep7102: Thanks! @jdilag90 
 - Updated now, `currentSlideNumber()` to be return whole number or undefined
 - About handle delay I had in mind RAF is more better in all cases epically in performance, I had read few articles in past like as: https://css-tricks.com/using-requestanimationframe/#aa-why-better or https://www.paulirish.com/2011/requestanimationframe-for-smart-animating. But after your comment I think about it more... and bit confuse now, because you have point too. It run with every frame also RAF come with some good stuff which `setTimeout()` or `setInterval()` not supported (like stops automatically when inactive, more battery-friendly etc.)
 - I can take some help here please guide, should we keep as RAF? or I can replace with `setInterval()`, because both ceases have some give or take.

tmcconechy: Can we avoid both a Raf and setTimeout? Thats ideal. So far we avoided setTimeout entirely and it would be the very first one.

deep7102: @tmcconechy @jdilag90 
Updated now, please review

tmcconechy (changes requested):
@deep7102 i think the `loop` setting was not fully removed in all places

Also in terms of functionality i see one thing
1. http://localhost:4300/ids-scroll-view/example.html
2. swipe left and it updates the pager
3. but keep swiping to the end one and the pager is not updated... ui swipe so last one is in view.

src/components/ids-scroll-view/README.md
```diff
@@ -14,6 +14,37 @@ The scroll view component allows swipe and scroll in between a number of slides/
 - **Circle Pager**: The older 4.x name for his component
 - **Carousel**: Another name for this sort of component with slides and buttons
 
+## Settings and Attributes
+
+- `loop` {boolean} sets the loop, true will loop back after next/previous reached to end
```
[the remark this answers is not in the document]
deep7102: - When swipe the pager is updating soon it stops swiping, otherwise it flickering because too many call to set selected state. It move depends on how much force was applied while swipe. 
- `loop` setting we need it use with next/previous so if it set to true it can keep going around

src/components/ids-scroll-view/demos/example.html
```diff
@@ -14,7 +14,7 @@
       <ids-layout-grid-cell col-span="1">
         <ids-scroll-view>
           <img class="camera-1" slot="scroll-view-item" alt="Slide 1, Sony Camera, Front"/>
-          <img class="camera-2" slot="scroll-view-item" alt="Slide 3, Sony Camera, Back Display"/>
```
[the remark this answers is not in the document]
deep7102: Nothing wrong, looks like was just typo in alt text it should be 2nd slide so it updated

src/components/ids-scroll-view/ids-scroll-view.ts
```diff
@@ -64,105 +70,315 @@ export default class IdsScrollView extends Base {
...
+   * @param {boolean} value The value
+   */
+  set loop(value: boolean) {
+    const val = stringToBool(value);
+    if (val === this.loop) return;
+    if (val) this.setAttribute(attributes.LOOP, '');
+    else this.removeAttribute(attributes.LOOP);
+  }
+
+  get loop(): boolean {
```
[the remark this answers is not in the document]
deep7102: No need to remove this setting

src/components/ids-scroll-view/README.md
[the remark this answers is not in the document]
jdilag90: I think what Tim means is that the pager loses sync with the active slide. If you swipe left to the last camera image, the last circle in `ids-scroll-view-controls` isn't shown as selected

tmcconechy (approved):
Not seeing it now tho. @jdilag90 did you see it. Looked ok on device simulator but not with magic mouse. But will approve since i cant see it anymore

jdilag90 (approved):

PR #1267, opened 2023-05-03
1199 - Added row-start attribute to datagrid to indicate first row to load

### LINKED ISSUE #1199: IdsDatagrid: VirtualScroll true - Instantly display data using an index similar to ScrollRowIntoView without scrolling

clayinfor:

This PR adds `IdsDataGrid.rowStart` prop to the datagrid which controls the `row-start` attribute. If the `row-start` attribute is set, then the datagrid's result set will begin at the row specified by the `row-start` value:

`<ids-data-grid row-start="777"></ids-data-grid>`

Fixes #1199 

1. Check out branch run: `npm run start`
2. Open http://localhost:4300/ids-data-grid/row-start.html
3. The first record shown on the datagrid should be row # 777.
4. Open http://localhost:4300/ids-data-grid/virtual-scroll-row-start.html
5. The first record shown on the datagrid should be row # 777.
6. Ensure tests pass by running: `npm run test:ci`

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### 1.0.0-beta.9 Fixes
 
+- `[DataGrid]` Added rowStart prop to data grid. ([#1199](https://github.com/infor-design/enterprise-wc/issues/1199))
 - `[Button]` Renamed "type" attribute to "appearance", mapped a new "type" attribute that sets HTMLButtonElement's "type" attribute. ([#1062](https://github.com/infor-design/enterprise-wc/issues/1062))
 - `[ColorPicker]` Fix change event firing multiple times. ([#1181](https://github.com/infor-design/enterprise-wc/issues/1181))
 - `[DataGrid]` Fixed text overflow for editable cells with data grid. ([#1175](https://github.com/infor-design/enterprise-wc/issues/1175))
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -4,13 +4,13 @@
 
 ### 1.0.0-beta.10 Fixes
 
+- `[DataGrid]` Added rowStart prop to data grid. ([#1199](https://github.com/infor-design/enterprise-wc/issues/1199))
 - `[DateUtils]` Fix `weekNumber()` returning 0 value for all weeks. ([#1243](https://github.com/infor-design/enterprise-wc/issues/1243))
 
 ## 1.0.0-beta.9
 
 ### 1.0.0-beta.9 Fixes
 
-- `[DataGrid]` Added rowStart prop to data grid. ([#1199](https://github.com/infor-design/enterprise-wc/issues/1199))
 - `[Button]` Renamed "type" attribute to "appearance", mapped a new "type" attribute that sets HTMLButtonElement's "type" attribute. ([#1062](https://github.com/infor-design/enterprise-wc/issues/1062))
 - `[ColorPicker]` Fix change event firing multiple times. ([#1181](https://github.com/infor-design/enterprise-wc/issues/1181))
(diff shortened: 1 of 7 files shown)


tmcconechy (commented):
Hi @clayinfor i tried to review this but wasn't sure how. Some initial impressions:

1) does it need an example page and docs? changelog too?
2) can you avoid the `setTimeout` we have as a rule tried to only use `raf` no `setTimeouts` as those in the past got out of control. One idea i had is maybe we can use mutation observer or some other eventing to know when at least the first row is loaded so it can be focused?
3) Can you in the future use shorter branch names `1199-idsdatagrid-virtualscroll-true-instantly-display-data-using-an-index-similar-to-scrollrowintoview-without-scrolling` -> maybe just `1119-add-rowstart` reason is we will be adding a process where you get a URL with a branch name for testing and this would go in the URL

clayinfor:
hey @tmcconechy , this is actually still a work in progress (I've added the WIP tag). I wanted to get something out the door that worked, so I pushed it up, but I'm not happy about the setTimeout either, and I've been trying get some other approaches to work.  I will add example page, docs and changelog note once it's actually ready.

clayinfor:
Also @tmcconechy , my bad about the long branch name.  I used GitHub's create branch feature, and that's what it spit out (based on the ticket title)

tmcconechy:
No worries @clayinfor .If we have no options for the `setTimeout` i did notice it slipped in other places. But i do think tracking somehow when the first row was added would work for the case.

I'll check this out more when not WIP ~ just maybe start with the steps then

clayinfor:
hey @tmcconechy , just and update that I've been looking into getting this working without the `setTimeout()`... so far I've got a potential solution using `scrollTop` (as opposed to `scrollIntoView()`), but it's not working consistently.

tmcconechy:
Ok, what about a mutation observer? I guess the setTimeout isnt "that" bad. In this case we just need to wait for the row to be there before we focus it. Unless there is other solutions? Did you consider moving it? I suppose the issue is other delays with loading data.

tmcconechy:
Here are some thoughts to help finish this off.

1. If setTimeout is the best way that works. Lets go with it. But would prefer something else if it helps 👍🏻 
2. Can you add example that shows it jumping to the row ~ so we can test it. And update the PR description.
3. Add a change log (beta 10 section)
4. Fix any tests broken / Sync with main

tmcconechy (commented):
The only issue i see is on http://localhost:4300/ids-data-grid/row-start.html when i load the page i see both scroll bars scroll down. but http://localhost:4300/ids-data-grid/virtual-scroll-row-start.html doesnt do that.

I think i would expect only the grid to scroll. Maybe need a stopPropgation or something?

https://watch.screencastify.com/v/PgzyM70LCsqYhWpgexcT

jdilag90:
For the first row-start example (http://localhost:4300/ids-data-grid/row-start.html). It gets scrolled to row id 779, can be seen in Tim's screencast video.
Minor thing, title in the row-start.html has `virtual-scroll` in the title

Looks good on the virtual-scroll-row-start example, I see it scroll to row 777

tmcconechy (approved):

PR #1269, opened 2023-05-05
1223 - Add support label wrap for input and align items with layout grid

### LINKED ISSUE #1223: IdsInput/IdsTriggerField: Add `label-wrap` setting
See https://main-enterprise.demo.design.infor.com/components/form/test-long-labels.html for a demo of how the 4.x components control form label layout.

We should (but don't currently) support this in the Web Components, but in a more fine-grain way so we can fully control label alignment and truncation.

- [ ] Add a `label-wrap` setting with 3 options:
   - ellipsis
   - wrap (confined to the current component size - maybe should be default)
   - stretch (stretch to the width of the grid column)

We also need to decide where this belongs:
-  component-level (use IdsLabelStateMixin), or
-  grid-level (use IdsLayoutGrid)

**Additional context**
Discussed here: https://github.com/infor-design/enterprise-wc/pull/1220#issuecomment-1500425505

deep7102:
Added support `label-wrap` for IdsInput and `align-items` with IdsLayoutGrid

Closes #1223

- Navigate to: http://localhost:4300/ids-input/label-wrap.html
- See different types of input labels
- See top inputs should align bottom to match input element (left/right columns)


tmcconechy (approved):
Looks good. Just wondering for anyone else that reviews if they agree? Should "LABEL - ELLIPSIS - NO STRETCH" be the default? This is subjective so this is probably fine.

@jmacaluso711 can you take a look for the layout grid changes?

EdwardCoyle (approved):
@deep7102 I think the feature here looks great!  One note about my testing -- I also tried the `label-state` attribute on the fields alongside label wrap, and it appears to work pretty well.  Two other items for discussion:

1. I think we should create a follow up issue to push this feature into the `label-state-mixin` so that we can make the wrapping feature work on other types of fields that contain a label, such as IdsEditor and IdsText.  IdsLabelStateMixin is already present in those components, so I think it would make sense to share from there.

2. Answering @tmcconechy's question: I really think the fully-visible label should be the default, and the "ellipsis" mode should be opt-in.  Under normal circumstances (read: not a forms-heavy app) I'd expect labels to be completely visible

Issue #1010, opened 2022-11-22
Component schemas for IDE intellisense


anhallbe:
**Is your feature request related to a problem or use case? Please describe.**
When writing HTML that uses the `ids-enterprise-wc` components, there is no help from the IDE (vscode). There is no way to know whether something is miss-spelled, or if using an attribute that doesn't exist. There is also no way to "discover" the API of a component other than reading the documentation.

**Describe the solution you'd like**
There is [a (in progress) standard/schema](https://github.com/webcomponents/custom-elements-manifest) for how a module should describe the custom elements that it provides. There is also a way to [configure VS Code](https://github.com/microsoft/vscode-html-languageservice/blob/main/docs/customData.md) to read a file containing the schema for a component library. The JSON Schema can be found [here](https://github.com/microsoft/vscode-html-languageservice/blob/main/docs/customData.schema.json).

I propose that such a file be generated automatically when building the `ids-enterprise-wc` library and included in the package. It could of course be written by hand, but I imagine that would require a lot of work and be error prone.

**Describe alternatives you've considered**
If there was a way to achieve this through TypeScript typings, I guess that would be preferable. That would allow type-checking and completion for things like `document.createElement("ids-<Ctrl+Space for completion>` and `someCustomIdsElement.addEventListener("customEve<Ctrl+Space>"`-

I tried the following locally in my project, just to see if it worked:

.vscode/settings.json:
```json
"html.customData": [
  "./ids-enterprise-wc.json",
]
```

and ./ids-enterprise-wc.json:
```json
{
  "$schema": "https://raw.githubusercontent.com/microsoft/vscode-html-languageservice/main/docs/customData.schema.json",
  "version": 1.1,
  "tags": [
    {
      "name": "ids-button",
      "description": {
        "kind": "plaintext",
        "value": "The IDS Button component is a simple wrapper around a standard HTMLButtonElement that is styled with Infor branding, and contains some additional API that makes it easy to set text, icons, and functionality."
      },
      "attributes": [
        {
          "name": "type",
          "values": [
            {
              "name": "primary"
            },
            {
              "name": "secondary"
            },
            {
              "name": "tertiary"
            }
          ]
        }
      ],
      "references": [
        {
          "name": "GitHub",
          "url": "https://github.com/infor-design/enterprise-wc/tree/main/src/components/ids-button"
        }
      ]
    }
  ]
}

```
Which makes the "ids-button" component slightly easier to use:

**Additional context**
[...]

PR #1277, opened 2023-05-11
1010: POC for generating custom element schema for vscode

anhallbe:

A WIP / POC for solving #1010 

https://github.com/infor-design/enterprise-wc/assets/4991954/ba4ad97b-5e73-4ae4-86d8-38f9718c3864

Adds a `generate-schema` npm script that generates a `vscode-custom-data.json` (NOTE: needs a better name/location) file based on all the `@customElement`s in the repository.

TODO
- [ ] Make sure `generate-schema` is part of the build/publishing pipeline (right now it has to be run manually)
- [ ] Make the schema a part of the published package, so that it can be referenced with `node_modules/ids-enterprise-wc/the-schema-file.json`
- [ ] Treat the generated schema as a build artifact and `.gitignore` it.
- [ ] Add a way to include `attributes`. This is currently not done because they are not strongly typed (possible future enhancement)

- Check out the branch
- Run `npm run generate-schema` (NOTE: the schema file is already checked in to git so this is not necessary unless there are changes to the source
- Open any HTML file and
  - Ctrl+Space for auto-completion of elements
  - Hover elements to see the documentation


tmcconechy (commented):
This is a great start. I'll look at it in more depth. I think in order to get the attributes shouldn't we generate this from the generated types? Or we can pull the attributes arrow out and parse it and find the settings?

Will have to look at the rest in a bit more depth.

We also probably want it to be part of the `npm run watch` command for locale developers if its fast enough. And part of `npm run release` of course. The static copies are just for deploying to the static site https://main.wc.design.infor.com/ids-data-grid/ so maybe arent the right spot 100%.

But we can build on this together.

Also did you try this? https://www.npmjs.com/package/typescript-json-schema

tmcconechy:
Going to close this but i will address on a new PR

PR #1306, opened 2023-05-30
1302 - Added fix to ensure inputs arent typing RTL

### LINKED ISSUE #1302: IdsDatagrid: Wrong cursor position and Column Filter Fields aligned 'right'
**Describe the bug**
When an ids column header is right aligned, the filter input field's cursor is on the left, it should be on the right. 

**To Reproduce**

1. clone enterprise-wc
2. run npm i
3. edit disable-client-filter.ts
4. update column color, set align: 'right'
```
  columns.push({
    id: 'color',
    name: 'Color',
    field: 'color',
    sortable: true,
    resizable: true,
    reorderable: true,
    formatter: dataGrid.formatters.text,
    filterType: dataGrid.filters.text,
    **align: 'right'**
  });
```
5. run npm start
6. disable-client-filter.html
7. type on the color header filter
8. error: cursor is on the left, it should be on the right
9. press backspace
10. error: backspace is behaving as it should but the cursor is on the wrong position (on the left, should be right) 

**Expected behavior**
When typing, cursor should be on the right of the text.

**Version**

- ids-enterprise-wc: 1.0.0-beta.11

**Screenshots**
image

**Platform**
OS Version: Windows 11
Browser Name: [chrome
Browser Version: 113.0.5672.127 (Official Build) (64-bit)

**Additional context**

tmcconechy:

When a column is aligned right the input would type as if it was RTL. This is fixed

Fixes #1302 

- go to http://localhost:4300/ids-data-grid/disable-client-filter.html
- type in the color field and the input should be aligned right but typing should be LTR again
- tab through the header, should be in the correct order
- check http://localhost:4300/ids-data-grid/filter.html and http://localhost:4300/ids-data-grid/filter-trigger-fields.html and things should still be in order


clayinfor (approved):
👍

jdilag90 (approved):

PR #1312, opened 2023-06-01
1244 - Add support for uppercase, disabled rows, collapse/expand all and redrawn event with data grid

### LINKED ISSUE #1244: IdsDatagrid: Request for AfterRedraw Event for Datagrid Customization

We need additional customizations for our implementation of the ids-datagrid for example:
1) Set the Column filter <input> to uppercase
2) Set the datagrid to disabled. Editable cells will not be editable/clickable and all rows will be grayed out to make it look disabled.
3) Add a new button to the column header for the purpose of expanding all rows for tree view

To do these customizations, I am thinking of updating the html of the datagrid. For example, adding a new part to the column header filter input to make it uppercase. Currently there is no way to add/update the part for the filter input unlike cells. 

Instead of requesting and implementing the individual requirements above, I think it  would work better if we can update the part(s) of any elements inside the data grid. To do that, we need to know when the idsDatagrid body has been drawn/redrawn so that we can update the elements that we need to.
For example, setting the Column header filter to uppercase. After the idsDatagrid is redrawn and afterRedraw is triggered, our listener
will query all the column filter inputs and add an uppercase part (or class) to change the <input> to uppercase.

**Additional context**
Additional button in column header
image

Disabled datagrid
image

And what application do you work on? M3 H5A

deep7102:
Added support with data grid as:
 - Text uppercase (cells, editors and filters)
 - Disabled rows
 - A icon`(+)` in column header to collapse/expand all rows for expandable or tree rows
 - Api methods `collapseAll()`, `expandAll()` for expandable or tree rows
 - Event that omits after redraw
 - Tests/coverage/docs and cleanup

Closes #1244

http://localhost:4300/ids-data-grid/uppercase.html
- Add some text on the first column header filter
- See columns `Product Name`, `Activity` `Action` text should be uppercase
- Use editor/filter inputs should also text be uppercase

http://localhost:4300/ids-data-grid/disabled-rows.html
- See rows (2,3,4,5) should be disabled
- Cells not be editable/clickable and all rows grayed out to look disabled

http://localhost:4300/ids-data-grid/tree-grid-expand-collapse.html
- See 2nd column header `Name`, should have `(+)`icon
- Click on this (+) icon to collapse/expand all rows
- Top button use api methods to collapse/expand all rows
- FYI: here in this example page the api method calls and `(+)` icon in header are not in sync, for example after page load/refresh very first thing if clicked top button `Collapse All` then click in header `(+)` icon. It will not expand all rows. Need to click two times in order to expand all rows

http://localhost:4300/ids-data-grid/after-redraw.html
- Open developer tools
- Page load/refresh
- See logs in console should fired event `redrawn`


src/components/ids-data-grid/README.md
```diff
@@ -369,6 +370,7 @@ When used as an attribute in the DOM the settings are kebab case, when used in J
 |`filterButtonTooltipCssPart` | {string} | Allows you to sets the filter button tooltip css part.
 |`cellSelectedCssPart` | {string} | Allows customization of a selected cell's background color.
 |`editor` | {object} | Adds an editor to the column if editable is set on the grid. See editing section for more details.
+|`uppercase` | {boolean} | Allows you to set the text as uppercase.
```
tmcconechy, line 373: I think this should go in column settings section not grid settings section

src/components/ids-data-grid/README.md
```diff
@@ -489,6 +491,7 @@ The formatter is then linked via the column on the formatter setting. When the g
 - `settingschanged` Fires after settings are changed in some way.
 - `scrollstart` Fires when data-grid reaches the topmost row.
 - `scrollend` Fires when data-grid reaches the bottommost row.
+- `redrawn` Fires after redraw the data grid.
```
tmcconechy, line 494: Can we call this `afterrendered` as i think that matches other components better

src/components/ids-data-grid/README.md
```diff
@@ -514,6 +517,9 @@ The formatter is then linked via the column on the formatter setting. When the g
 - `resetDirtyCells` Clears all dirty cell indicators.
 - `dirtyCells` Gives a list of all currently dirty cells.
 - `exoprtToExcel(format: 'csv' | 'xlsx', filename: string, keepGridFormatting: boolean)` Export datagrid datasource to an excel file. This keeps grid formatting by default.
```
tmcconechy, line 519:
```suggestion
- `exportToExcel(format: 'csv' | 'xlsx', filename: string, keepGridFormatting: boolean)` Export datagrid datasource to an excel file. This keeps grid formatting by default.
```

tmcconechy (changes requested):
Looks really good but a few comments:

- see inline comments regarding docs
- for after-redraw can we call it afterrendered (example and setting name)
- i think for http://localhost:4300/ids-data-grid/tree-grid-expand-collapse.html when you click it it should at least toggle (when you click it toggles from open to closed) it doesnt have to match every row for mixed case but when you click it i think just toggle the state

src/components/ids-data-grid/README.md
deep7102: Updated now, event name to `afterrendered`

src/components/ids-data-grid/README.md
deep7102: Updated now, to move to column section

deep7102:
@tmcconechy
Updated now, all the points you saw.

tmcconechy (approved):

clayinfor (approved):
Looks very good 👍

PR #1347, opened 2023-06-22
1332 - Rename tokens / add themes to side by side / new colors

### LINKED ISSUE #939: General: Add theme switch to side-by-side examples
We can't currently see the different IDS themes' implementation between the Web Components and the 4.x components in their side-by-side examples -- there is no way to switch the theme

- [ ] Implement theme switcher on all side-by-side example pages
- [ ] Log any discrepancies between the themes in both libraries in a new issue
- [ ] Specifically it was requested we also do this in the Angular example library
~- [ ] Also Personalization and Locale~ -> scope to another issue

**Additional context**
Suggested by @whernebrink

tmcconechy:

- Renames all tokens to new structure
- Adds a side by side example to each page that switches both ep and wc
- Adds new colors for slate

New Structure for names as follows:

MicrosoftTeams-image

Fixes #1332 
Fixes #939 
Also #7624 

- test any components you want (should be same as before)
- one quick way is http://localhost:4300/ids-theme-switcher/theme-builder.html or http://localhost:4300/ids-theme-switcher/theme-builder-minimal.html and scroll down
- test a few side by side examples like http://localhost:4300/ids-accordion/side-by-side.html or http://localhost:4300/ids-badges/side-by-side.html When changing themes the old and new components should both change. note that because of the CDN it works better if you preload the page for each theme to cache the css file. Might in the future make enterprise serve local but didnt want to introduce a dependency for it right now
- review https://github.com/infor-design/enterprise-wc/pull/1347/files#diff-a505f8d55dec8311337f6b1f2813baf760823b4613b93398d01f0597216a3b3fR1 or be aware of it for support
- look at the PR for the name changes, if anything you really disagree with we can (try to) push back
- check http://localhost:4300/ids-color/palette.html -> slate column is updated (compare to https://main.wc.design.infor.com/ids-color/palette.html)


jdilag90:
Changes look ok, I mainly tested components in the theme-builder examples.

Only weird rename I think is the switch from `--ids-font-size-md` to `--ids-font-size-40`.
Not a deal breaker, just think the naming `size-md` is more intuitive than `size-40` for the developers if they want font sizes to look 'normal' or 'default'

tmcconechy:
@jdilag90 i mentioned that as well. But it is intentional this way. Also there will be another layer that we would actually use so you would only see these in the lower level tokens. So I'm ok with it for now. It does take some thinking about when looking at though. So good flag i felt the same way at first

```
--ids-tag-font-size-default
--ids-label-font-size-default
  ```

ect..

jdilag90 (approved):

jdilag90 (approved):

Issue #1382, opened 2023-07-13
IdsDataGrid: Column title, dropdown filter etc. are hard to read when hovering in Dark Mode


glenlieorillo:
**Describe the bug**
Column title, dropdown filter etc. are hard to read when hovering in Dark Mode.

**To Reproduce**

1. Go to https://main.wc.design.infor.com/ids-data-grid/disable-client-filter.html
2. Change to Dark Mode
3. Hover column header

**Expected behavior**
Column header should be readable.

**Version**

- ids-enterprise: [e.g. v4.9.0 or v4.10.0]

**Screenshots**

https://github.com/infor-design/enterprise-wc/assets/102151352/bfb5ffde-8423-41c5-ad2a-4a2660ed1c62

**Platform**
 - All major browsers

tmcconechy:
Noticed this as well and is slated to be fixed on the attached PR

jbrcna:
this is now working as expected. moving ticket to Done. thanks
<img width="1587" alt="image" src="https://github.com/infor-design/enterprise-wc/assets/56722184/c3b91491-448f-4ad2-ae56-dbea2f1a9485">

Issue #1367, opened 2023-07-03
IdsDataGrid: Scrollrowintoview removed rows displayed on the list


jjorge1999:
**Describe the bug**
Scrollrowintoview removed rows displayed on the list.

**To Reproduce**

1. Go to 'http://localhost:4300/ids-data-grid/virtual-scroll.html'.
2. Open dev tools > Console tab.
3. Run scrollRowIntoView: $('ids-data-grid). scrollRowIntoView(400,false) 
4. Notice rows being removed.

**Expected behavior**
Scrollrowintoview without scrolling.

**Version**

- ids-enterprise-wc: 1.0.0-beta.12

**Screenshots**
Video: https://github.com/infor-design/enterprise-wc/assets/122866382/9a673c5e-c9c0-452e-8def-c088a82fd439

**Platform**
-Infor Application/Team Name: M3 UX
-OS Version: Windows 11
-Browser Name: chrome
-Browser Version: 113.0.5672.127 (Official Build) (64-bit)

**Additional context**

tmcconechy:
@jjorge1999 I think you want `$('ids-data-grid').scrollRowIntoView(400)` and never `$('ids-data-grid').scrollRowIntoView(400, false)`

But i think we should try to remove that second parameter as its for internal use cases. But this should solve your issue.

jbrcna:
QA Passed.

PR #1384, opened 2023-07-13
1367 - Change scrollRowIntoView api

tmcconechy:

Removes the "internal" second parameter of the `scrollRowIntoView` function. Also spotted an issue in dark mode.

Fixes https://github.com/infor-design/enterprise-wc/issues/1367
Fixes https://github.com/infor-design/enterprise-wc/issues/1382

- review inline code
- notice that `scrollRowIntoView` no longer has in internal second parameter
- test a few examples regarding row start, virtual scroll, infiinite scroll and see it is unaffected
- also go to http://localhost:4300/ids-data-grid/editable.html and change to dark mode, hover the header and it shouldnt be all black as per https://github.com/infor-design/enterprise-wc/issues/1382

### DIFF
--- a/src/components/ids-data-grid/demos/virtual-scroll-infinite-scroll.html
+++ b/src/components/ids-data-grid/demos/virtual-scroll-infinite-scroll.html
@@ -14,7 +14,6 @@
         <ids-text font-size="12" type="h1">Data Grid (Virtual Scrolling)</ids-text>
       </ids-layout-flex-item>
       <ids-layout-flex-item grow="1" overflow="hidden">
-        <ids-text>Test Text</ids-text>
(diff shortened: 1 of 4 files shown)


src/components/ids-data-grid/demos/virtual-scroll-infinite-scroll.html
```diff
@@ -14,6 +14,7 @@
         <ids-text font-size="12" type="h1">Data Grid (Virtual Scrolling)</ids-text>
       </ids-layout-flex-item>
       <ids-layout-flex-item grow="1" overflow="hidden">
+        <ids-text>Test Text</ids-text>
```
jdilag90, line 17: can remove this test line now

PR #1388, opened 2023-07-17
1308 - IdsLayoutGrid content max width

### LINKED ISSUE #1308: IdsLayoutGrid: Content max width restrictions
So this is actually a feature that we already have using the enterprise-ng library. In enterprise-ng we would use row and column classes to create the page layouts. As per default the row class will restrict the layout width to make sure content does not get stretched indefinitely. 

Here is an example of how that works in enterprise-ng:

image

Currently the enterprise-wc library does not provide any limitation to this so content does stretch all available space which might not be a preferred behavior. I could not find any built in way to deal with it currently however I would expect it be fairly straight forward for users to deal with it manually. Even so perhaps some build in handling could be nice. 

I would like this to be an opt in behavior. So perhaps an attribute for `<ids-layout-grid>` could be a way to do it. Here you already have nice definitions of the [breakpoints](https://github.com/infor-design/enterprise-wc/tree/main/src/components/ids-layout-grid#responsive-breakpoints) so perhaps one could just specify which breakpoint or a custom value to set a "MaxWidth" attribute. 

I was thinking if the row and column and similar layout classes from enterprise could be useful to port into the wc library. But at the same time I think that might just be messy. Right now the  `<ids-layout-grid>` is the way to do page layouts. And a flexible one at that with that wide range of [attributes](https://github.com/infor-design/enterprise-wc/tree/main/src/components/ids-layout-grid#attributes). So instead of adding other tools I'm thinking improving the ones available is likely the better approach.

**Additional context**
Infor M3 Portals

jmacaluso711:
This PR adds the ability to set a maximum width on the `ids-layout-grid`.  The available sizes for `maxWidth` are based on the current predefined breakpoint sizes. However, the `maxWidth` attribute also allows users to specify a custom width if needed, giving them flexibility in defining the maximum width according to their specific requirements. While custom widths can be used, it is generally recommended to utilize the predefined breakpoint sizes whenever possible for consistency and responsiveness across different devices and viewports.

closes #1308 

- pull and run branch
- go to http://localhost:4300/ids-layout-grid/example.html
- Scroll to bottom and see contained grid example

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -57,6 +57,7 @@
 - `[Draggable]` Removed the `ids-` from the event names. ([#1328](https://github.com/infor-design/enterprise-wc/issues/1328))
 - `[Icons]` Added new icons. ([#7510](https://github.com/infor-design/enterprise/issues/7510)
 - `[Locale]` Added new translations. ([#7512](https://github.com/infor-design/enterprise/issues/7512)
+- `[LayoutGrid]` Added support for max-width and centering. ([1308](https://github.com/infor-design/enterprise-wc/issues/1308))
 - `[Themes]` Added theme support and css variables for all components. These can be used for customizing components and creating themes. See ([CUSTOMIZING.md](../doc/CUSTOMIZING.md)) for details. Note that with the `<ids-theme-switcher>` its now better to use the full theme name `<ids-theme-switcher theme="default-light">`. ([#1118](https://github.com/infor-design/enterprise-wc/issues/1118))
 - `[Splitter]` Added new designs and minor fixes. ([#1328](https://github.com/infor-design/enterprise-wc/issues/1328))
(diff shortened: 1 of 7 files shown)


jmacaluso711:
> Looks good but wondering two things
> 
> 1. could call it in the example `Contained Grid (margin auto / maxWidth lg)` (minor)
> 2. when i resize to mobile sizes it doesnt "break down to mobile. Not sure if thats a bug or something needed in the setting?

@tmcconechy I address 1., Regarding 2. yes, to achieve responsive behavior, the `ids-layout-grid-cell` component requires additional attributes to be set per breakpoint, like this `<ids-layout-grid-cell col-span="12" col-span-sm="6" col-span-lg="4">` There's an example here: http://localhost:4300/ids-layout-grid/responsive.html

tmcconechy (approved):

tmcconechy:
hey @jmacaluso711 can you merge main on here? and then put the change log in the beta.14 section as we have since released beta 13 already (minor thing)

EdwardCoyle (approved):

PR #1419, opened 2023-08-03
1313 - IdsSwaplist data refactor

### LINKED ISSUE #1313: IdsSwapList: Working with swaplist and underlaying dataset
**Describe the bug**
The swaplist seems to not be fully linked with the underlaying dataset. From what we've seen this is making it difficult to use and manage the data. 

There are some separate parts to this:
1. How to input a custom name on the different posts? 
List just use the out of the box name "List #X" 

image

2. How to input data into different posts? Currently the [documentation](https://design.infor.com/code/wc-beta/latest/ids-swaplist) only shows the "count" setting to add number of columns. However any data added gets added in all three. I noticed some ways of dealing with these things in the [non wc library](https://design.infor.com/code/ids-enterprise/latest/swaplist) however that seems to not be compatible with the wc component and would also not be good practice using jquery to target DOM elements in angular.

3. how to retrieve the data from specific post in the swaplist?
It seems the underlaying dataset is not updated when changes are made. So even if we see items switching posts visually the dataset remains unchanged as far as I could tell. 

image

**To Reproduce**
I've been looking at the[ swaplist demo component in enterprise-wc](https://github.com/infor-design/enterprise-wc/tree/main/src/components/ids-swaplist). Specifically the angular example found [here in the wc-examples](https://github.com/infor-design/enterprise-wc-examples/tree/main/angular-ids-wc/src/app/components/ids-swaplist/demos/example)

1. Go to the swaplist angular example
2. notice the titles
5. swap around some items
7. check the swaplist data in the click events using the dev console
8. notice no attributes keeping the dataset per post or the list titles.

**Expected behavior**
1. Ideally we can send the data with a title attribute that will then be the visible title.
2.  Ideally we can input the dataset sorted by which data should end up under which title from issue 1. (Like available, selected, deselected ecample below from enterprise-ng)
image
3. Data to be tracked under each title/post so that when clicking the swap or when saving the document we can easily get data from all lists or one specific list depending on needs.

**Version**
- ids-enterprise-wc beta-11

**Platform**
enterprise-wc with angular

Additional context
Infor M3 Portals

jdilag90:
Refactored swaplist to connect UI interactions to datasource. 
Setting new data should now also re-render the lists.
Added `updated` event for users to listen for datasource changes after UI interactions.
Added types for list data structure to allow more customization of lists' names/items.

Fixes #1313 

2. Go to http://localhost:4300/ids-swaplist/example.html
3. Interact with swaplist via list arrows and drag and drop
4. Check console for logged `updated` events and that data matches UI
5. Run code below in console to set new swaplist data
6. Swaplist should be re-rendered with 2 lists now with new titles and 1 item each
7. Try steps 3 & 4 again to make sure data syncing is still intact

```
$('ids-swaplist').data = [
  {
    "id": 1,
    "name": "Testing first list",
    "items": [
      {
        "id": 1,
        "text": "Option AAA",
        "value": "option_a"
      }
    ]
  },
  {
    "id": 2,
    "name": "Testint second list",
    "items": [
      {
        "id": 3,
        "text": "Option BBB",
        "value": "option_b"
      }
    ]
  }
]
```


tmcconechy (approved):
Looks Really Good 👍🏻

clayinfor (dismissed):
Looks really good @jdilag90 👍 . I do see one potential problem where I'm unable to drag multiple items from the center-column to the side-columns.  Not sure if this by design... see video below.  The empty drag-areas probably just need a min-height set.

https://github.com/infor-design/enterprise-wc/assets/93789065/cbd46e82-7014-47b0-8d04-bb24fc78bc27

tmcconechy (approved):
I'm ok if you want to punt on the dragging issue. Good from my side.

jdilag90:
@tmcconechy @clayinfor Addressed the drag/drop issue by adding min height to drop zone. Thanks for the tip!

clayinfor (approved):
🚀

PR #1461, opened 2023-09-07
1454 - IdsDataGrid virtual/infinite scrollend fix

### LINKED ISSUE #1454: IdsDataGrid: Virtual Scroll not triggered after infinite scroll reaches MAX_ROWS_IN_DOM
**Describe the bug**
Virtual scroll isn't triggered when infinite scroll loads enough data to reach MAX_ROWS_IN_DOM.
Problem was hard to detect because our infinite scroll examples weren't fully simulating async data fetches when `scrollend` was triggered.

1) Make these changes inside virtual-scroll-infinite-scroll.ts 
image

2) Scroll down until there are 165 rows in DOM. 

3) Scroll down again and see that rows aren't added/recycled into the DOM at the next `scrollend` event, even though there are 198 rows loaded into the grid's datasource

**Expected behavior**
Virtual scroll should work as expected after Step 3 mentioned above

**Version**
- ids-enterprise-wc: v1.0.0-beta.15

**Screenshots**
See below that datagrid stuck at row 165 even though there are 198 rows inside dataGrid.data
image

**Additional context**
Many of our scroll examples aren't fully simulating async infinite scroll fetches. It's worth updating them to more accurately represent the async fetches made by client applications.

jdilag90:
Fixes a bug where the data grid isn't updated/recycled with newly appended rows to datasource after MAX_ROWS_IN_DOM (150) is reached. The fix was to manually trigger row recycling in these scenarios.
Also updated infinite scroll examples to make async fetches for each `scrollend` event and not just the first fetch. That way it's similar to how it behaves in production. 

For example, when
`data.length` = 165
`MAX_ROWS_IN_DOM` = 150
`rows.length` = 150

The variable `rowsNedded` below would equal `0`, which prevented any new rows from being rendered/recycled
https://github.com/infor-design/enterprise-wc/blob/f58a515a078b3ae883cb41ef680abd9568101291/src/components/ids-data-grid/ids-data-grid.ts#L1208

Closes #1454 

1. Checkout branch
2. Go to examples related to infinite scrolling
3. http://localhost:4300/ids-data-grid/virtual-scroll-infinite-scroll.html
4. http://localhost:4300/ids-data-grid/virtual-scroll-infinite-scroll-m3.html
5. http://localhost:4300/ids-data-grid/infinite-scroll.html
6. Check that you can still infinite scroll to the end of the dataset


clayinfor (approved):

tmcconechy (dismissed):
Everything looks good but src/components/ids-data-grid/demos/virtual-scroll-infinite-scroll-m3.html could use a label and a theme-switcher like the other examples if that doesnt effect it?

tmcconechy (approved):

tmcconechy:
Its minor but that custom css we could use a variable of some sorts so themes works or add variables for each theme to the custom css but doesn't matter at all for now. i think it just copied from another example.

And the better way now might be to change the data-grid-selection-background-color. We can deal with that another time

EdwardCoyle (approved):

PR #1495, opened 2023-09-21
1394 - Add Personalization API

### LINKED ISSUE #1394: Personalization: Add as a first class feature
**Describe the bug**
The parts for personalization are there with tokens but need to add a few things

**Needed**

- [x] An api to IdsGlobal 
- [x]  Also add version to IdsGlobal
- [x]  Api should take a color and call color progression generation (via ids-color-primary tokens)
- [x] incorrect color progress on `#2AC371`
- [x]  A way to show it (possibly the ... button) on some pages / or reuse the theme pages
- [x] Check/Fix https://main.wc.design.infor.com/ids-header/color.html (using tokens)
- [x] Show it on header
- [x] Show it on buttons
- [x] Show it on tabs
- [x] Show it on module tabs
- [x] See what else can be personalized and add to the page
- [x] Test dark and contrast mode (button, header (color), personalize)
- [x] Document complete list and API
- [x] Fix up https://main.wc.design.infor.com/ids-theme-switcher/theme-builder-minimal.html as a demo page
- [x]  The search field is wierd

tmcconechy:

This PR adds a personalization API but also in doing so.

- fixed incorrect tabs colors
- fixed checkbox issues
- refactored header to use tokens instead of color-variants so it can be personalized

Still could use some more fixes for lower end colors and more flexibility but good enough for now, until we get more use cases. Will add tests later

Fixes #1394 

- PRIMARY: Try http://localhost:4300/ids-personalize/example.html and pick a color
- try http://localhost:4300/ids-checkbox/example.html in both themes
- try http://localhost:4300/ids-tabs/example.html
- try http://localhost:4300/ids-header/example.htm 
- try http://localhost:4300/ids-header/colors.html and try to change the color property inline.
- Check out the API and docs


jdilag90:
looks good, couple small things

Dirty tracking checkbox has the check cut off (http://localhost:4300/ids-checkbox/example.html)
image

Getting this typescript error for ids-personalize.ts.
get color/set color have to share similar types.
image

tmcconechy:
@jdilag90 those two items should be fixed

tmcconechy:
@EdwardCoyle thanks, should be fixed. (although that color is dark so it can only go as dark as #000 for the last few colors)

jdilag90 (approved):

PR #1524, opened 2023-10-05
1498 - Added support for base href

### LINKED ISSUE #1498: Themes: Loading of ids-theme-default-light.css is failed because of lack of base href in the url
**Describe the bug**
Loading of the theme failed because of missing base href in the request URL:

image

**To Reproduce**
1. Add a base href to your Angular project https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base 
2.  After adding it the theme file will be failed with 404 status.

**Expected behavior**
When calling the theme endpoint, the URL should include the baseHref if it's presented in the application:
`${baseHref}/themes/ids-theme-default-light.css`

Example:

http://localhost:4200/process-intelligence/themes/ids-theme-default-light.css

image

**Version**

- ids-enterprise-wc: [1.0.0-beta.14]

**Additional context**
The problem occurs in the Angular 15 version.

I think the same issue occurs in the React, Vue.js frameworks. It would be nice if you could also check if we can fix it in all frameworks. Thanks!

tmcconechy:

Changed the theme fetching logic to use any `<base>` tags if they exist in the page.

Fixes #1498

- go to page http://localhost:4300/ids-theme-switcher/example.html
- check the network tab in dev tools
- switch themes to dark and contrast
- in all cases theme should work and the path should be like `http://localhost:4300/themes/...`
- http://localhost:4300/ids-theme-switcher/base-href.html
- switch themes to dark and contrast
- in all cases theme should work and the path should be like `http://localhost:4300/myappui/themes/...`

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 1.0.0-beta.16 Features
 
+- `[Themes]` For theme switching added logic to check for a `<base href="">` tag. ([#1498](https://github.com/infor-design/enterprise-wc/issues/1498))
+
 ### 1.0.0-beta.16 Fixes
 
 - `[DataGrid]` Fixed a bug on the size of the `xxs` filter row inputs. ([#1456](https://github.com/infor-design/enterprise-wc/issues/1456))
--- a/src/components/ids-theme-switcher/demos/base-href.html
+++ b/src/components/ids-theme-switcher/demos/base-href.html
@@ -7,7 +7,7 @@
 </head>
 <body>
   <ids-container role="main" padding="8" hidden>
-    <ids-theme-switcher mode="light" self-managed></ids-theme-switcher>
+    <ids-theme-switcher mode="light"></ids-theme-switcher>
     <ids-layout-grid auto-fit="true" padding="md">
       <ids-text font-size="12" type="h1">Themes</ids-text>
     </ids-layout-grid>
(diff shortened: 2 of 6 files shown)


EdwardCoyle:
@tmcconechy I noticed on the example with the different base path that when the page initially loads, there is no style.  However you can use the theme switcher to select a theme, and it does load successfully:

Screenshot 2023-10-10 at 8 07 13 AM

tmcconechy:
@EdwardCoyle - what page? or is it NG? Do i need something to try?

EdwardCoyle:
@tmcconechy sorry it's http://localhost:4300/ids-theme-switcher/base-href.html, I tried after a fresh build/cleared cache

jdilag90:
I'm experiencing the same lack of styles on page load that Ed mentioned. I think it's the `self-managed` attr on theme switcher.

`<ids-theme-switcher mode="light" self-managed></ids-theme-switcher>`

I think we also add a line of code to resolve the `onThemeLoaded` promise we have on IdsGlobal, when theme-switcher has `self-managed` enabled. This is handled inside of IdsElement's `loadTheme()` fn
image

tmcconechy:
@EdwardCoyle @jdilag90 -> fixed that. I dont need self-managed on this example it was copied from another example

jdilag90 (approved):

clayinfor (approved):

PR #1117, opened 2023-02-02
941 - Fix angular property binding

deep7102:
**Explain the details for making this change. What existing problem does the pull request solve?**
Fixed to angular where attributes property binding was not work properly for:
- ids-toggle-button
- ids-app-menu
- ids-menu-button
- ids-pager
- ids-text
- ids-toolbar
- Added new `ids-text` example in angular examples repo

**Related github/jira issue (required)**:
Closes #941
Related to #931

**Steps necessary to review your pull request (required)**:
- Pull WC branch ([941-angular-property-binding](https://github.com/infor-design/enterprise-wc/tree/941-angular-property-binding)) and build/run
- Use `npm run publish:link` to push WC branch to local NPM repo for linking
- Pull the latest copy of main from the WC Examples repo
- Pull WC Examples branch ([941-angular-property-binding](https://github.com/infor-design/enterprise-wc-examples/tree/941-angular-property-binding)), or use `main` WC Examples branch after PR ([NG#28](https://github.com/infor-design/enterprise-wc-examples/pull/28)) merged
- Switch to `angular-ids-wc` examples directory, and run `npm i && npm link ids-enterprise-wc`
- Use `npm run start` to run the angular examples demo app
- Navigate to below examples and see all attributes should working properly
- http://localhost:4200/ids-toggle-button/example
- http://localhost:4200/ids-toggle-button/icon-align-right
- http://localhost:4200/ids-app-menu/example
- http://localhost:4200/ids-app-menu/example-footer
- http://localhost:4200/ids-menu-button/example
- http://localhost:4200/ids-menu-button/disabled
- http://localhost:4200/ids-menu-button/display-selected
- http://localhost:4200/ids-pager/example
- http://localhost:4200/ids-pager/list
- http://localhost:4200/ids-pager/sandbox
- http://localhost:4200/ids-text/example

Also smoke test WC demo app
- http://localhost:4300/ids-toggle-button
- http://localhost:4300/ids-app-menu
- http://localhost:4300/ids-menu-button
- http://localhost:4300/ids-pager
- http://localhost:4300/ids-data-grid/pagination-client-side.html
- http://localhost:4300/ids-data-grid/pagination-server-side.html
- http://localhost:4300/ids-data-grid/pagination-standalone.html
- http://localhost:4300/ids-text/typography.html

**Included in this Pull Request**:
- [x] An e2e or functional test for the bug or feature.
- [x] A note to the change log.

PR #1584, opened 2023-10-30
1570 - Fix input readonly and disabled

### LINKED ISSUE #1570: IdsInput: Impossible to remove readOnly state
When set a field initially as readonly, it is impossible to remove readonly from it. This was not happening on previous versions.

```html
<ids-input id="ids-field" label="IDS Text Field" readonly></ids-input>
```

Now remove the readonly property:
```javascript
document.getElementById('ids-field').removeAttribute('readonly');
```

Btw the doc (https://github.com/infor-design/enterprise-wc/tree/main/src/components/ids-input) says "readonly {boolean} sets the input's readonly state." (for disabled and readonly accepts boolean value), but when set `document.getElementById('ids-field').readonly = false`, still doesn't work.

Visually looks like is not readonly anymore, but it is impossible to type in the field.

**Version**
1.0.0-beta.16

**Screenshots**
MicrosoftTeams-image

**Platform**
 - OS Version: Mac OS
 - Browser Name: Chrome
 - Browser Version: 118.0.5993.88 (Official Build) (x86_64)

tmcconechy:

Some fields (readonly and disabled but feel free to suggest others). Are made as toggles meaning if the attribute is there it should be removed to turn it off. Since adding #1117 we found some inputs that removing readonly stopped working

cc @clayinfor for testing too

Fixes #1570 

- go to http://localhost:4300/ids-input/example.html
- try the following in the console: `$('#readonly').removeAttribute('readonly')`
- make sure you can now type in the (formerly readonly) input
- also try: `$('[disabled]').removeAttribute('disabled')`
- make sure you can now type in the (formerly disabled) input


clayinfor (approved):

clayinfor (approved):
👍

PR #1631, opened 2023-11-13
1478 - add ids-lookup field to data-grid editor

### LINKED ISSUE #1478: IdsDataGrid: Add IdsLookup Input for IdsDataGridEditor
To add IdsLookup type on `IdsDatagridEditor.input` so that it's possible to add lookup on editable IdsDatagrids

It would be good to have the same functionalities that the `Soho.Editors.Lookup` offers. This should probably also work like a modal where we can provide a callback function to call whenever we resolve a value inside a lookup modal/form

N/A

**Additional context**
For additional context:
The `Soho.Editors.Lookup` that's used by the `SohoDataGridColumn.editors` API:
image

image

clayinfor:

Added `lookup` type to `IdsDataGridEditor` so that it's possible to add an ids-lookup field on editable IdsDatagrids.

Closes #1478

1. Check out this PRs' branch and run: `nvm use && npm run start`
2. Go to: http://localhost:4300/ids-data-grid/editable-lookup.html
3. Click on any cell in the "Location Lookup" column.
4. Click on the "search" icon, and a modal should appear with a list of countries.
5. Select a country, then click the "Apply" button.
6. Click away from the edited field
7. Ensure the edited field is populated with the country you selected.


tmcconechy (commented):
@clayinfor this looks very good.

1. Some lint issues (cyclical issue?)
2. Try to set the row height to xxs, xs (try them all) and then focus the field. Notice that the padding of the text changes 

Screenshot 2023-11-13 at 1 05 42 PM

EdwardCoyle (changes requested):
The lookup feature works great, but I think it needs a bit of normalization against how the other cell editors with "external" views behave, specifically around the trigger/indicator icon.

If you compare the new lookup cell editor with [4.x Datagrid's Lookup cells](https://main-enterprise.demo.design.infor.com/components/datagrid/test-editable-lookup.html), or with the date/time cell editors on the provided WC example page, they demonstrate the behavior I'm referencing:

- When hovering over these cells, an indicator icon appears containing the "type" of cell editor.  This indicator also appears when using keyboard navigation to focus the cell (accessibility-friendly).
- There are two areas within the "cell" that can be clicked when the Datagrid is not in "edit mode".  Clicking on the text portion causes the cell to become focused, and for edit mode to begin.  Clicking on the indicator icon causes the same, but will additionally activate the "external view" (in this case, activating the lookup window).

I bring this up because the way it's implemented now is different compared to the other editors, and I fear some pushback from teams expecting the 4.x-style implementation.  However, the current way has some nice things about it (I do like the ability to clear the cell from edit mode instead of going into the lookup). Wondering what we can do to normalize this?

clayinfor:
Thanks for your feedback @EdwardCoyle!  Definitely insightful seeing the way this is done in 4.x...

`IdsDataGridEditor` currently has a `type` property, and the following types currently exist: `input, checkbox, dropdown, timepicker, datepicker, lookup, tree`...  Maybe we could use this value to help define the hover-icon you mentioned?

However, I would say let's stick that in another ticket, so that we can get this ticket out the door?  cc: @tmcconechy

clayinfor:
Actually, never mind--I see other cells already have the icon-hover state @EdwardCoyle  mentioned, so it's not a new thing for WC... so I'll add a hover state for lookup-cells into this ticket. cc: @tmcconechy

tmcconechy (approved):

clayinfor:
hey @tmcconechy, this is good to go... I've added changes to address @EdwardCoyle review suggestions.

tmcconechy (approved):

EdwardCoyle (approved):
@clayinfor Thanks for this, works great now!

PR #1632, opened 2023-11-13
1608 - IdsInput Fix number mask

### LINKED ISSUE #1608: IdsInput: mask=number not working as expected
**Describe the bug**
ids-input is accepting letters and special characters even though using mask=number 

**To Reproduce**
1. Go to https://main.wc.design.infor.com/ids-input/example.html
2. Click on any input field it accepts letters and special characters even though using mask=number

**Expected behavior**
Input field should not accept letters and special characters when mask=number is used.

**Version**
**"ids-enterprise-wc": "^1.0.0-beta.15"**

**Screenshots**
image

**Platform**
 - Infor Application/Team : ION

andrivarhanov:
I'm not sure what exactly caused the issue, haven't found any recent changes that can be related to the issue, but I recall It was working fine not a while ago.
Made the native input allowing only value that went through the mask

Closes #1608 

- pull and build
- go to http://localhost:4300/ids-input/example.html
- check `Phone Number` and `Number` examples
- see that you can enter only characters allowed in the mask
- go to http://localhost:4300/ids-mask/localized-numbers.html
- see that all examples are correct


tmcconechy (approved):
Yeah wierd. Must have been some of the angular changes for forms but also could not inpoint it. Thanks for the fix anyways @andrivarhanov

EdwardCoyle (approved):

PR #1646, opened 2023-11-15
1538 - Remove duplicate/incorrectly-placed tooltips on IdsDropdown

### LINKED ISSUE #1538: IdsDropdown: Tooltip dropdown shows two tooltips when hovering mouse over dropdown option
**Describe the bug**
After selecting an option from the dropdown and attempting to select another option, when the mouse hovers over one of the options in the dropdown, two tooltips show up: one that is already in the selected state and one that is in the dropdown options

**To Reproduce**

1. Go to https://main.wc.design.infor.com/ids-dropdown/example.html
2. Click on "Dropdown with Tooltips"
3. Select an option
4. Click on "Dropdown with Tooltips" again
5. Hover mouse over another option
6. See error

**Expected behavior**
The tooltip that should show is the tooltip of the option where the mouse hovers over

**Version**
- ids-enterprise-wc: latest

**Screenshots**

In this example, the mouse is hovering over Option Three but shows tooltip for the current selected option as well
image

**Platform**
 - Browser Name: Any

**Additional context**

EdwardCoyle:
This PR fixes some issues with IdsTooltipMixin/IdsDropdown integration which was previously causing duplicate and/or incorrectly-placed tooltips:
- Reuse/remove tooltips when a new one is generated from IdsTooltipMixin
- Refactors the way IdsTooltipMixin sets the tooltip's target element to be configurable per component
- Moves some logic for placing tooltips from IdsTooltipMixin into the correct components (mostly IdsDropdown)
- Adds some logic into IdsDropdownList for keeping track of the last-known hovered list item, and uses this reference to place the tooltip correctly.
- Adds a new Dropdown tooltip demo page

Closes #1538

- Pull/build/run
- Open http://localhost:4300/ids-dropdown/example.html
- Hover the "Dropdown with Tooltips" field while closed.  It should produce a tooltip containing the text "Additional Info on Option Two".  The option here should always reflect the selected item
- Click the Dropdown list to open it.  Hover over each list item.  The tooltip should be placed directly over top of each item and read "Additional Info on [whichever Option is hovered]"
- See that there are no duplicate tooltips presented.  In a dev tools inspector, no IdsTooltip elements should be present in the IdsContainer once hovering stops.
- Open http://localhost:4300/ids-dropdown/tooltips.html
- Hover the first dropdown while its closed.  a Tooltip should display with the text content of the selected item since it's cut off by ellipsis.
- Open the menu and hover each item.  Only the items cut off by the list should display a tooltip.
- Choose a "short" item.  When the list closes, hover the dropdown element.  It shouldn't display a tooltip since the list item is fully displayed.
- Hover the second dropdown and list items.  All should be displaying tooltips since they all have additional information attached.


src/components/ids-dropdown/ids-dropdown-list.ts
```diff
@@ -116,6 +120,19 @@ export default class IdsDropdownList extends Base {
         this.triggerSelectedEvent();
       }
     });
+
+    this.offEvent('mouseover.dropdown-list-box');
+    this.onEvent('mouseover.dropdown-list-box', this.listBox, (e: any) => {
```
tmcconechy, line 125: @EdwardCoyle behaviorwise i think it would be better to use the `hoverend` event which "debounces" the hover. The test is if you pass the mouse across all the elements when its open at a moderate speed then it doesnt show the tooltip until you "stop" on one. Althought before it didnt even do that (compared to https://main.wc.design.infor.com/ids-tooltip/example.html)

But give it a try changing this to hoverend.

tmcconechy: Ok, strange. Figured it should work in most cases i saw. What about instead debouncing it at least so it mimics the behavior? If not we can live with it but just checking...

EdwardCoyle: @tmcconechy I've resolved the issue of the debounce on the Dropdown List tooltip generally, but as it turns out this event handler is not responsible for that part.  This handler only deals with storage of the "last hovered" list option, so I needed to keep this one in.  

The `hoverend` part is still what drives the tooltip display and I was able to augment that in the Tooltip Mixin to switch to using `mouseover/mouseout` for this case, which does work with the debounce 👍

tmcconechy (approved):
Ok sounds good. I introduced a merge conflict FYI via merging some PRs

tmcconechy:
Sorry made another one. Can just delete that snap

EdwardCoyle:
@tmcconechy The same lint errors from your PR are present here so let's merge that one first.  This one is rebased against all recent PRs except that one

tmcconechy (approved):

clayinfor (approved):

jdilag90 (approved):

PR #1654, opened 2023-11-20
1573- IdsDataGrid tree grid virtual scrolling refactor

### LINKED ISSUE #1573: IdsDatagrid.scrollRowIntoView() does not work on tree grid
**Describe the bug**
When a large number of rows are loaded on idsDatagrid tree grid, idsDatagrid.scrollRowIntoView() does not work as it should. It has to be executed a number of times to go to the intended row/row index.

**To Reproduce**

1. clone enterprise-wc, npm i, npm start
2. open http://localhost:4300/ids-data-grid/tree-grid-virtual-scroll-infinite-scroll.html
3. scroll down to row # 514
4. open dev tools - console
5. run `document.querySelector("#tree-grid-virtual-scroll").scrollRowIntoView(349)  ` to go to row `#350`
6. Error: datagrid did not go to the intended row `(#350)`

**Expected behavior**
Row `#350` should be visible

**Version**

- ids-enterprise-wc beta 16

**Screenshots**

Video:

https://github.com/infor-design/enterprise-wc/assets/53523076/802c1ab0-0c01-48a4-8d4b-42014f0b1404

**Platform**
OS Version: Windows 11
Browser Name: chrome
Browser Version: Version Version Version 118.0.5993.89 (Official Build) (64-bit)

**Additional context**
ScrollRowIntoView() needs to executed a number of times before the row shows up on the datagrid.

### LINKED ISSUE #1587: IdsDatagrid: Tree grid moves/scrolls rows after appendDataset()
**Describe the bug**
The datagrid moves/scrolls row after appendDataset() is executed. This happens on a large number of rows or when all rows are collapsed. In the tree-grid-virtual-scroll-infinite-scroll.html example, the rows are moved after scrolling at the end to get new records.

**To Reproduce**

1. clone entreprise-wc, npm i, npm start
2. open tree-grid-virtual-scroll-infinite-scroll.html
3. click collapse all
4. scroll down to the end to load new data, take note of the last row before and after new data is loaded
5. repeat a couple of times until the expected last row changes
6. error: last row is moved after appending new data

**Expected behavior**
When appendDataset() is called, the last row should not move, only the scrollbar should move/resize indicating that new records where added.

**Version**

- ids-enterprise-wc beta 16

**Screenshots**

https://github.com/infor-design/enterprise-wc/assets/53523076/25402f95-2846-4e23-be47-eee268d04b48

**Platform**
OS Version: Windows 11
Browser Name: chrome
Browser Version: Version Version Version 118.0.5993.89 (Official Build) (64-bit)

**Additional context**

jdilag90:
This PR adds a virtual scroll/row recycling algorithm specific to virtual scroll on tree grids.
The current virtual scroll algo doesn't account for dynamic rows (hidden sub rows) in its calculation, which led to 
problems like a jumpy scrollbar, a scrollbar handle that doesn't track with the mouse, seemingly empty grid body with large nested datasets, and broken `scrollRowIntoView` and `rowStart` settings/methods. 

This PR adds tree virtual scrolling along side the default virtual scrolling, so the current working non-tree virtual scrolling should work as is.

Fixes #1573 
Fixes #1587 

1. Checkout branch
2. Go to http://localhost:4300/ids-data-grid/tree-grid-virtual-scroll-infinite-scroll.html

***Test dynamic row rendering***
1. Collapse/Expand any of the rows
2. Child rows that are collapsed/hidden should not be rendered in DOM (they're recycled to become other visible rows)
3. Child rows should keep their collapsed/expanded/selected states when their parents are expanded/collapsed
4. Number of rows in DOM should never exceed 100 (default max rows)
5. You can check dom row count via `$('ids-data-grid').rows.length`

***Check arias & row-index attributes***
1. Check that  grid row `aria-setsize`, `aria-level`, `ara-posinset`, `aria-expanded` are correct
2. The `row-index` and `data-index` of tree grid row attributes should reflect their actual data index in `$('ids-data-grid').data`

***Test infinite scrolling***
1. Scroll to the bottom to trigger an infinite scroll
3. The datagrid should not jump/shift up/down when infinite scroll is triggered
4. Test infinite scroll after clicking `Expand All`/`Collapse All`, there should be no `jumpiness`
5. Test that you are able to infinite scroll to the end of data. There are a total of `2100 rows` to load

***Test scrolling***
1. Check that the scrollbar size is correct on page load
2. Check that the scrollbar is correctly resized as you're expanding/collapsing and infinite scrolling rows
3. Scroll up/down the grid by dragging the scrollbar handle, the scroll handle should track with your mouse

***Test scrollRowIntoView***
1. For ease of testing, there's an added input at the top right of the [example page](http://localhost:4300/ids-data-grid/tree-grid-virtual-scroll-infinite-scroll.html) to `scrollRowIntoView()`
2. Input various row indexes and the grid should correctly scroll to that row
3. If the target row is within a collapsed parent, scrolling to that row should internally expand the parent row and make the target row visible

***Test `scrollMaxRows` setting***
1. This setting allows user to define virtual scroll max rows in dom. Default is 100
2. Change max rows via `$('ids-data-grid').scrollMaxRows = 50;`
3. The datagrid row count should immediately adjust to the new max scroll value

***Smoke test other virtual scroll examples***
Please give these a quick look to ensure they're still working ok
* http://localhost:4300/ids-data-grid/tree-grid.html
* http://localhost:4300/ids-data-grid/tree-grid-virtual-scroll.html
* http://localhost:4300/ids-data-grid/virtual-scroll.html
* http://localhost:4300/ids-data-grid/virtual-scroll-infinite-scroll.html
* http://localhost:4300/ids-data-grid/virtual-scroll-infinite-scroll-m3.html
* http://localhost:4300/ids-data-grid/virtual-scroll-row-start.html
* http://localhost:4300/ids-data-grid/tree-grid-expand-collapse.html

**Additional**
There could be some optimizations made in the future to improve performance. 
For example, we could re introduce the use of `requestAnimationFrame` during scrolling.
For tree-grids, `get virtualRows()` could be cached and only updated when rows are expanded/collapsed/added/removed.


tmcconechy (commented):
@jdilag90 i did see something a bit wierd im wondering about

1) load page and hit expand all / collapse all -> it seems to change row visible - but maybe thats the best it can do. Maybe cleaner if we go to top first (but just as an example)
2) Collapse all rows -> then scroll to the bottom then keep scrolling in this case it seems to just shuffle down only a little at a time. Maybe this is ok too?

http://localhost:4300/ids-data-grid/tree-grid-virtual-scroll.html works very well

jdilag90:
@tmcconechy thanks for the feedback

1) I think the reason the scroll position jumps when using collapse/expand all is because there's a big difference in row count.
Collapse All makes the visible row count go from 420 rows to 50 rows.
Expand All makes the visible row count go from 420 rows to 1050 rows.
I can make it so that expand/collapse all scrolls to the top, let me know

2) The space created at scroll bottom is the infinite scroll appending rows.
I pushed a commit doubling the rows appended on infinite scroll so there's more space to scroll down.
It's up to the user on how many rows they want to append when infinite scroll is triggered.

tmcconechy (approved):
OK thanks. I figured it was probably something like that.

EdwardCoyle (approved):
Looks great @jdilag90!

PR #1662, opened 2023-11-27
1544- IdsCalendar event form popup cutoff

### LINKED ISSUE #1544: IdsCalendar: Modal/bottom border is cut off
**Describe the bug**
Modal/bottom border is cut off.

**To Reproduce**

1. Go to https://main.wc.design.infor.com/ids-calendar/example.html
2. Resize to a smaller window
3. Click any events
4. See the bottom part of the window

**Expected behavior**
Modal/bottom border should not be cut off.

**Version**

**Screenshots**
image

**Platform**
 - All major browsers

jdilag90:
Fix calendar event form popup styles when user is on a smaller screen.

Closes #1544 

1. Checkout branch
2. Go to http://localhost:4300/ids-calendar/example.html
3. Resize browser to smaller dimensions (500 x 650)
4. Click on a calendar event to open form popup
5. Check that the problem shown below is fixed

274827144-f2f64e6d-f075-4446-8088-5c1d31b163b8


EdwardCoyle (approved):

tmcconechy (changes requested):
@jdilag90 i think a problem with the `contain: content;` approach is now the popups are cut off over the buttons.

you can see this with the datepicker but i think it would happen with any popup component like dropdown or others

Anything else we can do you can think of?

Screenshot 2023-12-04 at 12 45 13 PM

jdilag90:
@tmcconechy thanks Tim good catch! 

I updated the styles so that it overflow scrolls the form content section only and removed `content: contain`

jdilag90:
@tmcconechy in your screenshot with the datepicker looking cut off, you should be able to scroll down to see the rest of the popup now

tmcconechy:
ok i see @jdilag90 that is interesting because when the modal is not mobile it does show over the buttons but in this mobile case it doesnt

Also it adds space a little much for the time picker

Screenshot 2023-12-04 at 4 36 59 PM

If this is all you think of it seems like its just on this dialog? But wondering why it doesnt use the modal entirely or if there is an issue there?

tmcconechy (approved):
👍🏻

PR #1665, opened 2023-11-29
1664 - IdsDataGrid empty message height

### LINKED ISSUE #1664: Datagrid: Virtual-scroll height and sticky-header issues when data set to empty-array
@see [original MS Teams post by Mikolaj](https://teams.microsoft.com/l/message/19:b649a7c426e64e3dbedb0cc5202ce2d4@thread.skype/1701260072055?tenantId=457d5685-0467-4d05-b23b-8f817adda47c&groupId=4f50ef7d-e88d-4ccb-98ca-65f26e57fe35&parentMessageId=1701260072055&teamName=IDS%20Enterprise%20Development&channelName=Web%20Components&createdTime=1701260072055)

To summarize the problem: While virtual-scrolling, the data-grid's header loses stickiness and its height becomes full-screen after setting data to an empty-array, and then repopulating data.

**Describe the bug**
Hi Web Components, I am facing issues with data grid when using virtual scroll feature. I have enterprise-wc-examples repository locally. 

I have simple code that is loading data into grid component. It should render cells with "First data fetch" text, after 5 seconds it should render cells with "Second data fetch" text, after next 5 second it should render empty data grid and finally after next 5 seconds it should render cells with "Third data fetch" text. When using virtual scroll feature something goes wrong after setting grid data to empty array. If we set data again then column header that was always sticky to the top and visible disappears as we scroll. Additionally the grid fills the entire height of the page even though I set the height to 500px. Without setting virtual-scroll to "true" in HTML code everything works fine. I plan to use this component for a large dataset, so this virtual scroll feature is important to me. I pasted code below. I will try to provide a video in the comment.
Html code: 
```html
<div style="height: 500px; width: 100%; ">
    <ids-data-grid #dataGrid id="idsDataGridTest"
                   auto-fit="true"
                   list-style="true"
                   virtual-scroll="true"
                   row-height="xs">
    </ids-data-grid>
</div>
```
 

Component code: 
```typescript
export class ExampleComponent implements AfterViewInit {
    @ViewChild('dataGrid', {read: ElementRef}) dataGrid;
    public columns = [];
    ngAfterViewInit(): void {
        this.dataGrid.nativeElement.columns = [{
            id: 'value',
            name: 'value',
            field: 'value',
            formatter: this.dataGrid.nativeElement.formatters.text,
            filterType: this.dataGrid.nativeElement.filters.text
        }];
        this.dataGrid.nativeElement.data = [...Array.from(Array(100).keys()).map((x) => ({value: `First data fetch value: ${x}`}))];
        setTimeout(() => {
            this.dataGrid.nativeElement.data = [...Array.from(Array(100).keys()).map((x) => ({value: `Second data fetch value: ${x}`}))];
        }, 5000)
        setTimeout(() => {
            this.dataGrid.nativeElement.data = [];
        }, 10000)
        setTimeout(() => {
            this.dataGrid.nativeElement.data = [...Array.from(Array(100).keys()).map((x) => ({value: `Third data fetch value: ${x}`}))];
        }, 15000)
    }
}
```

**To Reproduce**

Steps to reproduce the behavior (see screen recording):

https://github.com/infor-design/enterprise-wc/assets/93789065/bc9413f0-e370-4f1d-a159-8635f3605bfc

**Expected behavior**

https://github.com/infor-design/enterprise-wc/assets/93789065/f9a9e141-a344-4751-867a-4ea16d0f7d37

jdilag90:
Fixes an issue where the datagrid container loses its original inline style height when empty message mode is enabled.
https://github.com/infor-design/enterprise-wc/blob/349398009a9e081244c44d3c82044ad13442cb52/src/components/ids-data-grid/ids-data-grid-empty-message.ts#L115
I moved the line above to css where `height: unset !important` is only applied when ids-data-grid `.has-empty-message`.
The inline height applied by `auto-fit` should remain intact

Closes #1664 

1. Checkout branch
2. Go to http://localhost:4300/ids-data-grid/virtual-scroll.html
3. Run `const copy = $('ids-data-grid').data;` to create data copy
4. Run `$('ids-data-grid').data = [];` to make datagrid empty
5. Run `$('ids-data-grid').data = copy;` to re populate datagrid
6. Check that datagrid virtual-scroll still works ok
7. Also test on http://localhost:4300/ids-data-grid/example.html, to check non-virtual scroll examples

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -26,6 +26,7 @@
 - `[DataGrid]` Fix virtual scrolling for tree grids. ([#1573](https://github.com/infor-design/enterprise-wc/issues/1573))/([#1587](https://github.com/infor-design/enterprise-wc/issues/1587))
 - `[DataGrid]` Add `filterAlign` setting to columns for independently aligning filter row contents. ([#1575](https://github.com/infor-design/enterprise-wc/issues/1575))
 - `[DataGrid]` Fix an issue where inline-editable cells in the first column were not considered when navigating the grid in edit mode with the Tab key. ([#1616](https://github.com/infor-design/enterprise-wc/issues/1616))
+- `[DataGrid]` Fix an issue where empty message permanently removes the datagrid container height. ([#1664](https://github.com/infor-design/enterprise-wc/issues/1664))
 - `[DatePicker]` Set trigger field's value with today's date when today button is clicked. ([#1614](https://github.com/infor-design/enterprise-wc/issues/1614))
 - `[DatePicker]` Fixed time picker value change is not reflected dynamically in date time picker trigger field. ([#1576](https://github.com/infor-design/enterprise-wc/issues/1576))
 - `[FlexLayout]` Added a flex example using IdsButtons and IdsInputs. ([#1395](https://github.com/infor-design/enterprise-wc/issues/1395))
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -1,5 +1,11 @@
 # What's New with Enterprise Web Components
 
+## 1.0.0-beta.18
+
+### 1.0.0-beta.18 Fixes
+
+- `[DataGrid]` Fix an issue where empty message permanently removes the datagrid container height. ([#1664](https://github.com/infor-design/enterprise-wc/issues/1664))
+
 ## 1.0.0-beta.17
 
 ### 1.0.0-beta.17 Breaking Changes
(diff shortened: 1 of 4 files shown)


EdwardCoyle (approved):
This looks great, thanks @jdilag90!  For sake of being thorough, I checked these as well -- all seems good:

- http://localhost:4300/ids-data-grid/empty-message.html
- http://localhost:4300/ids-data-grid/empty-message-custom-thru-slot.html
- http://localhost:4300/ids-data-grid/empty-message-custom-thru-settings.html

tmcconechy (changes requested):
I think might be a bit more here.

1. when i run step `$('ids-data-grid').data = [];` first changes the size of the grid, this is probably an issue with the `auto-fit` setting, but as you change back and forth from the copy and [] you see it change
2. the empty message doesnt seem centered in auto-fit. Even on http://localhost:4300/ids-data-grid/empty-message.html it looks a little high up.

Can we take a look?

EdwardCoyle (approved):
@jdilag90 looks good to me

tmcconechy (approved):

PR #1668, opened 2023-12-01
1553 - IdsValidationMixin: Allow required indicator display separate from validation rules

### LINKED ISSUE #1553: IdsInput:  need 'required field' asterisk with customized validation message
Need to localize the 'required' validation message. We are using addValidationMessage() but then there is no sudo ui '*'. Tried to add 'validate="required" ' but got two validation messages.

It would be great if we could customize the 'required' message while keeping the asterisk. 

**Version**
ids-enterprise-wc: 1.0.0-beta.15

**Platform**
WFM

EdwardCoyle:
This PR makes some adjustments to IdsValidationMixin, creating a `required` boolean attribute that can be set outside of using `validate="required"` 

Closes #1553

- Pull/build/run
- Open http://localhost:4300/ids-input/validation-required.html
- See that the bottom 3 input fields have a required validation indicator.  The 2nd and 4th fields down are configured to display the required indicator simply by using `<ids-input required>`.
- Remove the text from all 4 fields.  Only the 3rd and 4th fields down display a validation message.  The 4th field down has custom validation applied (which mimics the required validation rule but is NOT the built-in "required" rule), and shows as "required" due to the new attribute only.

Also review other required examples for accuracy.  Some I've found are:
- http://localhost:4300/ids-input/example.html (Last name and Email fields)
- http://localhost:4300/ids-trigger-field/example.html (Photos field)
- http://localhost:4300/ids-dropdown/example.html ("required" Dropdown)

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -14,6 +14,7 @@
 - `[Tree]` Added `beforeExpanded` async loading pattern to dynamically add tree children. ([#1516](https://github.com/infor-design/enterprise-wc/issues/1516))
 - `[Tree]` Renamed `useToggleTarget` to `expandTarget`, now available as a setting to toggle tree nodes only when clicking the icon. ([#1528](https://github.com/infor-design/enterprise-wc/issues/1528))
 - `[Tests]` Changed test framework to playwright. ([#1225](https://github.com/infor-design/enterprise-wc/issues/1225))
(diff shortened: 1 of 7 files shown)


clayinfor (approved):

tmcconechy (dismissed):
@EdwardCoyle i added a couple docs

PR #1694, opened 2023-12-13
1520 - Add full page loading indicator

### LINKED ISSUE #1520: IdsLoadingIndicator: Add example of blocking the UI full page
- For IDS loading indicator doesn't have the property for blocking the backend UI.
- **For Ex:** I have added Soho busy indicator block UI for the reference (below snippet)
image
- Take the above snippet as a reference same functionality is  expected to work for **circular & linear ids-loading-indicator**
-  If possible provide an example of  **circular & linear** ids-loading-indicators and how to make use of this functionality in angular.

**use case:**
-  when the ids-loading-indicator appears on the complete page it needs to block the backend UI.
- when a user does some action there is some ongoing communication between the UI and the backend, the user should be blocked from doing other activities until the response is received from the backend.  This is a common use case in any application.
- Please consider this as a priority enhancement

tmcconechy:

Adds an example and settings `overlay` and `backgroundColor` and `align="center"` so that a full page loading indicator may be shown.

Also:

- made moduleNav easier to import
- add settings to IdsOverlay
- convert tests

Fixes #1520 

- http://localhost:4300/ids-loading-indicator/example.html should still work ok
- see http://localhost:4300/ids-loading-indicator/full-page.html -> page should be covered and loading indicator centered
- try themes to do this run `$('ids-loading-indicator').stop()` in the console then change theme to dark/contrast then run `$('ids-loading-indicator').start()`


jdilag90 (approved):

tmcconechy:
@talkaboutdesign did you want to check this out? For the Web components we dont have the branch system so have to check locale? Would like your feedback here

EdwardCoyle (approved):

PR #1700, opened 2023-12-15
1697 - Trigger popup menu examples

### LINKED ISSUE #1697: IdsPopupMenu: Cannot make IdsTriggerField use an IdsPopupMenu
ids-popup-menu in combination with ids-trigger-field
I am triggering ids-popup menu for click event of an icon in input trigger field. But ids-popup menu is opening even if I click anywhere in a page.

1. Please use the below code to reproduce the issue
```
<ids-trigger-field
  size="mm"
  label="Date Field"
  field-height="md"
  aria-haspopup="menu"
>
  <ids-trigger-button slot="trigger-end">
    <ids-text audible="true">Activation Date</ids-text>
    <ids-icon id="trigger-field" icon="schedule"></ids-icon>
  </ids-trigger-button>
</ids-trigger-field>
<ids-popup-menu
  class="popupmenu-position"
  id="my-menu"
  align="bottom, left"
  target="trigger-field"
  trigger-type="click"
  role="none"
>
  <ids-menu-group role="group">
    <ids-menu-item id="item-one"
    >Last hours
    </ids-menu-item>
    <ids-menu-item id="item-one"
    >Last days
    </ids-menu-item>
  </ids-menu-group>
</ids-popup-menu>
```

2. After running the project go to the component page where this code is placed and try to click anywhere in the screen the pop up menu is opening which is not expected. popup should be opened only when clicked on the icon in trigger-field

**Additional context**
"ids-enterprise-wc": "^1.0.0-beta.17"

tmcconechy:

Adds an example for support of a trigger field with a popupmenu that opens on click.

Fixes #1697

- go to http://localhost:4300/ids-trigger-field/example.html
- click on the trigger field in "Menu Trigger Field"
- should open a menu and line up


jdilag90:
IdsPopup has settings for `y` and `arrow`. Maybe worth adding those settings to IdsPopupMenu to pass into IdsPopup?

https://github.com/infor-design/enterprise-wc/blob/5c2d9733122f89b722c3c3ebe81cb511bc5a2a91/src/components/ids-popup/demos/example.html#L21

tmcconechy:
Yeah ok. I got a bit lazy there.Me and @EdwardCoyle discussed that too. Let me see if i can get that in as it is slightly better fix since you can do it with no JS. But may have to punt that.

tmcconechy:
@jdilag90 @EdwardCoyle I made x, y and arrow pass through options instead as suggested. Example updated to use markup but the behavior should be the same. A lot cleaner this way 👍🏻

jdilag90 (approved):

EdwardCoyle (approved):

PR #1723, opened 2023-12-26
1543 - Add fixed to listview events

### LINKED ISSUE #1543: ListView: Row Click Event is Selecting the row item and Checkbox Too in Angular
**Issue Description:**
**Multiselect list view:**
-  On clicking the list row(multi-select), observed that two click events are triggered, one for the checkbox and another row selection event. ( in HTML used **(click)** event both check box & row selected).
- observe below snippet=> for **(click)**:
image
-  The same thing was tried by using **(**selected**)** only one event triggered but both the checkbox and row were selected in this scenario. Observe the below snippet. 
image

**Mixed Select List View:**
-   On clicking the list row(mixed-select), observed that two click events are triggered, and the checkbox is not selected. ( in HTML **(click)** event used, row only selected). observe below snippet
image
- The same thing was tried by using **(selected)** no events were triggered for the row selection. Observe the below snippet. 
image
-  On clicking the checkbox(mixed-select), observed that two click events are triggered, and the row is not selected. ( in HTML **(click)** event used). observe the below snippet.
image
-  The same thing was tried by using **(selected)**, one event was triggered for the checkbox selection. Observe the below snippet. 
image
**Query:** For mixed row selection is there a separate click event property available?

**Please refer to the below video:**
https://github.com/infor-design/enterprise-wc/assets/105051938/dd32d8a5-c474-42fb-80ef-8dee8b5d6ce3

**Reproduction steps:**
- Open multi & mixed select list view example in ids-wc examples angular application.
-  Add a click event for the list and inside the method put one console log. (you can use (click) and (selected)
- observed the above-described issue.
**Expectations:**
* separate click events will need to be triggered for the checkbox and row selection.
* Only one event will trigger either the checkbox or row selection

tmcconechy:

Improve selection methods in list view and ensure events fire all at the right time and not duplicated. Also changed aria-activedescendent to use an ID instead of a number. And added docs and tests around these events.

- also made a few api changes (activated vs active) and before/after/activated vs itemactivated
- and removed some tests that do not need any conversion

@ashachodavarapu  once this is fixed i think what you want is mixed selection mode so only the checkbox works and one item can be activated. This is cleaned up after this fix.

Fixes #1543

- test http://localhost:4300/ids-list-view/selectable-single.html and watch console for clicking and space and enter on the keyboard
- test http://localhost:4300/ids-list-view/selectable-mixed.html and watch console for clicking and space and enter on the keyboard
- test http://localhost:4300/ids-list-view/selectable-multiple.html and watch console for clicking and space and enter on the keyboard
- smoke test http://localhost:4300/ids-list-view
- check docs for keyboard info


EdwardCoyle (dismissed):
I'm thinking this looks good, but one thing of note:  If you try to use the Enter key on http://localhost:4300/ids-list-view/selectable-multiple.html (or any multi-select example) you can only select records, but not deselect records.  Spacebar does select AND deselect however.  Not sure if that's intentional

tmcconechy:
@EdwardCoyle your correct thats an issue. I have fixed it now

EdwardCoyle (approved):

clayinfor (approved):

PR #1735, opened 2024-01-02
1679 - Add lookup search and related features

### LINKED ISSUE #1679: IdsLookup: Have input above the datagrid like soho-lookup
In Experience Designer we are using the soho-lookup with the [source] to search for data. Current issues with soho-lookup are:

1. You need to have the pager, if we say `paging:false` the lookup source is not working
2. The data is filtered based on what you've typed in the input, which does not work for us since we are doing a search, where the input is a query like: `Dg* SortBy:ITNO` which would gives us e.g. 68 results back, but nothing visible in the actual datagrid, since it will try to filter the list for what we've typed.

1. Be able to have lookup with a searchfield above the datagrid
3. Be able to opt out from automatic filtering by the datagrid on the input value
4. If opted out, ids code shouldn't format/make the text lowercase (it breaks our search queries), which I assume it does in order to do the filtering/compare text values
5. Be able to have column filters on the datagrid (seems to already be supported)
6. Be able to not have to have a `pager` in order for the `source` to work. At least we want to be able to hide it.
7. Be able to not have the more menu for row height (perhaps already possible)

To sum up: 
We want to have the following, but with an input field above the grid like before, where the inputted text is passed on enter (`searched`) to the source function (not modified/lowercase) for us to use in our api calls. Also, without having to have a pager like soho-lookup:

https://main.wc.design.infor.com/ids-lookup/filter.html
image

- Filing issue on ids-enterprise-ng... :)

**Additional context**
In the following screenshot I've:
- Manually hidden the `.toolbar-pager` with css
- Formatted the filterExpr[0].value

image

For soho-lookup, the current code workaround we have (which isn't 100% good for us) is:
* Format the filterExpr[0].value, since soho makes this lowercase and we need it to still be e.g. `SortBy:ITNO`
* When we get the data back, remove things like `ITNO:123 SortBy:ITNO` -> `123`, but it is not fully working for us since the user might have something like `ITNO:123 ITDS:DG* SortBy:ITNO` and in this case, I can't really return `123 DG` since that would yield 0 results after grid is automatically filtered (even if we do have records form API), so the user is really limited to only one `KEY:VALUE` parameter as of now. 

```ts
readonly source = (
    req: TypeSafeSohoDataGridSourceRequest,
    response: SohoDataGridResponseFunction
  ): void => {
    const emptyParameters: MpParameters = { record: {}, records: [] };

    if (req.trigger === 'searched') {
      const query = req.filterExpr[0]?.value || ''; // 👈 takes what has been typed in the input

      // Need to format since soho is making all input text lowercase
      const formattedQuery = this.formatSearchQuery(query); // 👈 format query back to e.g. sortby -> SortBy

      this.getDataset(formattedQuery)
        .pipe(
          tap((dataset) => {
            // Needed to remove search query parts from the filterExpr value
            const value = this.removeSearchQueryParts(formattedQuery); // 👈 remove things like SortBy:ITNO
            req.filterExpr[0].value = value; // 👈 add it back here
            
            req.total = dataset.length;
            response(dataset, req);
          })
        )
        .subscribe();
    }
```

tmcconechy:

Adds a new feature to lookup `searchable` that allows search. Also added:

- fixes to the selection events firing at wrong times
- restructure the event names
- fix a bug with the change event firing on parent elements when in a template
- add a search count setting

Fixes #1679 

- go to new example http://localhost:4300/ids-lookup/searchable.html
- open the lookup
- type for example "Gre" and hit enter. see that the events in the console fire and the grid updates
- Click the x see that the events in the console fire and the grid resets
- review all lookup examples http://localhost:4300/ids-lookup


EdwardCoyle (approved):
Paging currently doesn't work on http://localhost:4300/ids-lookup/pagination-client.html, but I'm betting my PR will fix that.   Otherwise this looks great 👍

tmcconechy:
Thanks yeah i didnt cover http://localhost:4300/ids-lookup/pagination-client.html but when you showed me it didnt work, i figured you had it.

jdilag90 (approved):

PR #1739, opened 2024-01-04
1565 - Retain selection of IdsDataGrid rows across paged datasets

### LINKED ISSUE #1565: IdsDataGrid: Select across pages / contextual toolbar?
**Describe the Issue**
Doesn't seem to be a way to allow selecting across pages? I can't find any setting for that, and it doesn't work out of the box (using server side paging). Also, is there any alternative to the contextual toolbar available in 4.x?

Select across pages 4.x
https://design.infor.com/code/ids-enterprise/latest/demo/components/datagrid/example-paging.html
1. Select one row
2. Go to other page
3. Go back to first page
Result: Selection kept
Web component: Selection not kept (or, it seems it's actually selected, but selection column shows unchecked)

Contextual toolbar 4.x
https://design.infor.com/code/ids-enterprise/latest/demo/components/datagrid/example-multiselect.html
1. Select one or more rows
Result: A contextual toolbar appears which contains actions for the selected row(s)

Is there any plan to support these two features, or is there any alternative?

I guess we can achieve/keep the contextual toolbar simply by toggling it ourselves based on the selection, and make it animate as in 4.x.

So, the selection issue is more important out of the two.

EdwardCoyle:
This PR improves the handling of selected rows on IdsDataGrid to retain their state when paging is involved.  It's now possible to switch pages, both in client-side and server-side paging scenarios, and retain a selected state on rows previously selected on the component.

In order to achieve this, the following changes were made:
- IdsDataSource has been updated with a setting for adding a primary key (defaults to `id`), as well as methods for creating, updating, and deleting records.
- Additionally methods were added for forcing the update of the internally stored "cached" dataset, and the `pagination` setting from IdsDataGrid was pushed down to the IdsDatasource where it can be used more broadly, and to influence how slices of data are gathered in different paging scenarios (server side paging works on a slice of the overall data, whereas client side works against the entire dataset)
- In IdsDataGrid, there's now a store for selected records that is accessible in `server-side` paging mode that retains records and corresponding IDs.
- In IdsDataGrid, `columnId` maps to IdsDataSource's primary key and both will change when set.
- Some paging bugs were fixed.
- Added some tests

Closes #1565

Pull/build/run, then test the following:

_Client side selection and CRUD operations:_
- Open http://localhost:4300/ids-data-grid/pagination-client-side.html
- Try any combination of page number, page size, and selection.  When navigating to a page that previously had selected records, the same records should be selected.
- The text on the top/left of the Datagrid toolbar should reflect the correct number of records across pages (this is not built into the grid, but is managed by select/deselect events emitted from the grid).
- Try using the "Add" button to add records at the end of the grid.  The page should change to the last one, and the record should have a unique ID.
- Try selecting records, then using the "clear" and "delete" icons at the top-right of the datagrid to manipulate them.

_Server side selection:_
- Open http://localhost:4300/ids-data-grid/pagination-server-side.html
- Try any combination of page number, page size, and selection.  When navigating to a page that previously had selected records, the same records should be selected.
- The text on the top/left of the Datagrid toolbar should reflect the correct number of records across pages (this is not built into the grid, but is managed by select/deselect events emitted from the grid).

_Tree grid + paging + selection:_
- Open http://localhost:4300/ids-data-grid/tree-grid-paging.html
- Use the "+" on the first row to expand its child row with the name "Battery" (has ID 1.1).  Select this record
- Page forward and backward a few times.  This record should stay selected.
- Also note that other records named "Battery" should not be selected, unless chosen.

_Also smoke test these examples for accuracy:_
- http://localhost:4300/ids-data-grid/tree-grid.html
- http://localhost:4300/ids-data-grid/virtual-scroll.html
- http://localhost:4300/ids-data-grid/tree-grid-virtual-scroll.html


tmcconechy:
@EdwardCoyle looks really good. I did find a couple things:

A)
- go to http://localhost:4300/ids-data-grid/tree-grid.html
- click first row (building one) to select and then expand it
- now try to select building 4 (error: "Cannot read properties of undefined (reading 'id')")

B)
Also on http://localhost:4300/ids-data-grid/tree-grid-virtual-scroll.html and http://localhost:4300/ids-data-grid/tree-grid-paging.html
- expand row one (crawler)
- try to select it and the child isnt selected (i did not check but maybe select child rows is off here? if it is not an issue) 

C)
- should the selected rows text be translateable? (can use ids-text for that) - Can use same text "2 Selected"

jdilag90:
@tmcconechy @EdwardCoyle 
For issue `B)` that Tim mentioned, it looks like the child rows aren't selected because the setting `group-selects-children="true"` isn't set on the datagrid. 
I tried it on tree-grid-virtual-scroll and tree-grid-paging example example, and it seems to work as expected

Other than Tim's comment A), Looks good to me!

tmcconechy:
OK i should have checked. Its just A and C for me then. Thanks!

EdwardCoyle:
@tmcconechy @jdilag90 I've fixed comment A, and raised #1740 to handle comment C at a later time.

jdilag90 (approved):

tmcconechy (approved):

PR #1743, opened 2024-01-08
1564 - IdsCalendar legend content projection

### LINKED ISSUE #1564: IdsCalendar: Support for proper legend content projection
Currently, the ids-calendar component supports content projection for the desktop legend but not for the mobile one. I've noticed that the desktop and mobile legends are quite different and come from different components.

For the desktop legend:

image
The legend resides inside the ids-calendar component but not within any specific view.

For the mobile legend:
image
The legend comes from the month-view component.

We're planning to create a customized legend. However, due to the differences between the desktop and mobile legends and the lack of content projection support for the mobile legend, implementation becomes complex.

Ideally, both desktop and mobile views should use the same legend. One solution would be to allow clients to use the current mobile legend for both views and also support content projection for the mobile legend. This way, we would only need to implement one customized legend for both views.

jdilag90:
Added slot `custom-legend` to calendar to allow users to inject their own legends.
Added `eventsrendered` event to provide user's current rendered calendar event data
Created example with a custom legend showing event types and their count of occurrence in the current rendered view.

Closes #1564 

1. Checkout branch 
2. Go to http://localhost:4300/ids-calendar/custom-legend.html
3. Check that the custom legend is rendered without errors
4. Resize the browser between mobile/desktop sizes, check that the custom legend looks ok
5. Switch views/change dates and check that the custom legend event type counts change depending on rendered events in view


tmcconechy (commented):
Looks good but some minor suggestions

1. Alphabetize the new change log item
2. Check the text alignment on the legend (or we could add the one they showed in their design if thats not too much work but im ok with not doing that if it is). 

Screenshot 2024-01-09 at 12 34 23 PM

PR #1748, opened 2024-01-09
1290 - Fix to set menu arrow property

### LINKED ISSUE #1290: IdsPopupMenu: Add pass through option to remove arrow in `ids-popup-menu` component
None

There is no option for `ids-popup-menu` component to remove the arrow pointing to the target element.
image
We want to have this to make it easier for us to create dropdown look when clicking Tabs or other elements. Here's a sample reference from the enterprise controls we want to achieve with this web component: https://main-enterprise.demo.design.infor.com/components/tabs/example-dropdown-tabs.html

None

**Additional context**

tmcconechy:

Some previous work i added some pass through options from ids-popup-menu. One of them was arrow which is needed for the use case on #1290 however this needed one small tweak to make it work for this use case.

Fixes #1290 

- go to http://localhost:4300/ids-menu-button/example.html
- click new + button, should not show an arrow

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -1,5 +1,11 @@
 # What's New with Enterprise Web Components
 
+## 1.0.0-beta.19
+
+### 1.0.0-beta.19 Features
+
+- `[PopupMenu]` Added ability to set `arrow` on the IdsPopupMenu so that it filters down to IdsPopup underneath, can now disable or change arrow direction. ([#1290](https://github.com/infor-design/enterprise-wc/issues/1290))
+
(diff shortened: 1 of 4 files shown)


EdwardCoyle:
This seems to be interfering with the other examples too - I noticed the first two buttons also don't have an arrow but they should IIRC:

Screenshot 2024-01-09 at 3 42 09 PM

tmcconechy:
@EdwardCoyle thanks i fixed that, woops

EdwardCoyle (approved):

jdilag90 (approved):

Issue #1672, opened 2023-12-04
IdsPie: Unable to hide legend + unable use legend-placement attribute


volante007:
**Describe the bug**
Two issues related to IdsPie (in my case, used as donut)

1. There doesn't seem to be a way to hide the legend, which was possible in 4.x using the `showLegend` option. In a 1x1 widget, it might make sense to hide it.
2. Can't set legend placement using the attribute, unless binding it (Angular, similar to https://github.com/infor-design/enterprise-wc/issues/1625)
```
<!-- Doesn't work -->
<ids-pie-chart legend-placement="bottom"></ids-pie-chart>

<!-- Works -->
<ids-pie-chart [attr.legend-placement]="'bottom'"></ids-pie-chart>
<ids-pie-chart [legendPlacement]="'bottom'"></ids-pie-chart>
```

The second might be a generic problem in many components when used with Angular, but not sure how to best verify/catch them all?

jbrcna:
QA Passed
<img width="1269" alt="image" src="https://github.com/infor-design/enterprise-wc/assets/56722184/f8d74429-7f82-4888-a08c-66fcdcb6045d">
<img width="609" alt="image" src="https://github.com/infor-design/enterprise-wc/assets/56722184/b8629800-1d43-4fe1-9ec9-cca1a08ac66f">
<img width="532" alt="image" src="https://github.com/infor-design/enterprise-wc/assets/56722184/a8d7bae6-7b1b-4863-8559-6af596a62c99">
<img width="616" alt="image" src="https://github.com/infor-design/enterprise-wc/assets/56722184/1510ec74-f238-4c8d-9ce7-2cbd1ecd5ba7">
<img width="721" alt="image" src="https://github.com/infor-design/enterprise-wc/assets/56722184/bcec7eb1-c717-4e7f-8958-4a5fc34c6692">

PR #1753, opened 2024-01-10
1672 - IdsChartLegendMixin: Add 'none' setting, change usage in IdsPieChart

EdwardCoyle:
This PR changes the way the `legend-placement` attribute works in IdsChartLegendMixin, which solves a couple issues:
- Removes the need to default the legend placement to 'right', which fixes a problem when trying to set the attribute in the Angular examples
- Also naturally creates a default where legends aren't present in charts using the new `none` value.

Related to #1672

Pull/build/run, then smoke test the following chart example pages.  There should be no visual discrepancies:
- http://localhost:4300/ids-area-chart
- http://localhost:4300/ids-axis-chart
- http://localhost:4300/ids-bar-chart
- http://localhost:4300/ids-line-chart
- http://localhost:4300/ids-pie-chart

_Test new settings:_
- Open http://localhost:4300/ids-pie-chart/legend-placement.html
- By default, no legend should be present
- Try setting the different values in the radio group to change the visibility/layout of the legend

_Test in Angular examples:_
- Run `npm run publish:link` on this branch, and test with infor-design/enterprise-wc-examples#52


tmcconechy:
@EdwardCoyle 

1. what about keeping bottom (or right for pie) as the defaults? im ok if not but more often than not the legend would be there
2. also in the pie example the top and bottom legend the "swatch" could be the same size as the left and right legend it looks incorrect

Screenshot 2024-01-11 at 1 49 16 PM

EdwardCoyle:
We could try defaulting it to `bottom` or `right`, but the issue in the Angular components was that running `.setAttribute('legend-placement', 'right')` in the constructors wasn't appearing to allow binding in Angular (since it would get automatically set to the default).  Having an "off/empty" default helps, but maybe we can just make better detection for `none`

src/components/ids-pie-chart/ids-pie-chart.ts
```diff
@@ -111,7 +111,6 @@ export default class IdsPieChart extends Base {
    */
   connectedCallback(): void {
     super.connectedCallback?.();
-    this.legendPlacement = 'right';
```
EdwardCoyle, line 114: @tmcconechy this was part of the Angular binding problem

tmcconechy (approved):

PR #1783, opened 2024-01-19
1588 - datagrid fix for left/right arrow-keys to work inside editable-cell

### LINKED ISSUE #1588: IdsDatagrid: Left/right arrow keys does not work inside a text editable cell
**Describe the bug**
On a text editable cell, the user cannot navigate on the text value by using the left/right arrow keys. The focus goes to the next cell.

**To Reproduce**

1. clone enterprise-wc, npm i, npm start
2. open http://localhost:4300/ids-data-grid/editable-inline.html
3. click on the first editable cell
4. once the value is highlighted, press left to go to the first letter
5. Error: the focus goes to the next cell

**Expected behavior**
The expected result should be to navigate using left/right arrow keys inside the text editable cell (similar to excel or windows explorer when updating cell value or renaming a file). Once finished with the edit (by pressing enter or escape), then the arrow keys will navigate within the cells. 

**Version**
- ids-enterprise-wc beta 16

**Platform**
 - Browser Name: Any
 
**Additional context**

clayinfor:

On a text editable cell, the user could not navigate on the text value by using the left/right arrow keys. The focus goes to the next cell.

Closes #1588 

1. Checkout this branch and run: `nvm use && npm run start`
2. Go to: http://localhost:4300/ids-data-grid/editable.html
3. Verify arrow-keys, escape and enter keys work.
4. Go to: http://localhost:4300/ids-data-grid/editable-inline.html
5. Verify arrow-keys, escape and enter keys work.
6. Escape/Enter keys should stop cell-edit and allow arrow-keys to navigate datagrid cells and rows.


tmcconechy (changes requested):
@clayinfor looks good but there is two keyboard "behaviors" that have become now problematic but some of it will change here https://github.com/infor-design/enterprise-wc/issues/1682

We wanted to copy how that works. But the two things we should maybe look at now are

1. click in # cell, then arrow right to the editable cell. Start to type "a" for example and as you type it wil auto enter editable mode. But the value will change wrong 101 -> becomes 1101. I think instead it should select all text and type over it (so 101 becomes 1)
2. The tab is not working. Focus the first editable cell now tab and it should go to the next editable field. This works on https://main.wc.design.infor.com/ids-data-grid/editable-inline.html but poorly

Also tabing doesnt work on http://localhost:4300/ids-data-grid/editable.html (compared to https://main.wc.design.infor.com/ids-data-grid/editable.html)

With those i think we will be in a good place with this

clayinfor:
hey @tmcconechy , I updated the PR to get it closer to what you suggested in #1682 ... there's still some room for improvement, but I figure whoever gets that ticket can take it across the finish line =)

tmcconechy (approved):

tmcconechy (approved):

tmcconechy (approved):

tmcconechy (approved):

jdilag90 (approved):

PR #1832, opened 2024-02-09
1638 - ids-text dynamic tooltip

clayinfor:

- Max-width and overflow(ellipsis) attributes are not applied to the IDS text, able to see the full text.
- If the tooltip text is passed dynamically unable to see the tooltip in UI.
**NOTE**:  If the tooltip text is passed statically by using the tooltip attribute it's working fine.

1. Check out this branch and run: `nvm use && npm i && npm run dev`
2. Go to http://localhost:4300/ids-text/example.html
3. Hover over the "Lorem ipsum" `ids-text` at the very bottom and ensure tooltip appears.


tmcconechy (approved):
@clayinfor looks good, not entirely necessary but..

1) could we put this example with the overflow on its own page?
2) For the longer tooltip can add maxWidth setting so it looks a bit better

Might not wait for those however but if you can
Thanks

jdilag90 (approved):

PR #1840, opened 2024-02-12
1765 - IdsInput/IdsTextarea prevent multiple input events

### LINKED ISSUE #1765: IdsInput/IdsTextarea: issues with how event is fired in web components
**Describe the bug**
`input event` in for example `ids-input` is fired twice whenever a change happens. For plain HTML text input the event is only fired one time, and that's the expected behaviour.

In the case of Angular, you can use `ngModel` and bind it with `ngModelChange` with e.g. a function,  which will fire each time the model is changed. When using it with web components, it will fire one time after the component has been loaded. When you use `ngModel` with plain HTML text input it does not fire on load, which is the expected behaviour. 

You can see the ngModel issue in your example project

**To Reproduce**

You can see the ngModel issue here
https://github.com/infor-design/enterprise-wc-examples/blob/main/angular-ids-wc/src/app/components/ids-forms/demos/example/example.component.html

1. build and run the angular examples
2. go to the IDS forms example

**Expected behavior**

For the `ngModel` case:
- During load it should to run the function binded with `ngModelChange`

For `input` event case:
- It should not fire twice whenever a input happened.
- It should not fire input change when you have set the value attribute

jdilag90:
This PR suppresses the native InputEvent so that only the webcomponent's CustomEvent fires.

Closes #1765 

1. Checkout branch
2. Go to http://localhost:4300/ids-textarea/example.html & http://localhost:4300/ids-input/example.html
3. In console, add an event listener for `input` to any of the input/textarea examples 
4. `$('ids-input').addEventListener('input', (evt) => console.log('Input Handler', evt));`
5. `$('ids-textarea').addEventListener('input', (evt) => console.log('Input Handler', evt));`
6. Check that the `input` event only fires once per character

***Test in Angular Forms***
1. Copy over this branch's build to the angular examples project via `npm run publish:link` or copy/pasting `dist`
2. Go to http://localhost:4200/ids-forms/example & http://localhost:4200/ids-reactive-forms/example
3. Check that these form examples still work when updating and submitting forms

***Test individual event binding in Angular***
1. Bind to an ids-input's `input` event
2. Inside [IdsInput's Component Example Class](https://github.com/infor-design/enterprise-wc-examples/blob/main/angular-ids-wc/src/app/components/ids-input/demos/example/example.component.ts), add the `log()` function below
image
3. Inside [IdsInput's Component Example HTML](https://github.com/infor-design/enterprise-wc-examples/blob/main/angular-ids-wc/src/app/components/ids-input/demos/example/example.component.html), add input event binding
image
4. Type a value and check that the `input` event only logs once per character in console
5. Do the same for the `ids-textarea` angular example


tmcconechy (dismissed):
Looks good 👍🏻

clayinfor (approved):

Issue #1871, opened 2024-02-19
Tests: Convert Button Tests


tmcconechy:
**Describe the task**
Convert card tests from https://github.com/infor-design/enterprise-wc/tree/main/tests-to-convert/ids-button to https://github.com/infor-design/enterprise-wc/blob/main/tests/ids-button/ids-button.spec.ts

**Additional context**
N/A

PR #1895, opened 2024-02-26
1871 - IdsButton Test Coversion

n-ace-ancog:
Converted test scripts of `ids-button` component to Playwright.

Closes enhancement [#1871](https://github.com/infor-design/enterprise-wc/issues/1871)

- Test should pass
- No flaky result

### DIFF
--- a/tests/ids-button/ids-button.spec.ts
+++ b/tests/ids-button/ids-button.spec.ts
@@ -4,12 +4,14 @@ import { expect } from '@playwright/test';
 import { test } from '../base-fixture';
 
 import IdsButton from '../../src/components/ids-button/ids-button';
+import { IdsButtonAppearance, IdsButtonIconAlignment } from '../../src/components/ids-button/ids-button-common';
 
 test.describe('IdsButton tests', () => {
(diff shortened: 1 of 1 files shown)


tmcconechy (commented):
A few suggestions.

a) Delete the tests from tests-to-convert that have been converted.
b) Name the tests as if you were describing the scenario in it to a non-programmer familiar with the problem domain
c) `ID` could be `id` (super minor but its a bit mixed)
d) Not sure what i think of so many `test.steps` since i never used it before? But at least if there is only one step we may not need it. Probably have to discuss.
e) dont need `then` because await is already doing this.
f) also not sure why it looks like every single file changed in percy?


Nice work though. Would like to see how much this converted but need to delete the tests from the other file, or if all of them then remove the whole file from tests-to-convert

Just giving feedback here so a bit more comments initially

n-ace-ancog:
@tmcconechy, It seems like I used an outdated version and I feel its safer to just create another PR with the updated branch and your review points. Also, this was the caused for Percy to fail.

PR #1898, opened 2024-02-27
1871 - IdsButton Test Conversion

n-ace-ancog:
Converted test scripts of `ids-button` component to Playwright.

Closes enhancement https://github.com/infor-design/enterprise-wc/issues/1871

- Test should pass
- No flaky result

Updated to the latest branch and addresses review points from #1895.

In connection with the review points from https://github.com/infor-design/enterprise-wc/pull/1895#pullrequestreview-1901203168:

> a) Delete the tests from tests-to-convert that have been converted. **_DONE_**
b) Name the tests as if you were describing the scenario in it to a non-programmer familiar with the problem domain **_Renamed tests to be informative as possible to non-programmers**
c) ID could be id (super minor but its a bit mixed) **_DONE_**
d) Not sure what i think of so many test.steps since i never used it before? But at least if there is only one step we may not need it. Probably have to discuss. **_I've reduced the number of test.steps to lines which needs to be "grouped" together_**
e) dont need then because await is already doing this. **_DONE_**
f) also not sure why it looks like every single file changed in percy? **_Copied from a very outdated version of main hence the new pull request and branch_**


tests/ids-button/ids-button.spec.ts
```diff
@@ -4,12 +4,14 @@ import { expect } from '@playwright/test';
...
 
 import IdsButton from '../../src/components/ids-button/ids-button';
+import { IdsButtonAppearance, IdsButtonIconAlignment, IdsButtonContentAlignment } from '../../src/components/ids-button/ids-button-common';
 
 test.describe('IdsButton tests', () => {
   const url = '/ids-button/example.html';
 
   test.beforeEach(async ({ page }) => {
     await page.goto(url);
+    await page.waitForLoadState();
```
tmcconechy, line 14: Not sure if this is needed. Did you confirm? Might just take extra time

tmcconechy (commented):
I think this looks good. Thanks @n-ace-ancog

tests/ids-button/ids-button.spec.ts
n-ace-ancog: You're right, seems like that the `goto` has already built in waiting for the `load` event to fire this makes the `waitForLoadState` redundant in this case. 

image

andrivarhanov (approved):

ericangeles (approved):
Looks good.

PR #1949, opened 2024-03-04
Create ids-locale.spec.ts

AAlviar:
** Adding folder for locale**


tmcconechy:
Empty files wont work. I think we dont need to add this right now. Until we convert the tests.

PR #2017, opened 2024-03-08
Chore: Add await lint rule

tmcconechy:

To avoid confusing when writing tests of when you need an `awaits` added a lint rule. That will enforce the right place that awaits are needed.

Fixed several places it was needed but skipped some in the code (mainly setters i didnt want to change to async.

NA

- lint passes


tmcconechy:
@n-ace-ancog FYI this should help enforce where await is needed with a lint rule

jdilag90 (approved):

PR #2222, opened 2024-04-15
2192 - IdsWizard Disabled steps

### LINKED ISSUE #2192: IdsWizard: Add `disabled` property to `ids-step` for the wizard
**Describe the bug**
Hello Team, We would like use ids-wizard(web component) & show to user if any particular step is disabled or not visually. Just like below example using ids-enterprise-ng.
https://stackblitz.com/edit/ids-quick-start-1730-icuaaq?file=src%2Fapp%2Fapp.component.html
 
I don't see `disabled` property on `ids-step`.  Even though we have clickable property which makes step not clickable but it doesn't have styling to show visually if step is clickable or not.

**To Reproduce**

We don't have any example with disabled step: https://main.wc.design.infor.com/ids-wizard/sandbox.html

**Expected behavior**

Expected Styling: (step1 is disabled)
Screenshot 2024-04-08 121211

Current Styling:
Screenshot 2024-04-08 121957

**Version**

-     "ids-enterprise-wc": "^1.0.0-beta.22",

andrivarhanov:
Added disabled attribute to the steps

Closes #2192 

- pull and build
- go to http://localhost:4300/ids-wizard/sandbox.html
- check `Steps 1, 3 and 5 are disabled` example
- click on 1, 3 or 5 step and see the active step doesn't change
- click on 2 or 4 steps and see the active step changes


tmcconechy (commented):
@andrivarhanov 

1. More important (but not critical) I think if the step is filled it should be in lighter "disabled" blue. Try ids-color-action-weaker or  ids-color-action-weakest

Screenshot 2024-04-15 at 11 02 07 AM

2. Less important but good to add if we are changing this anyways. Its prefereable to add component specific tokens and use the semantic tokens. i.e.

- add ids-wizard-disabled-background-color,  ids-wizard-disabled-checked-background-color in here https://github.com/infor-design/enterprise-wc/blob/main/src/themes/default/ids-theme-default-core.scss#L1717

They would match to ids-color-action-weaker or  ids-color-action-weakest and probably this semantic token https://github.com/infor-design/ids-foundation/blob/main/tokens/theme-soho/semantic-light.scss#L16

3. Can we add a theme switcher to that sandbox page to test?

Due to the release might just do these later

PR #2237, opened 2024-04-17
2173 - Autocomplete popup

### LINKED ISSUE #2173: IdsSearchField : Search suggestion are not visible
**Describe the bug**
IdsSearchField does not show suggestions. I was able to observe popup opening related errors in console.

**To Reproduce**
1. Add ids-search-field web component
2. provide dataset with data in component
3. start application
4. Observe behavior

`<ids-search-field
        [data]="dataset$ | async"
        id="search-field-autocomplete"
        label-state="collapsed"
        placeholder="Search"
        caps-lock="true"
        align="start"
        search-field="label"
        autocomplete
        clearable
    ></ids-search-field>`

**Expected behavior**
We should be seeing suggestions in popup menu. However, I was observing errors in console.

**Version**
- ids-enterprise: 1.0.0-beta.21

**Screenshots**
image

image

**Platform**
 - Infor Application/Team Name: Infor Birst ETLP
 - Device: MacBook
 - OS Version: MacOS 14.4.1
 - Browser Name: chrome
 - Browser Version: 123.0.6312.106

andrivarhanov:
Fixed the issue when the popup is not being attached if the autocomplete setting is set to non-autocomplete component. Spotted in the angular examples, can be reproduced in the web components

Closes #2173

- pull and build
- go to http://localhost:4300/ids-search-field/example.html
- run it in the console `$('ids-search-field').autocomplete = true`
- check `Product` required example
- enter some text and see `No results` dropdown
- same can be done for `ids-input` `ids-trigger-field` `ids-lookup`


jdilag90:
I see the `No results` dropdown in all example except ids-lookup
1) I went to this example http://localhost:4300/ids-lookup/example.html
2) I ran `$('ids-lookup').autocomplete = true`

andrivarhanov:
fixed for ids-lookup

tmcconechy (changes requested):
1. There is a minor styling issue here: 
Screenshot 2024-04-19 at 3 01 56 PM

a) should use variables like these (FYI i am fixing in dropdown so if dropdown applies can ignore) https://github.com/infor-design/enterprise-wc/blob/main/src/themes/default/ids-theme-default-core.scss#L650-L654
b) can we add two new examples to this page one with no results and one with actual data to search , looking at the bug report again seems like he is adding actual data but cant test that this is working
c) also noticed some are different size? (maybe some reason on purpose but worth a quick look)
Screenshot 2024-04-19 at 3 04 18 PM

andrivarhanov:
@tmcconechy 
1. It looks like dropdown styles inherited from dropdown and input styles inherited from input, the only difference search field is border radius 4px, not sure if it's by design or we're going to make it like input
2. Modified `Autocomplete Search Bar` example to be autocomplete=false initially and enabling it after data is loaded which should cover the case where the issue was before the fix, added `Autocomplete Search Bar (No data)` example -> gives No results on any input
3. Fixed (button's top/bottom margins)

updated changelog

tmcconechy (dismissed):

andrivarhanov:
@tmcconechy can't figure out why the test is failing randomly, had to skip so we can merge

andrivarhanov:
fixed the test

tmcconechy (dismissed):

tmcconechy (approved):

PR #2274, opened 2024-04-30
2227 - IdsTree redraw when assigning empty array

### LINKED ISSUE #2227: IdsTree: Unable to redraw tree when assigning empty array
**Describe the bug**
When setting data property of IdsTree with an empty array, it does not rerender or redraw the tree and only shows the previous data visually. Inspecting the actual data property reveals that it is empty and properly set.

**To Reproduce**
1. Go to https://stackblitz.com/edit/ids-enterprise-wc-v1?file=index.html
2. Open console for additional info
3. Click on 'Set one item' and the tree should only have one node left
4. Click 'Reset' to bring back the nodes
5. Click 'Delete all nodes'
6. The tree still displays the nodes

**Expected behavior**
The tree should not display anything

**Version**
ids-enterprise-wc@1.0.0

**Screenshots**
IdsTree_empty_array

jmacaluso711:
This PR fixes a bug where the tree did not redraw when assigning an empty array

closes #2227 

- pull and run branch
- go to http://localhost:4300/ids-tree/reset.html
- click on 'Set one item' - see one node left in the tree
- Click 'Reset' to bring back the nodes
- Click 'Delete all nodes' and all nodes will be removed
- Check console for messages to corresponding actions

~- [ ] Some documentation for the feature.~
~- [ ] A test for the bug or feature.~

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -16,6 +16,7 @@
 - `[BarChart]` Converted bar chart tests to playwright. ([#1919](https://github.com/infor-design/enterprise-wc/issues/1919))
 - `[BreadCrumb/Hyperlink]` Fix focus state on click bug. ([#2238](https://github.com/infor-design/enterprise-wc/issues/2238))
 - `[Editor]` Converted editor tests to playwright. ([#1931](https://github.com/infor-design/enterprise-wc/issues/1931))
+- `[Tree]` Fixed bug where redraw did not trigger when assigning an empty array. ([#2227](https://github.com/infor-design/enterprise-wc/issues/2227))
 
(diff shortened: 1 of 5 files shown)


tmcconechy (commented):
Looks good. Maybe with the change we should check what happens with:

`data = []`
`data = undefined`
`data = null`
`data = ''`

Might be all those should set it to empty. Also looks like a test needs a look

jmacaluso711:
@tmcconechy I made that change. If you test in the console $('ids-tree').data = '' | null | undefined. The tree will clear

tmcconechy (approved):

clayinfor (approved):

PR #2287, opened 2024-05-03
1983 - Convert ids-toast test to playwright

### LINKED ISSUE #1983: Tests: Convert Toast Tests
**Describe the task**
Convert Toast tests from https://github.com/infor-design/enterprise-wc/tree/main/tests-to-convert/ids-toast to https://github.com/infor-design/enterprise-wc/blob/main/tests/ids-toast/ids-toast.spec.ts

**Additional context**
N/A

jbrcna:

closes #1983 

Run the test suite
Check the code conversion
New converted tests should pass

**Note:**
Test with TODO comments have errors.


tmcconechy:
@jbrcna this looks good overall but

a) seems like some lint errors. The page object is not used
b) similar to above instead of `await toast.evaluate` i think we should use `await page.evaluate` like other tests
c) 3 is failing for me

```
  3 failed
    [chromium] › ids-toast/ids-toast.spec.ts:390:9 › IdsToast tests › functionality tests › can  restore saved position 
    [chromium] › ids-toast/ids-toast.spec.ts:417:9 › IdsToast tests › functionality tests › can  reset saved position 
    [chromium] › ids-toast/ids-toast.spec.ts:446:9 › IdsToast tests › functionality tests › can  clear saved position
```

tests/ids-toast/ids-toast.spec.ts
```diff
@@ -2,12 +2,26 @@ import AxeBuilder from '@axe-core/playwright';
...
 test.describe('IdsToast tests', () => {
   const url = '/ids-toast/example.html';
+  let toast: any;
+  let options: any;
 
   test.beforeEach(async ({ page }) => {
     await page.goto(url);
+    toast = await page.locator('ids-toast').first();
+    const button = await page.locator('button[aria-label="Toast Message"]').first();
+    await button.click();
```
n-ace-ancog, line 24: The `button.click()` in `beforeEach` here requires the rest of the test cases inside the `test.describe` to follow a specific action even when not needed. Either isolate the button clicking on each individual test case using `await page.locator('#btn-toast-demo').click()` or group them better.

tmcconechy: Yes i agree @n-ace-ancog i always use `await page.locator(` only in the individual test. good call

tests/ids-toast/ids-toast.spec.ts
```diff
@@ -45,4 +59,775 @@ test.describe('IdsToast tests', () => {
...
+        toastEl.show(opt);
+        const toastContainer = toastEl.toastContainer();
+        const messageEl = toastContainer.querySelector('ids-toast-message');
+        return messageEl;
+      }, options);
+
+      expect(msg).toBeTruthy();
+    });
+
+    test('can  check if can save position to local storage', async ({ page }) => {
```
n-ace-ancog, line 343: You can use the code below for this test case which check the local storage of the browser. Reference it as well on other test case that includes validating local storage. I've added some comments for explanation and considered that clicking the button to trigger the toast is isolated within the test case. Take note of `await page.waitForFunction(() => window.localStorage.length > 0);` as this is important in retrieving and waiting for the local storage to be filled before validating further.

```ts
test('can check if can save position to local storage', async ({ page }) => {
  // get information from local storage
  const localStorage = async () => {
    const store = await page.evaluate(() => window.localStorage);
    return store;
  };
  const prefix = 'ids-toast-container';
  const suffix = 'usersettings-position';
  const uniqueId = 'some-uniqueid';
  const transform = 'translate(100px, 250px)';

  // check if local storage is empty
  expect(await localStorage()).toEqual({});
  // button is declared and located on the immediate test.describe
  await button.click();
  await toast.evaluate((element: IdsToast, opt) => {
    element.uniqueId = opt.uniqueId;
    element.savePosition = true;
    element.draggable = true;
    element.toastContainer()!.style.transform = opt.transform;
  }, { uniqueId, transform });
  // wait for the toast to be attached in DOM - very important!!
  await toast.waitFor({ state: 'attached' });
  // wait for the localStorage to have an item - very important!!
  await page.waitForFunction(() => window.localStorage.length > 0);
  // validation if information is saved in local storage
  const storage = await localStorage();
  expect(storage.hasOwnProperty(`${prefix}-${uniqueId}-${suffix}`)).toBeTruthy();
  expect(storage[`${prefix}-${uniqueId}-${suffix}`]).toEqual(transform);
});
```

jbrcna:
I wasn't sure what to do with the 3 failing tests. it's about local storage. tried to fix it bot to no avail. I'm stuck. need help.

tests/ids-toast/ids-toast.spec.ts
jbrcna: it says TypeError: localStorage.setItem is not a function

this also received false. 
`expect(storage.hasOwnProperty(${prefix}-${uniqueId}-${suffix})).`
and undefined
`expect(storage[${prefix}-${uniqueId}-${suffix}]).toEqual(transform);`

tmcconechy: @jbrcna maybe the problem is you can only use localStorage in a `await page.evaluate(() => {});`

tests/ids-toast/ids-toast.spec.ts
```diff
@@ -45,4 +59,845 @@ test.describe('IdsToast tests', () => {
...
+      }
+
+      await setPosition('xyz');
+      toastContainer = await getContainer(DEFAULTS.position);
+      await expect(toast).not.toHaveAttribute('position');
+      await expect(toastPosition).toEqual('top-start'); // DEFAULTS.position causes error
+      await expect(toastContainer).toEqual(true);
+    });
+
+    test('can  set toast to progress bar', async () => {
```
n-ace-ancog, line 202: Remove the extra space between **can** and **set** in the test title. Check the rest of the test titles.

tests/ids-toast/ids-toast.spec.ts
```diff
@@ -45,4 +59,845 @@ test.describe('IdsToast tests', () => {
...
+      };
+      const prefix = 'ids-toast-container';
+      const suffix = 'usersettings-position';
+      const uniqueId = 'some-uniqueid';
+      const transform = 'translate(100px, 250px)';
+
+      // check if local storage is empty
+      expect(await localStorage()).toEqual({});
+      await await button.click();
+      await toast.evaluate((element: IdsToast, opt) => {
```
n-ace-ancog, line 367: There's a lint error in the `opt` variable. Just explicitly state it as any like `opt:any`.

tests/ids-toast/ids-toast.spec.ts
```diff
@@ -45,4 +59,845 @@ test.describe('IdsToast tests', () => {
...
+      await toast.waitFor({ state: 'attached' });
+      // wait for the localStorage to have an item - very important!!
+      await page.waitForFunction(() => window.localStorage.length > 0);
+      // validation if information is saved in local storage
+      const storage = await localStorage();
+      expect(storage.hasOwnProperty(`${prefix}-${uniqueId}-${suffix}`)).toBeTruthy();
+      expect(storage[`${prefix}-${uniqueId}-${suffix}`]).toEqual(transform);
+    });
+
+    // test('can  check if can save position to local storage', async ({ page }) => {
```
n-ace-ancog, line 383: If not needed, you can remove this block of code that is commented out.

tests/ids-toast/ids-toast.spec.ts
```diff
@@ -45,4 +59,845 @@ test.describe('IdsToast tests', () => {
...
+      expect(await localStorage()).toEqual({});
+      await button.click();
+      options = {
+        title: 'Test',
+        message: 'Some test text',
+        timeout: 10
+      };
+      const uniqueId = 'some-uniqueid';
+      const transform = 'translate(100px, 250px)';
+      await localStorage.setItem(id(uniqueId), transform);
```
n-ace-ancog, line 449: The issue here is you wanted to use the [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) from the window api, but instead refers to the `localStorage` function declared at the top of the test case, hence the error. As what @tmcconechy, mentioned on the previous review - you can only use the [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) inside a `page/locator.evaluate` method to `setItem`.

You can use the code below for the restoring of saved position.

```typescript
test('can restore saved position', async ({ page }) => {
  const prefix = 'ids-toast-container';
  const suffix = 'usersettings-position';
  const uniqueId = 'some-uniqueid';
  const key = `${prefix}-${uniqueId}-${suffix}`;
  const transform = 'translate(-596px, 232px)';
  await page.evaluate((storage) => { window.localStorage.setItem(storage.key, storage.transform); }, { key, transform });
  // wait for the localStorage to have an item - very important!!
  await page.waitForFunction(() => window.localStorage.length > 0);
  await page.evaluate((uId) => {
    const elem = document.createElement('ids-toast')! as IdsToast;
    elem.setAttribute('id', 'toasty-example');
    elem.uniqueId = uId;
    elem.savePosition = true;
    elem.draggable = true;
    document.querySelector('ids-container')!.appendChild(elem);
    elem.show({
      title: 'Test',
      message: 'Some test text'
    });
  }, uniqueId);
  const theToast = await page.locator('#toasty-example');
  await theToast.waitFor({ state: 'attached' });
  await expect(theToast).toBeAttached();
  await expect(await theToast.locator('ids-draggable')).toHaveAttribute('style', `transform: ${transform};`);
});
```

n-ace-ancog (changes requested):

tests/ids-toast/ids-toast.spec.ts
jbrcna: I was able to use it w/o page.evaluate (as Storage)
image

tmcconechy:
ok @jbrcna just skip the local storage tests for now.

tests/ids-toast/ids-toast.spec.ts
```diff
@@ -45,4 +59,874 @@ test.describe('IdsToast tests', () => {
...
+      expect(toastContainer).toEqual('none');
+      await settoastContainer(transform);
+      toastContainer = await getContainer();
+      await expect(toastContainer).toEqual(transform);
+      await page.evaluate(async () => {
+        await document.querySelector<IdsToastMessage>('ids-toast-message')?.removeToastMessage();
+      });
+      const toastmsg = await page.locator('ids-toast-message').first();
+      // eslint-disable-next-line no-unreachable-loop
+      while (await toastmsg.count() > 0) { return; }
```
n-ace-ancog, line 655: Either replace this with `expect(toastmsg).not.toBeAttached()` or `await page.waitForFunction(() => !document.querySelector('ids-toast-message'));`. Both waits for the element to be removed with built in timeouts.

ericangeles: waitForFunction would be ideal, since it returns a promise of jshandle (an in-page JS object). Good idea, @n-ace-ancog.

tests/ids-toast/ids-toast.spec.ts
n-ace-ancog: @jbrcna if you were able to use localStorage outside of `page.evaluate` then you are using it with the context of Playwright instead of the browser/page in test. `page.evaluate` lets you run javascript code in the browser itself. I assume, based from your screenshot, the validation passes as its checking for `null` - which will always be `null` since it was not setup in the first place and ran on Playwright's context while `el.clearPosition()` ran on the browser's context.

image
Reference: https://playwright.dev/docs/evaluating#introduction

tests/ids-toast/ids-toast.spec.ts
jbrcna: thanks for suggesting @n-ace-ancog . I've tried using that but gives me error.
expect(toastmsg).not.toBeAttached()
image
await page.waitForFunction(() => !document.querySelector('ids-toast-message'));
image

jbrcna:
Thank you for the code change @n-ace-ancog 👍

n-ace-ancog (approved):

tmcconechy (approved):

PR #2305, opened 2024-05-13
2142 - IdsPersonalize token table

### LINKED ISSUE #2142: IdsPersonalize: Rethink theming page
**Describe the bug**
After doing tokens http://localhost:4300/ids-theme-switcher/theme-builder-minimal.html is broken in that if you change the primary color it doesnt update. This is because mapped tokens do not map to the theme colors any more.

Need to discuss this with design. 

Key Points:

- what should change?
- should we remap these tokens?
- should we add the datagrid where you can change values ect.

**Expected behavior**
It is still possible to map all the tokens but a lot more work for personalization

jmacaluso711:
This PR adds a token table that shows the token names and values as well as the inheritance of each token for each theme.

The following scripts were added to help create the table:
* `./scripts/copy-tokens.js`: copies the tokens from node_modules and moves them to the `./src/themes/tokens/` folder
* `./scripts/generate-theme-data.js`: Reads the theme files and generates a json file containing all the tokens for the given theme. The json files are then written to the `./src/assets/data/themeData/` directory.

Lastly, the data grid color formatter has been modified to except an option called `suppressColorToolip` in order to fix a bug that was breaking virtual scroll.

closes #2142

- Pull and run branch
- go to http://localhost:4300/ids-theme-switcher/token-table.html
- Check each theme and make sure the tokens are accurately represented.


tmcconechy (commented):
@jmacaluso711 here are my comments looks like a good start.

1. Is it necessary to add the files to themes/tokens. Instead you could just pull them from node_modules/ids-foundation? Not a big deal but just asking
2. Im looking at the hierarchy and think it makes sense, but wondering if its better all in one line?
3. Can you make all the columns sortable
4. And make them filterable too with text filters on the name and value and a dropdown filter on the source?
5. Im expecting to see "Core, Semantic, and Component" in a column?

jmacaluso711:
@tmcconechy Moved the table to the IdsPersonalize directory for now - http://localhost:4300/ids-personalize/theme-table.html

tmcconechy (approved):

jdilag90 (approved):

PR #2330, opened 2024-05-17
2289 - Change position default to fixed

### LINKED ISSUE #2289: IdsPopupMenu: Add setting to set the position style
When using `IdsPopupMenu`, the `IdsPopup` that the component uses is absolutely positioned, which causes an issue when the menu has an ascendant which is relatively positioned since the menu will be contained within the ascendant, but the position of the menu seems to be based on the `IdsContainer`, which causes the menu to be clipped in this scenario. Here's an example in OS Portal:

containers
menu

Stackblitz: https://stackblitz.com/edit/ids-enterprise-wc-100-57mcqe?file=index.html

`IdsPopup` has a setting `positionStyle` where you can set how it should be positioned, `IdsPopupMenu` should have the same setting.

1. Setting `overflow-x: auto` on the ascendant that's relatively positioned. This makes it possible to actually see the entire menu, but the scrollbar makes it ugly and awkward.
overflow_auto
2. Manually setting `positionStyle` on the `IdsPopup` via the `IdsPopupMenu`: 
`document.querySelector("ids-popup-menu").popup.positionStyle = "fixed"` (or using viewchild)
This works but 
a) we use Angular and would prefer to not do this. Being able to pass it as an input to the component would be much better: 
`<ids-popup-menu positionStyle="fixed"></ids-popup-menu>`
b) the menu gets misaligned at the top:
misaligned

**Additional context**
And what application do you work on? `OS Portal`

tmcconechy:

By request on the issue, changed `position-style` default to `fixed` for the `ids-popup-menu` only. 

Fixes #2289 

- open http://localhost:4300/ids-popup-menu/position-style.html / bottom of menu should not be cut off
- retest http://localhost:4300/ids-popup examples
- and http://localhost:4300/ids-popup-menu examples


andrivarhanov (approved):
tested by steps from the description as well as tested `position-style.html` file as independent html page, including popup menu changes merged today from the different PR, everything looks good

anhallbe:
@tmcconechy I checked out the main branch & tried it in Portal.

The `<ids-tab-more>` was one of the components that previously didn't work because of this issue. Now, it works slightly better but the fixed positioning together with `top:0; left: 0` still causes the popup to be misplaced:

image

tmcconechy:
@anhallbe thanks will have a look at that case

Issue #2283, opened 2024-05-02
General: Implement locally served fonts


tmcconechy:
**Describe the bug**

1. Implement locally served fonts similar to enterprise
2. Try variable fonts to know how it works there
3. Server them with the demo app(s)
4. Add to ids-foundations

**Additional context**
https://github.com/infor-design/enterprise/pull/8641/files#diff-095f13a1db74c1fcb71deb9b1430d1942cd2e44251ef9271e8afab60fea7c4fb

PR #2345, opened 2024-05-20
WIP - Added variable font

tmcconechy:

Added a variable font for #2283. Not sure if we will go this way so marking WIP.  I wanted to run the visual tests to see how it looks. 

DO NOT MERGE

- pretty much ignore 
- but will check the visual tests after


tmcconechy:
I think this looks like it would work but parking this for now

PR #2349, opened 2024-05-21
8502 - Remove isMobile spec

ericangeles:

This PR removes `isMobile` spec same with the Enterprise.

Closes https://github.com/infor-design/enterprise/issues/8502

- Go to http://localhost:4300/ids-about/standalone-css.html
- There should be no mobile info
- Check the `ts` of about component
- `isMobile` spec should now removed

~~- [ ] Some documentation for the feature.~~
~~- [ ] A test for the bug or feature.~~

### DIFF
--- a/src/components/ids-about/ids-about.scss
+++ b/src/components/ids-about/ids-about.scss
@@ -5,6 +5,14 @@
   contain: content;
 }
 
+.ids-modal {
+  &.ids-about {
+    .ids-modal-content.has-scrollbar {
+      overflow: auto;
+    }
+  }
+}
+
 .ids-about {
   .ids-modal-container {
     max-width: var(--ids-about-width-max);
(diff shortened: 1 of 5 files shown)


tmcconechy (dismissed):
@ericangeles probably will get an error to update the snapshot tests

ericangeles:
@tmcconechy, I've added more fixes. It seems that the specs was not showing because overflow is hidden (it's in the modal component as the first precedence in CSS).

src/components/ids-about/ids-about.scss
```diff
@@ -35,7 +35,7 @@
     border-radius: var(--ids-border-radius-xxs);
     max-height: 199px;
     outline: 0;
-    overflow: auto;
+    overflow: auto !important;
```
tmcconechy, line 38: Can try to not use `!important`? In this repo we have very few

tmcconechy (dismissed):

PR #2594, opened 2024-07-13
2586 -  Added new method `resetDirtyRow` allowing to reset dirty state for an specific row

### LINKED ISSUE #2586: IdsDataGrid: resetDirtyCells for specific row
After editing a record and sending to the server, I need to remove the dirty but just for that record not for the whole grid.

A method like `resetDirtyCells(row|rowIdx)`

```javascript
resetFlagRow: function(rec) {
   var rowIdx = this.store.indexOf(rec),
      row = this.idsDom.rowByIndex(rowIdx),
      flagCls = 'is-dirty';

   Ext.each(row.querySelectorAll('ids-data-grid-cell.' + flagCls), function(cell) {
      cell.classList.remove(flagCls);
   });

   if (flagCls === 'is-dirty' && row.dirtyCells) {
      delete row.dirtyCells;
   }
}
```

fabiojpoli:
When adding the method `resetDirtyRow` it allows to reset dirty state on cells only for a specific row instead of the whole grid. This is interesting when is editing a single record and sync with the server by saving the specific row only. When the server return success, we want to deal with that record only.

Closes #2586

- Run grid editable example
- Change some records line row 1, 2 and 3
- Call the method `grid.resetDirtyRow(1)`
- It will reset only the second row of the grid and other records will keep the dirty state.

- [X] Some documentation for the feature.
- [X] A test for the bug or feature.

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -9,6 +9,7 @@
 - `[Dates]` Added a new `twoDigitYear` setting to set the locale system to set the dates cut over for two digit years, note that two digit years should be avoided. ([#8061](https://github.com/infor-design/enterprise-wc/issues/2425))
 - `[LayoutGrid]` Adds an example of a responsive grid layout inside a splitter component. ([#2411](https://github.com/infor-design/enterprise-wc/issues/2411))
 - `[Switch]` Updated switch design. ([#2347](https://github.com/infor-design/enterprise-wc/issues/2347))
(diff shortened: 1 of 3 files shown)


tmcconechy (approved):

tmcconechy:
@fabiojpoli Nice thanks for this 👍🏻

tmcconechy:
@fabiojpoli  can you sync main and put this in 1.3.1 change log so it will reflect in the right version? Thanks

PR #2601, opened 2024-07-18
1948 - Convert Locale tests to Playwright

### LINKED ISSUE #1948: Tests: Convert Locale Tests
Describe the task
Convert Locale tests from https://github.com/infor-design/enterprise-wc/tree/main/tests-to-convert/ids-locale

Additional context
N/A

n-ace-ancog:
- Converted Locale tests to Playwright
- Moved the tests helper classes to their own files. Now under /tests/helpers/...
- Added `pageDate` object to retrieve the `Date` object from the browser instead from **node.js**
```ts
// if the test case below is executed outside of America/New_York timezone, the results differ
test('compare dates', async ({ page }) => {
   const nodeDate = new Date();
   const pageDate = await page.evaluate(() => new Date());
   // this may pass or fail depending on where or when the execution happens
  expect(nodeDate).toEqual(pageDate);
});

// using pageDate
test('compare dates', async ({ page, pageDate }) => {
   const testData = pageDate.newDate();
   const browserDate = await page.evaluate(() => new Date());
   // this will always pass regardless of timezone
  expect(testData).toEqual(browserDate);
});
```

Closes #1948 

1. Pull this branch, build, and run the app
2. Run the test suite
3. Check the code conversion
4. New converted tests should pass and no flaky test


tmcconechy:
@n-ace-ancog  can you sync main and put this in 1.3.1 or 1.4.0 change log section depending which one is there when its done? Thanks

n-ace-ancog:
@tmcconechy not entirely sure. There are only about more than 5 test cases I remove due to being duplicate and the rest are converted without any changes in the scenario.

In fact, the coverage for `ids-locale` went up

**[main](https://coveralls.io/builds/68780694)**
image

**[this branch](https://coveralls.io/builds/68772811)**
image

Also checked some of the coverage lost, I think they should be covered in their respective components rather than here. For example, the `ids-app-menu-container.ts` lost 66% coverage, but there should be a test case in its component group covering it instead. Let me know your thoughts. 

image

tmcconechy (approved):
@n-ace-ancog great thanks for showing that. maybe the badge i was looking at is not refreshed or something. Looks great! Thank You

tmcconechy (approved):

tmcconechy:
@n-ace-ancog i tried to mergethis since it passed but every locale test fails on main now once i merged.

Screenshot 2024-07-26 at 12 36 14 PM

I skipped the tests for now perhaps you can take another look?

PR #2622, opened 2024-07-23
2271 - Datagrid Add Group By Rows and Personalization Dialog

### LINKED ISSUE #2271: IdsDataGrid: Add Grouped Rows
**Describe the bug**

Add Group By Rows https://main-enterprise.demo.design.infor.com/components/datagrid/example-grouping.html

tmcconechy:

Adds the start or logic for grouping on rows (needs more things in the future but had to stop here). Also added the personalization dialog

Fixes #2271 

- go to http://localhost:4300/ids-data-grid/example.html
- open the ... menu
- try to reorder rows and close/hide them and restore them similar to the previous version (https://main-enterprise.demo.design.infor.com/components/datagrid/example-index.html) 
- go to http://localhost:4300/ids-data-grid/grouped-rows.html
- notice the rows are grouped by type similar to the previous enterprise version (https://main-enterprise.demo.design.infor.com/components/datagrid/example-grouping.html)
- sort / paging ect should work as expected


n-ace-ancog:
Nice Work!

Just a couple of issues I found:
1. Columns which have disabled switch are unable to be dragged/re-order. As a workaround, an enabled column will be dragged before it to be able to re-order.

https://github.com/user-attachments/assets/03986882-4fb2-4bdd-a268-b98c5fa597e4

2. Dragging seems off. Had to click the item first before dragging even if the pointer is in the "dragging" icon.

https://github.com/user-attachments/assets/c4695dd1-451e-493e-8fe6-8d7990879b1a

jbrcna:
in Safari, the carets were in the middle. Also when you move/drag from LTR, the focus is on the 2nd column instead of the hovered column
image

moving/dragging from RTL is fine
image

tmcconechy:
@n-ace-ancog @jbrcna Those are both valid points but

a) The disabled checkbox is correct. I set some to not be reorderable as a configuration option
b) I see the issue with that you have to click the item first. its a bug in the http://localhost:4300/ids-swappable/example.html and would be better if you could just click and drag immediately but i think its seperate issue
c) the safari / RTL bug was existing as i didnt change the drag


So lets make two new issues?

n-ace-ancog:
> @n-ace-ancog @jbrcna Those are both valid points but
> 
> a) The disabled checkbox is correct. I set some to not be reorderable as a configuration option b) I see the issue with that you have to click the item first. its a bug in the http://localhost:4300/ids-swappable/example.html and would be better if you could just click and drag immediately but i think its seperate issue c) the safari / RTL bug was existing as i didnt change the drag
> 
> So lets make two new issues?

On non reorderable columns, they can be forced to be reorderable by dragging one of the reorderable column before the non reorderable or disabled columns. Is this expected?

jbrcna:
@tmcconechy , thanks for clarifying. here's the link for the new issue https://github.com/infor-design/enterprise-wc/issues/2632

Issue #2639, opened 2024-07-26
IdsSearchField: Search icon is misplaced


SAIPRANAY-GANGULA:
**Describe the bug**

**To Reproduce**

1. Go to https://stackblitz.com/edit/ids-enterprise-wc-130-kfvywx?file=index.html
2. observe the search icon being moved right.
![image](https://github.com/user-attachments/assets/6fa0ead5-30e7-4707-ae2c-1f3c242e0595)

**Expected behavior**
icon should placed correctly like in version 1.2.0
![image](https://github.com/user-attachments/assets/6ff893b2-59db-4ba1-b96d-1792fded10e4)

**Version**
    "ids-enterprise-wc": "^1.3.0",

Issue #2505, opened 2024-06-28
Clear Button is visible when collapsible search-field combined with other icons on toolbar


SAIPRANAY-GANGULA:
**Describe the bug**
I'm currently adopting `ids-toolbar` in our project. We have a requirement to show  `ids-search-field` along side few ids-menu-buttons. The problem is when we enter some text close the search-filed `clear-button` is still visible.

**To Reproduce**
1. Go to https://stackblitz.com/edit/ids-enterprise-wc-120-m61j7z?file=index.html
2. open `search-field`,  enter some text & close it
3. Observe that `clear-button` is still visible

**Expected behavior**
Either the `clear-button` should be hidden or don't collapse search-field when we have some text

**Version**
    "ids-enterprise-wc": "^1.2.0",

**Screenshots**
![image](https://github.com/infor-design/enterprise-wc/assets/56952302/84a2b413-46c7-4efd-a25f-9446c141f53c)

PR #2643, opened 2024-07-26
Misc - Added fixes to search field

tmcconechy:

Fixes a layout issue and visible x when closed after expansion. Also updated the log to reflect next version 1.4.0.
Also added a refresh for button prototype classes.

Fixes https://github.com/infor-design/enterprise-wc/issues/2505
Fixes https://github.com/infor-design/enterprise-wc/issues/2639

- go to http://localhost:4300/ids-toolbar/searchfield.html
- expend the search and type in it then let it close -> should not be any x
- the button should be evenly spaced with the other buttons
- smoke test  http://localhost:4300/ids-button

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -9,28 +9,25 @@
 
 ### 1.4.0 Fixes
 
-- `[Themes]` Fixed path for esbuild script to include `themes/` in dist development/production builds. ([#2641](https://github.com/infor-design/enterprise-wc/issues/2641))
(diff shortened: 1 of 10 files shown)


n-ace-ancog (commented):
The clear button now showing properly, but in mobile it seems that the search field is overshooting on the menu button on the left.

image

jbrcna (changes requested):
the field cuts off in smaller viewport
image

jdilag90 (approved):

PR #2756, opened 2024-08-25
Fix/gitlab ci

corpulent:


tmcconechy (changes requested):
@corpulent seems like you removed all the stuff i got working for the tests. Maybe your out of sync?

PR #2783, opened 2024-08-30
2780 - Prevent current item from being clickable

### LINKED ISSUE #2780: Ids-Breadcrumb with ids-hyperlinks: Disabled attribute not working for last item
**Describe the bug**

We are using the `ids-breadcrumb` component with `ids-hyperlinks` on our Angular project for Process Mining. The problem is that with this solution, the last breadcrumb is not being disabled, meaning the styling remains the same instead of being greyed out and the breadcrumb remains clickable. Even when we set this attribute to be always true, all breadcrumbs are disabled except the last one. Is this per design? We would like to disable the last breadcrumb since it represents the current page in our project and should the user should not be able to interact with it
 
**Code example**

```
<ng-container *ngIf="breadcrumbs$ | async as breadcrumbs">
    <ids-breadcrumb #breadcrumb *ngIf="breadcrumbs.length" truncate="true">
        <ids-hyperlink
            *ngFor="let breadcrumb of breadcrumbs; let last = last"
            [attr.disabled]="last"
            [attr.path]="breadcrumb.path"
            text-decoration="hover"
            >{{ breadcrumb.label }}</ids-hyperlink
        >
    </ids-breadcrumb>
</ng-container>
```

**Demo**

[screen-capture (3) 1.webm](https://github.com/user-attachments/assets/c4a9d0b2-53a0-4af1-8f0b-894a0923dd86)

**Expected behavior**

The last breadcrumb hyperlink is disabled (greyed out in styles and not clickable)

**Version**

ids-enterprise-wc: 1.0.0
Angular: 16.2

tmcconechy:

Fixes an issue where the current item was clickable.

Fixes #2780

- go to http://localhost:4200/ids-breadcrumb/example
- try to click the last item (current page)
- should not be clickable/ should not have a cursor / will not put a message in the console when clicked

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### 1.4.3 Fixes
 
+- `[Breadcrumb]` Fixed bug where the current item could be clicked. ([#/2780](https://github.com/infor-design/enterprise/issues//2780))
 - `[ColorPicker]` Fixed bug where custom colors were being overidden. ([#8964](https://github.com/infor-design/enterprise/issues/8964))
 - `[Datagrid]` Fixed bug where the custom select-color is lost during row-recycling/virtual-scrolling. ([#1392](https://github.com/infor-design/enterprise-wc/issues/1392))
 - `[Datagrid]` Fixed bug editor dropdown can't be re-opened after closing. ([#2589](https://github.com/infor-design/enterprise-wc/issues/2589))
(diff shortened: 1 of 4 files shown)


jbrcna (approved):
image

glenlieorillo (approved):
https://github.com/user-attachments/assets/513ce061-f08e-41e9-a8ed-cc8af91735a5

ericangeles (approved):

PR #2801, opened 2024-09-05
2793 - Add validation indicator to IdsTab

### LINKED ISSUE #2793: IdsTabs: Show validation error indicator

`ids-enterprise` automatically shows an error indicator when there is a validation error in a form control inside a tab. This is especially useful inside a Modal where you might not be able to Save / Apply / Submit while a form inside the modal is invalid, but the invalid control is hidden in a tab.

See:
image
https://main-enterprise.demo.design.infor.com/components/tabs/test-modal-error.html

One of:

- Tabs that have an invalid form element shows the error indicator by default, but a developer can opt out through some attribute.
- The behavior is disabled by default, but can be opt-in through an attribute
- There is a slot in the IdsTab where the developer can add their own error icon

Tried adding a `<ids-icon icon="error" color="error" size="small">` inside the `<ids-tab>`, but the styling is hard to get right, which means inconsistencies between apps.

**Additional context**
And what application do you work on? Portal

anhallbe:

Adds an error indicator to IdsTab when its' IdsTabContent contains an invalid input. 

**Related github/jira issue (required)**: Fixes #2793

**Summary:**
- Adds an `error` slot to IdsTab for customization (custom icon, tooltips, legacy IDS compat etc)
- Adds a `has-error` attribute to IdsTab which can be used to display the slotted `error`.
- Show the error in the tab title when there is a validation error inside the tab content.
- The IdsValidateMixin's `validate` event now bubbles, and has stronger typings.

**Related fixes:**
- Adds missing imports in IdsTab (icon & dismiss button)
- Fix vertical alignment of dismiss button

image

- Check out the branch
- `npm start`
- Go to http://localhost:4300/ids-tabs/validation.html
- Go to every test/sample page that contains either form validation or tabs, and make sure they still work.

### DIFF
--- a/src/components/ids-tabs/ids-tabs-context.ts
+++ b/src/components/ids-tabs/ids-tabs-context.ts
@@ -13,7 +13,7 @@ import styles from './ids-tabs-context.scss';
 import type IdsTabContent from './ids-tab-content';
 import type IdsTabs from './ids-tabs';
 import { stringToBool } from '../../utils/ids-string-utils/ids-string-utils';
-import type { IdsValidateEvent, IdsValidatedElement } from '../../mixins/ids-validation-mixin/ids-validation-mixin';
+import { isIdsValidatedElement, type IdsValidateEvent } from '../../mixins/ids-validation-mixin/ids-validation-mixin';
 
 const Base = IdsOrientationMixin(
   IdsEventsMixin(
@@ -64,9 +64,8 @@ export default class IdsTabsContext extends Base {
       if (!tab) {
         return;
       }
-      const validatedElements = content.querySelectorAll<IdsValidatedElement>('[validate]');
-      const valid = [...validatedElements].every((el) => el.isValid);
-      tab.hasError = !valid;
+      const validatedElements = [...content.querySelectorAll('[validate]')].filter(isIdsValidatedElement);
+      tab.hasError = validatedElements.some((el) => !el.isValid);
     });
 
     this.#afterConnectedCallback();
--- a/src/components/ids-tabs/ids-tabs-context.ts
+++ b/src/components/ids-tabs/ids-tabs-context.ts
@@ -12,8 +12,8 @@ import styles from './ids-tabs-context.scss';
 
 import type IdsTabContent from './ids-tab-content';
 import type IdsTabs from './ids-tabs';
-import type IdsTab from './ids-tab';
 import { stringToBool } from '../../utils/ids-string-utils/ids-string-utils';
+import type { IdsValidateEvent, IdsValidatedElement } from '../../mixins/ids-validation-mixin/ids-validation-mixin';
 
 const Base = IdsOrientationMixin(
   IdsEventsMixin(
@@ -55,18 +55,18 @@ export default class IdsTabsContext extends Base {
       this.#changeContentPane(this.value, this.value);
     });
 
-    this.onEvent('validate', this, () => {
-      this.querySelectorAll<IdsTabContent>('ids-tab-content').forEach((content) => {
-        // TODO: Figure out how to extract the type from the mixin
-        type ValidationMixinElement = HTMLElement & { isValid: boolean };
-
-        const tab = this.tabList?.querySelector<IdsTab>(`ids-tab[value="${content.value}"]`);
-        if (tab) {
-          const validated = content.querySelectorAll<ValidationMixinElement>('[validate]');
-          const valid = [...validated].every((el) => el.isValid);
-          tab.validationHasError = !valid;
-        }
-      });
+    this.onEvent('validate', this, (e: IdsValidateEvent) => {
+      const content = e.detail.elem?.closest<IdsTabContent>('ids-tab-content');
+      if (!content) {
+        return;
+      }
+      const tab = this.tabList?.tabs.find(({ value }) => value === content.value);
+      if (!tab) {
+        return;
+      }
+      const validatedElements = content.querySelectorAll<IdsValidatedElement>('[validate]');
+      const valid = [...validatedElements].every((el) => el.isValid);
+      tab.hasError = !valid;
     });
 
     this.#afterConnectedCallback();
--- a/src/components/ids-tabs/ids-tabs-context.ts
+++ b/src/components/ids-tabs/ids-tabs-context.ts
@@ -12,6 +12,7 @@ import styles from './ids-tabs-context.scss';
 
 import type IdsTabContent from './ids-tab-content';
 import type IdsTabs from './ids-tabs';
+import type IdsTab from './ids-tab';
 import { stringToBool } from '../../utils/ids-string-utils/ids-string-utils';
 
 const Base = IdsOrientationMixin(
@@ -52,6 +53,20 @@ export default class IdsTabsContext extends Base {
     // Set active pane when content is inserted dynamically
     this.onEvent('slotchange', this.container, () => {
       this.#changeContentPane(this.value, this.value);
+    });
+
+    this.onEvent('validate', this, () => {
+      this.querySelectorAll<IdsTabContent>('ids-tab-content').forEach((content) => {
+        // TODO: Figure out how to extract the type from the mixin
+        type ValidationMixinElement = HTMLElement & { isValid: boolean };
+
+        const tab = this.tabList?.querySelector<IdsTab>(`ids-tab[value="${content.value}"]`);
+        if (tab) {
+          const validated = content.querySelectorAll<ValidationMixinElement>('[validate]');
+          const valid = [...validated].every((el) => el.isValid);
+          tab.validationHasError = !valid;
+        }
+      });
     });
 
     this.#afterConnectedCallback();
--- a/src/mixins/ids-validation-mixin/ids-validation-mixin.ts
+++ b/src/mixins/ids-validation-mixin/ids-validation-mixin.ts
@@ -51,6 +51,17 @@ export type IdsValidateEvent = CustomEvent<{
  */
 export type IdsValidatedElement = InstanceType<ReturnType<typeof IdsValidationMixin>>;
 
+/**
+ * Check whether an element has the {@link IdsValidationMixin} applied.
+ * @param {Element} element The element to check.
+ * @returns {boolean} `true` if the element has the mixin.
+ */
+export function isIdsValidatedElement(element: Element): element is IdsValidatedElement {
+  // Note: Could introduce a Symbol or similar instead.
+  return typeof (element as IdsValidatedElement)?.checkValidation === 'function'
+  && typeof (element as IdsValidatedElement)?.validate === 'string';
+}
+
 type Constraints = IdsConstructor<EventsMixinInterface>;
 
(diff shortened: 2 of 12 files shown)


src/components/ids-tabs/ids-tabs-context.ts
```diff
@@ -54,6 +55,20 @@ export default class IdsTabsContext extends Base {
       this.#changeContentPane(this.value, this.value);
     });
 
+    this.onEvent('validate', this, () => {
+      this.querySelectorAll<IdsTabContent>('ids-tab-content').forEach((content) => {
+        // TODO: Figure out how to extract the type from the mixin
+        type ValidationMixinElement = HTMLElement & { isValid: boolean };
```
anhallbe, line 61: This

src/components/ids-tabs/ids-tabs-context.ts
```diff
@@ -54,6 +55,20 @@ export default class IdsTabsContext extends Base {
...
+
+        const tab = this.tabList?.querySelector<IdsTab>(`ids-tab[value="${content.value}"]`);
```
anhallbe, line 63: Maybe a method `IdsTabs.getTab(value: string): IdsTab` would make sense?

src/components/ids-tabs/ids-tabs-context.ts
```diff
@@ -54,6 +55,20 @@ export default class IdsTabsContext extends Base {
...
+        if (tab) {
+          const validated = content.querySelectorAll<ValidationMixinElement>('[validate]');
```
anhallbe, line 65: Not sure if all `[validate]` elements are guaranteed to be `ValidationMixinElement`. May need some form of `instanceof` check

src/mixins/ids-validation-mixin/ids-validation-mixin.ts
```diff
@@ -199,7 +200,9 @@ const IdsValidationMixin = <T extends Constraints>(superclass: T) => class exten
         }
       });
       this.isTypeNotValid = null;
-      this.triggerEvent('validate', this, { detail: { elem: this, value: (this as IdsInputInterface).value, isValid } });
+      if (isValid !== wasValid) { // Only trigger if the state has changed, for performance.
+        this.#triggerValidateEvent(isValid);
```
anhallbe, line 204: This used to fire on every keypress. Added the `isValid !== wasValid` check for performance, but I'm not sure if this breaks anything.

src/mixins/ids-validation-mixin/ids-validation-mixin.ts
```diff
@@ -563,6 +566,17 @@ const IdsValidationMixin = <T extends Constraints>(superclass: T) => class exten
...
   }
 
+  #triggerValidateEvent(isValid: boolean) {
+    this.triggerEvent('validate', this, {
+      detail: {
+        elem: this,
+        value: (this as IdsInputInterface).value,
+        isValid,
+      },
+      bubbles: true,
```
anhallbe, line 576: Note: Added `bubbles: true` here. Not sure if that has any unintended consequences?

src/components/ids-tabs/ids-tabs-context.ts
tmcconechy: I would include the type in the event data so that you can do this...

```
this.onEvent('validate', this, (e: Event) => {
```

Somewhere around here but you changed it https://github.com/infor-design/enterprise-wc/blob/main/src/mixins/ids-validation-mixin/ids-validation-mixin.ts#L202 ? 

Maybe the whole rule including the type?
```
{ detail: { elem: this, value: (this as IdsInputInterface).value, isValid, rule: thisRule.rule } 
```

src/components/ids-tabs/ids-tabs-context.ts
tmcconechy: Or a getter `get tabByValue` in case we need id / ect later.

src/components/ids-tabs/ids-tabs-context.ts
tmcconechy: Yes they "should" all use the mixin for any components that support it. See also ids-form.ts

src/mixins/ids-validation-mixin/ids-validation-mixin.ts
tmcconechy: This seems ok. But does it account for the case of clearing the field? i.e. reseting the state by clearing the text ect.

src/mixins/ids-validation-mixin/ids-validation-mixin.ts
tmcconechy: It will just bubble if you do that. Its probably what we want here 👍🏻

tmcconechy:
> Handle different validation error levels (error, warning, info)

Be nice if you did... Not required for me

  > Add an indicator to the "more" menu in case the invalid tab is overflowed

We didnt have this but i think if we did it would just be the icon next to the item.

  >  Add an option to disable the error indicator?
 
 Yes maybe DISABLE_VALIDATION. What was the `VALIDATION_HAS_ERROR` for? Just a flag?
 
 > Handle / test validation errors from ids-enterprise controls?
 
 Not sure?
 
  > Right now this is implemented using the validation-has-error attribute. Not sure if this is the intended purpose of the attribute?
 
 I dont know what that is for? Its only used on radio button. I think it can go. I dont see what its doing maybe just to add the class and show it has an error? Would use hasError instead maybe 
  
> The 'validated' event now bubbles so that parent elements can detect it (IdsTab, IdsModal, IdsForm etc). Not sure if there's a better/preferred way to do this?
I think this one makes sense to do it that way.

> Test with more tab variants (module, counts etc).
Can push to later...

> Needs docs / changeling

Please and maybe a test.

src/components/ids-tabs/ids-tabs-context.ts
anhallbe: Added `IdsValidateEvent` and `IdsValidatedElement` in [bac6645](https://github.com/infor-design/enterprise-wc/pull/2801/commits/bac664593a570d5fb1664c8d73fc2be2fa506a7a)

src/components/ids-tabs/ids-tabs-context.ts
anhallbe: Added a `tabs` getter in [bac6645](https://github.com/infor-design/enterprise-wc/pull/2801/commits/bac664593a570d5fb1664c8d73fc2be2fa506a7a)

anhallbe:
Needs a bit of work to align the icon for the `dismissible`, `count` and `module` cases (and all of these combined with `orientation="vertical"`...)

image

ericangeles:
@anhallbe, have we considered the RTL style here?

tests/ids-tabs/ids-tabs.spec.ts
```diff
@@ -196,4 +197,36 @@ test.describe('IdsTabs tests', () => {
...
+
+    test('should only be visible when there is a validation error', async ({ page }) => {
+      const contexts = await page.locator('ids-tabs-context').all();
+      for (const context of contexts) {
+        const tab = context.locator('ids-tab[value="with-required"]');
+        const content = context.locator('ids-tab-content[value="with-required"]');
+        const input = content.locator('ids-input').first();
+        const icon = tab.locator('ids-icon[icon="error"]');
+
+        await input.evaluate((el: IdsInput) => { el.value = 'hello'; });
```
ericangeles, line 223: Maybe we can avoid repetition here by creating a function?

```js
const setInputValue = async (value: string) => {
  await input.evaluate((el: IdsInput) => { el.value = value; });
};
```

src/components/ids-tabs/ids-tabs-context.ts
```diff
@@ -54,6 +55,20 @@ export default class IdsTabsContext extends Base {
       this.#changeContentPane(this.value, this.value);
     });
 
+    this.onEvent('validate', this, (e: IdsValidateEvent) => {
```
ericangeles, line 58: This looks good to me. 👍🏼

src/components/ids-tabs/ids-tabs-context.ts
anhallbe: Added `isIdsValidatedElement(element)` in 9d00dd260111cba8f15bc9e802994147ad0d4b2f

src/mixins/ids-validation-mixin/ids-validation-mixin.ts
anhallbe: Like this?

https://github.com/user-attachments/assets/967e1b9b-f6ff-4bda-9c3c-d05c4018d921

anhallbe:
> @anhallbe, have we considered the RTL style here?

anhallbe:
> > Handle different validation error levels (error, warning, info)
> 
> Be nice if you did... Not required for me

Not implemented right now. But the icon is now slotted so people can implement that if they need to: cf59264

> > Add an indicator to the "more" menu in case the invalid tab is overflowed
> 
> We didnt have this but i think if we did it would just be the icon next to the item.

I guess for feature parity this can be saved for later, then?

> > Add an option to disable the error indicator?
> 
> Yes maybe DISABLE_VALIDATION. What was the `VALIDATION_HAS_ERROR` for? Just a flag?

Added a slot for the error, which can be hidden: cf59264

> > Handle / test validation errors from ids-enterprise controls?
> 
> Not sure?

Added a slot for the error, so this can be implemented by consumers in the meantime: cf59264

> > Right now this is implemented using the validation-has-error attribute. Not sure if this is the intended purpose of the attribute?
> 
> I dont know what that is for? It's only used on radio button. I think it can go. I dont see what it's doing maybe just to add the class and show it has an error? Would use hasError instead maybe

Switched to `has-error`.


> > Test with more tab variants (module, counts etc).
> > Can push to later...

@tmcconechy there appear to be an existing issue with the "dismissible" button (and this new error icon) being misaligned.

image


> > Needs docs / changeling
> 
> Please and maybe a test.

Added docs & test

tmcconechy:
@anhallbe usually in that case just go ahead and include i fix with it (when i see issues in the same place I just correct them)

anhallbe:
The alignment issue has been fixed, and I added a note in the Changelog (v 1.5.0). I think this is ready for review/test/merge.

tmcconechy:
@anhallbe This is great! Can you just check the tests https://github.com/infor-design/enterprise-wc/actions/runs/10826090266/job/30044182690?pr=2801#step:12:2667

Yep, This will be in 1.6.0 or 1.5.1 depending on semantic versioning as things change over the sprint. (Probably 1.6.0 considering whats in the Queue)

anhallbe:
@tmcconechy the failing test was because of my change in when the `validate` event fired.

I had a change of heart regarding that one so I reverted the change in 41533ba. The `'validate'` event will again always fire when `checkValidation()` is called. This is consistent with how native elements always fire the `invalid` event when `checkValidity()` is called.

Also added a 1.6.0 section to the Changelog

tmcconechy (approved):

PR #2804, opened 2024-09-06
2726 - IdsDataGrid dropdown editor async beforecelledit

### LINKED ISSUE #2726: IdsDropdown Editor - Options set are not showing at first time if is set after a delay in an event
**Describe the bug**
I need to populate the options on `beforecelledit` event because each dropdown editor even on the same column may bring different options, so we need one request per dropdown.

Also we need to consider the delay on server request by adding setTimeout.

The result is that on the first time you open a dropdown to see the options, it fails only at the first time. By removing the timeout is going to work fine.

**To Reproduce**

```html
<ids-data-grid id="data-grid-1" editable=true></ids-data-grid>
```

```javascript
var grid = document.querySelector('#data-grid-1');

grid.columns = [{
  id: 'id',
  name: 'ID',
  field: 'id'
},{
  id: 'role',
  name: 'Role',
  field: 'role',
  formatter: grid.formatters.dropdown,
  editor: {
    type: 'dropdown',
    editorSettings: {
      autoselect: true,
      dirtyTracker: true
      //options: [] // will be set later with the server
    }
  }
}];

grid.data = [{
  id: 1,
  role: 'Administration'
},{
  id: 2,
  role: 'Manager'
}];

// set option before the cell is edited because each dropdown editor even on the same column may bring different options
grid.addEventListener('beforecelledit', function(evt) {
  var ed = evt.detail.column.editor;

  if (ed && ed.type === 'dropdown') {
    // simulate the delay of the options comming from the server
    setTimeout(function() {
      ed.editorSettings.options = [{
        value: 'Administration',
        label: 'Administration'
      },{
        value: 'Manager',
        label: 'Manager'
      },{
        value: 'Developer',
        label: 'Developer'
      }];
    }, 200);
  }
});
```

**Version**
1.4.2

**Screenshots**
Screenshot 2024-08-14 at 11 16 51 AM

jdilag90:
Users can now pass an async function or Promise to the vetoable `beforecelledit` event's `response` callback.
This allows user to perform aync operations like database fetches before vetoing `beforecelledit` event.
image

Closes #2726 

1. Checkout branch
2. Go to http://localhost:4300/ids-data-grid/beforecelledit-async.html
3. Test `bookCurrency` column dropdowns and make sure dropdown options load properly
4. Test http://localhost:4300/ids-data-grid/editable.html for any regressions


tmcconechy (approved):

glenlieorillo (approved):
https://github.com/user-attachments/assets/5bbfe1c6-e011-4cd5-adf5-c1cb5dc034d5

